PR #76, opened 2015-01-14
Added baseclasses for allow_None and Tuple

philippjfr:
In response to the recent issue https://github.com/ioam/param/issues/74 and the general annoyance with allow_None being only supported by a few Parameters I've created a SupportsAllowNone baseclass as Jim suggested. Additionally I've added a Tuple parameter as the baseclass for NumericTuple.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -19,7 +19,7 @@ Parameters and Parameterized classes.
 import os.path
 
 from .parameterized import Parameterized, Parameter, String, \
-     descendents, ParameterizedFunction, ParamOverrides
+     SupportsAllowNone, descendents, ParameterizedFunction, ParamOverrides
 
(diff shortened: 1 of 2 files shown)


jlstevens:
How about naming the new class ``AllowNone`` instead of ``SupportsAllowNone``? You need to know what ``AllowNone`` (or ``allow_None``) means in param anyway; saying that you 'support' isn't particularly informative!

philippjfr:
I accidentally merged this pulled request but have undone that now. Will resubmit it.

jbednar:
SupportsAllowNone is awkward,  but AllowNone would be misleading.  Subclasses of it allow ``allow_None``, but instances may or may not allow None, depending on the value of the keyword.  Tricky, huh? And SupportsAllowNone is at least better than AllowsAllowNone....

PR #77, opened 2015-01-14
Added Tuple and AllowsNone Parameters

philippjfr:
Resubmission of PR https://github.com/ioam/param/pull/76, which was accidentally merged (I've undone that now).

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -527,9 +527,9 @@ class Number(Dynamic):
       AB = Number(default=0.5, bounds=(None,10), softbounds=(0,1), doc='Distance from A to B.')
     """
 
-    __slots__ = ['bounds','_softbounds','allow_None','inclusive_bounds','set_hook']
+    __slots__ = ['bounds','_softbounds','inclusive_bounds','set_hook']
 
-    def __init__(self,default=0.0,bounds=None,softbounds=None,allow_None=False,inclusive_bounds=(True,True),**params):
+    def __init__(self,default=0.0,bounds=None,softbounds=None,inclusive_bounds=(True,True),**params):
         """
         Initialize this parameter object and store the bounds.
 
@@ -541,7 +541,6 @@ class Number(Dynamic):
         self.bounds = bounds
         self.inclusive_bounds = inclusive_bounds
         self._softbounds = softbounds
-        self.allow_None = (default is None or allow_None)
         if not callable(default): self._check_value(default)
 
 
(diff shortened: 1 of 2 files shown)


param/__init__.py
```diff
@@ -775,22 +773,37 @@ def __init__(self,default=(0,0),length=None,**params):
...
         self._check(val)
-        super(NumericTuple,self).__set__(obj,val)
+        super(Tuple,self).__set__(obj,val)
+
+
+
+class NumericTuple(Tuple):
+    """A numeric tuple Parameter (e.g. (4.5,7.6,3)) with a fixed tuple length."""
+
+    __slots__ = []
```
ceball: Note that we add `__slots__` if Parameter subclass authors don't do it themselves:

```python
class ParameterMetaclass(type):
    [...]
    def __new__:
        [...]
        # To get the benefit of slots, subclasses must themselves define
        # __slots__, whether or not they define attributes not present in
        # the base Parameter class.  That's because a subclass will have
        # a __dict__ unless it also defines __slots__.
        if '__slots__' not in classdict:
            classdict['__slots__']=[]
```
https://github.com/ioam/param/blob/master/param/parameterized.py#L206

That seemed like a good idea at the time (when nobody had ever heard of `__slots__`, and when we had to discover things like this ourselves/read python's source code).

jbednar: Ah, that explains why I'd never seen it in our other code.

philippjfr: So what should we do going forward? Should we simply rely on the metaclass to patch the Parameters with __slots__ or be explicit and declare "empty" slots? The documentation is there now but it's still non-obvious and I probably would have missed it had I not gone out of my way to learn about slots.

philippjfr:
So what are everyone's thoughts on this. Should allow_None move onto Parameter or are we happy with an intermediate AllowsNone superclass like I have implemented here. I would like this stuff to be merged in the near future as my latest HoloViews refactor relies on it.

philippjfr:
BTW, once we've decided what to do about allow_None I'll redo these commits cleanly. Don't want the fix commits I had to make, to make it into master.

param/__init__.py
jbednar: On balance, I'd vote for leaving out the empty slots declarations, because we aren't using them consistently, and people will wonder (a) what it's there for (which they can find in Python documentation, but there's still probably no point in them going to look it up), and (b) why they aren't everywhere.  Seems better to be consistent; the fact that we use slots shouldn't matter to most users of Param, and it won't hurt if they add empty slots themselves, but we don't need to implicitly suggest that they do so by putting a few around to confuse them...

jbednar:
I don't have a strong opinion.  Given that there are arguments both ways, I think the person doing the work should get to decide. :-)

param/__init__.py
philippjfr: Agreed, I'll leave it out when I make my new, clean commits.

jlstevens:
Obviously if ``allow_None`` does work at the ``Parameter`` level (and it isn't too much work and no one forsees any problems) then it should go there. That way we always know that every ``Parameter`` has to potential to be  ``None``.

jbednar:
Well, I do foresee a problem, i.e. that any user-written Parameter that currently works and doesn't support allow_None will now be partially broken, because it will appear to (but won't really) support allow_None.  But that's not a major problem, and may be worth the potential savings in complexity by simply stating that all Parameter types must support allow_None.  Note that this is different from saying that every Parameter has the potential to be None, or at least that's misleading; Parameters can be None only if allow_None is True.  I.e., this is providing AllowAllowNone, not AllowNone!

jlstevens:
Looks good to me and the tests are all passing. 

As we need to use Tuple in some of our more recent work on holoviews, I'll merge this PR now. If there are any major objections, we can always revert the merge later. For anything else we wish to tweak, we can just commit our changes straight to master.

Issue #85, opened 2015-01-30
ParamOverrides seem to be broken


jlstevens:
Surely, ``ParamOverrides`` shouldn't be allow this? 

```python
import param

class Foo(param.ParameterizedFunction):
    a = param.String('bar')
    b = param.Boolean()
    
    def __call__(self, **params):
        print param.ParamOverrides(self,params)
```

```python
>>> Foo(a=slice, b=3)
{'a': <type 'slice'>, 'b': 3} overriding params from Foo(a='bar', b=False, name='Foo')
```

Shouldn't the types be checked? Otherwise what is the point of ``ParamOverrides``?

Maybe I am just getting confused and don't understand what ``ParamOverrides`` are for...

jbednar:
From the code, I don't see any type-checking done by ParamOverrides; it's only checking that there is a corresponding Parameter, not what its type or bounds are.  That does seem like it opens up a lot of holes, and it doesn't seem like a good idea at all.  It looks to me like it has access to the underlying Parameterized object overriden, and so it should be straightforward to call the checking methods on the data we have been supplied, so I don't know why it hasn't been done already.  Seems unsafe to me! :-(

ceball:
Oops.

ceball:
How far do we want ParamOverrides to go in behaving like the overridden object? 

For instance, should you be able to override a dynamic parameter with a function and have that function be transparently called when you request the parameter value from ParamOverrides? Right now, if someone overrides a dynamic parameter with a function, they'll get the function back rather than the result of calling that function (I think). So e.g. something like

```python
class A(param.Parameterized):
    a = param.String(default='hi')
    b = param.Number(default=0.5, bounds=(0,1))

a = A()

po = param.ParamOverrides(a, dict(b=lambda:0.75))

print po.a
print po.b
```

gives something like

```
hi
(lambda function...)
```

instead of what I think someone would expect:

```
hi
0.75
```

So, apart from not checking types/values etc, I think ParamOverrides doesn't behave enough like the overridden object in this way too. What do you think?

ceball:
If we want ParamOverrides to behave a lot like the overridden instance, then I guess I propose that instead of the current implementation, a ParamOverrides instance should basically be a Parameterized instance that shares the overridden Parameterized instance's parameters (i.e. the parameter objects) ;)

jbednar:
[...]

Issue #36, opened 2012-10-02
ParamOverrides: create overrides attr?


sf-issues:
Converted from SourceForge issue 2888861, submitted by ceball
Submit Date: 2009-10-29 15:43 GMT

set e.g. .overrides on Parameterized instance, and override __getattribute__ to look in .overrides first; would have to set .overrides=None after __call__. Or something like that.

ceball:
I suppose we could do it with a decorator...

```
class Fn(param.ParameterizedFunction):
    x = param.Number(1)

    def __call__(self,**params):
        p = ParamOverrides(self,params)
        return p.x
```

could become

```
class Fn(param.ParameterizedFunction):
    x = param.Number(1)

    @add_overrides
    def __call__(self,**params):
        return self._overrides.x
```

In some future less polluting api (#154) that might be ``self.params.overrides.x`` or something.

ceball:
I added in #92:
```
class A2(param.Parameterized):
    path = param.Path(search_paths=['/etc'])
    @overrides
    def __call__(self,**params):
        return self._overrides.path
    
a2 = A2(path=test_file)
assert a2() == test_path+test_file, a2()
assert a2(path=test_file) == test_path+test_file, a2(path=test_file)
```

What do people think?

jlstevens:
Personally, I prefer the explicit use of ``ParamOverrides`` instead of introducing a decorator...

sf-issues:
Same; decorator doesn't seem like an improvement.

ceball:
Not adding this feature

jlstevens:
Jim? Are you still logged in as sf-issues?!

jbednar:
Yep, that's me!

PR #92, opened 2015-02-15
Alternative implementation of ParamOverrides

ceball:
I've re-implemented ParamOverrides, following the suggestion made in #85 about sharing parameter objects. This PR isn't meant for merging at this stage, I'd just like some comments on the basic idea.

This new implementation passes all the tests (including one I added to check that dynamic overrides work), although the implementation's not yet complete (just missing some minor methods that the current ParamOverrides has). However, what I'd like to know is: is the whole idea crazy? If it's not crazy, is there a way to make the implementation simpler? I've left extra comments in the code for now to explain what I was thinking.

In case it's easier to look at the class in one piece:
https://github.com/ceball/param/blob/bug85_newParamOverrides/param/parameterized.py#L1580

(Also, note that I haven't considered speed.)

### DIFF
--- a/tests/testparameterizedobject.py
+++ b/tests/testparameterizedobject.py
@@ -375,6 +375,33 @@ class TestParamOverrides(unittest.TestCase):
         assert overrides.get('bogus') is None
         assert overrides.get('bogus',20) is 20
 
+    # this test is about checking things are looked up normally (like
+    # for dynamic tests above) so is probably redundant now
+    # paramoverrides is itself a parameterized object.
+    def test_search_paths(self):
+        # TODO: only good on linux/max - should do this properly
+        test_file = 'passwd'
+        test_path = '/etc/'
+
+        ###
+        # this will fail if the test itsel isn't working
+        class A(param.Parameterized):
+            path = param.Path(search_paths=['/etc'])
+        a = A(path=test_file)
+        assert a.path==test_path+test_file
+        ###
+
+        # the actual test...
+        class A2(param.Parameterized):
+            path = param.Path(search_paths=['/etc'])
+            def __call__(self,**params):
+                p = param.ParamOverrides(self,params)
+                return p.path
+    
+        a2 = A2(path=test_file)
+        assert a2() == test_path+test_file, a2()
+        assert a2(path=test_file) == test_path+test_file, a2(path=test_file)
+
 
 class TestSharedParameters(unittest.TestCase):
 
--- a/tests/testparameterizedobject.py
+++ b/tests/testparameterizedobject.py
@@ -28,7 +28,7 @@ import random
 from nose.tools import istest, nottest
 
 
-from param.parameterized import ParamOverrides, shared_parameters
+from param.parameterized import ParamOverrides, shared_parameters, overrides
 
 @nottest
 class _SomeRandomNumbers(object):
@@ -402,6 +402,36 @@ class TestParamOverrides(unittest.TestCase):
         assert a2() == test_path+test_file, a2()
         assert a2(path=test_file) == test_path+test_file, a2(path=test_file)
 
+    # this test's as bad as the one above except this one's still useful
+    def test_overrides_decorator(self):
+        # TODO: only good on linux/max - should do this properly
+        test_file = 'passwd'
+        test_path = '/etc/'
+
+        ###
+        # this will fail if the test itsel isn't working
+        class A2(param.Parameterized):
+            path = param.Path(search_paths=['/etc'])
+            def __call__(self,**params):
+                p = param.ParamOverrides(self,params)
+                return p.path
+
+        # the actual test...
+        a2 = A2(path=test_file)
+        assert a2() == test_path+test_file, a2()
+        assert a2(path=test_file) == test_path+test_file, a2(path=test_file)
+        ###
+
+        class A2(param.Parameterized):
+            path = param.Path(search_paths=['/etc'])
+            @overrides
+            def __call__(self,**params):
+                return self._overrides.path
+    
+        a2 = A2(path=test_file)
+        assert a2() == test_path+test_file, a2()
+        assert a2(path=test_file) == test_path+test_file, a2(path=test_file)
+        
 
 class TestSharedParameters(unittest.TestCase):
 
--- a/tests/testparameterizedobject.py
+++ b/tests/testparameterizedobject.py
@@ -375,6 +375,8 @@ class TestParamOverrides(unittest.TestCase):
         assert overrides.get('bogus') is None
         assert overrides.get('bogus',20) is 20
 
+        assert [('name', overrides.name), ('print_level', 0), ('some_param', 'y')] == sorted(overrides.items())
+
     # this test is about checking things are looked up normally (like
     # for dynamic tests above) so is probably redundant now
     # paramoverrides is itself a parameterized object.
--- a/tests/testparameterizedobject.py
+++ b/tests/testparameterizedobject.py
@@ -289,6 +289,50 @@ class TestParamOverrides(unittest.TestCase):
         self.assertEqual(overrides['name'], 'B')
         self.assertEqual(overrides['print_level'], 0)
 
+    # CB: would be nice if we could repeat an existing set of
+    # tests somehow (since here we're testing that ParamOverrides'
+    # behavior doesn't differ from that of the wrapped
+    # Parameterized instace).            
+
+    def test_dynamic_override(self):
+        class ATestPO(param.Parameterized):
+            a = param.Number(0.5, bounds=(0,1))
+
+        po = ParamOverrides(ATestPO(),dict(a=lambda:0.75))        
+        self.assertEqual(po.a,0.75)
+
+        po = ParamOverrides(ATestPO(),dict(a=lambda:2))
+        try:
+            po.a
+        except ValueError:
+            pass
+        else:
+            raise AssertionError("Dynamic value wasn't checked")
+    
+
+
+    def test_value_checking(self):
+        class ATestPO(param.Parameterized):
+            a = param.String('test')
+            b = param.Number(0.5, bounds=(0,1))
+
+        # these three similar tests might be overkill
+        try:
+            ParamOverrides(ATestPO(),dict(a=10))
+        except ValueError:
+            pass
+        else:
+            raise AssertionError("Shouldn't be able to override a param.String with a number")
+
+        try:
+            ParamOverrides(ATestPO(),dict(b=2))
+        except ValueError:
+            pass
+        else:
+            raise AssertionError("Shouldn't be able to override a param.Number to value outside its bounds")
+
+        self.assertRaises(ValueError, some_fn, num_phase='bad value') 
+
     # CEBALERT: missing test for allow_extra_keywords (e.g. getting a
     # warning on attempting to override non-existent parameter when
     # allow_extra_keywords is False)
--- a/tests/testparameterizedobject.py
+++ b/tests/testparameterizedobject.py
@@ -431,6 +431,15 @@ class TestParamOverrides(unittest.TestCase):
         a2 = A2(path=test_file)
         assert a2() == test_path+test_file, a2()
         assert a2(path=test_file) == test_path+test_file, a2(path=test_file)
+        assert not hasattr(a2,'_overrides')
+
+        class A3(param.ParameterizedFunction):
+            path = param.Path(search_paths=['/etc'])
+            @overrides
+            def __call__(self,**params):
+                return self._overrides.path
+
+        assert A3(path=test_file) == test_path+test_file, A3(path=test_file)
         
 
 class TestSharedParameters(unittest.TestCase):
--- a/tests/testparameterizedobject.py
+++ b/tests/testparameterizedobject.py
@@ -440,7 +440,27 @@ class TestParamOverrides(unittest.TestCase):
                 return self._overrides.path
 
         assert A3(path=test_file) == test_path+test_file, A3(path=test_file)
-        
+
+    # another test to simplify one day!
+    def test_extra_keywords(self):
+        try:
+            old = param.parameterized.warnings_as_exceptions
+            param.parameterized.warnings_as_exceptions = True
+            p = param.ParamOverrides(self.po,{'extra':100},allow_extra_keywords=False)
+        except:
+            pass
+        else:
+            raise AssertionError
+        finally:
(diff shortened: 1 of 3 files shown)


ceball:
Also, what do you think about this? Say a Parameterized instance ``a`` has a mutable parameter, e.g. ``b = param.List([1,2])``. If someone creates a ParamOverrides and sets that parameter to a new list either during initialization (``ParamOverrides(a,{'b':[3,4]})``) or afterwards (``ParamOverrides(a).b=[3,4]``), no problem. But if someone edits the mutable parameter without first having set it (``ParamOverrides(a).b[0]=3``), they'll actually be editing the parameter on the overridden instance (``a.b`` will be ``[3,2]`` in this case). To me that's probably ok with suitable documentation, but with a more complex implementation it would be possible to do better for ``instantiate=True`` mutable parameters.

(I think the above behavior is the same for the existing ParamOverrides.)

jbednar:
Seems ok to me, when finished.  Is it really necessary to make a new class with all the parameters?  I.e., is there no easy way just to point to the parameters as they are in the overridden class?  I haven't tried either approach, so I don't know.

I don't think the mutable object behavior you describe is worth putting in a special case for; people already have to know to deal with mutable objects differently in general.

ceball:
On Sun, Feb 15, 2015, at 07:23 PM, James A. Bednar wrote:
> Seems ok to me, when finished.  Is it really necessary to make a new
> class with all the parameters?  I.e., is there no easy way just to point
> to the parameters as they are in the overridden class? 

A parameterized instance doesn't have pointers to its parameter objects,
except via its class. The new class is where I store the pointers to the
overridden parameterized instance's parameter objects. No new parameter
objects are created (except the standard parameters of Parameterized,
since ParamOverrides is now a subclass of that).

Is this what you meant, or did you mean could I avoid having a new class
by storing elsewhere the pointers to the parameter objects (i.e. doing
something different from a normal Parameterized)?

Or am I on the wrong track completely?

jbednar:
You're the expert on Param, not me, so I'd usually be the one on any wrong track!  

What I meant was that I don't see why you need a *new* class to store the pointers to the parameterized instance's parameter objects.  Doesn't overridden's class already exist, with precisely the right set of parameter objects?  So I'm imagining (very vaguely, so that's as much detail as I can give) somehow pointing to the existing class, so that we don't have to make a new one.  But there could easily be technical reasons why this is not feasible.

philippjfr:
Is there any chance of finishing this up? The fact that ParamOverrides doesn't actually validate types is a major issue.

ceball:
When resuming this work, please consider addressing or closing:
* #36 
* #134 
* #173

ceball:
I think we've established that ParamOverrides has moved from something we're trying out to see if we want, to something that we use, so it's time to finish this! ;) I'll bring this PR up to date.

@jbednar: 
> What I meant was that I don't see why you need a new class to store the pointers to the parameterized instance's parameter objects. Doesn't overridden's class already exist, with precisely the right set of parameter objects? So I'm imagining (very vaguely, so that's as much detail as I can give) somehow pointing to the existing class, so that we don't have to make a new one. But there could easily be technical reasons why this is not feasible.

Yes, the overridden's class already exists, and yes it has exactly the right set of parameter objects. But...

Parameter objects are stored on the class. Different instances of ParamOverrides will override the parameters of different parameterizeds and will hence need different parameters. 

Your question is about having no new class, but in that case the ParamOverrides class would end up with links to parameters on multiple different parameterized classes i.e. it would not have the right parameters for any one parameterized class. 

My implementation is one new class for each ParamOverrides instance created, to store the links to the overridden's parameters. I did that because it's easy to do. Each new class is a subclass of overridden's class. However, it would instead be possible to create only one new class per *type* of overridden. I don't see how it's possible to do it with fewer classes than that while keeping the implementation of ParamOverrides inheriting from Parameterized. But to have only one new class per type of overridden, we'd need to keep the classes around.

Unless I missed something? Quite possible!

ceball:
What's left?

* [ ] add missing methods noted in code TODO
* [ ] test with holoviews
* [ ] test with topographica?
* [ ] test performance?
* [ ] fixed #134?
* [ ] fixed #173?
* [ ] address #36?

jlstevens:
I'm very happy to see these improvements to ParamOverride but I'm not sure about the decorator and corresponding underscore attribute described in #36.

ceball:
@jbednar: Maybe you could review this again? I think we could merge right after the next param release and then see what happens for a while. I've tested with holoviews but not yet topographica (the two things most likely to exercise paramoverrides the most). 

Notes:

* ParamOverrides is less like a dict than before (when it was a dict!). I've added what I think is the minimum dict-like behavior to support holoviews. We could make it emulate dict completely, though.

* ParamOverrides is at the same time slightly more dict-like than before, because now it raises KeyError rather than AttributeError when a name's not found via dict-like access.

* Because ParamOverrides now inherits from Parameterized, the name parameter will effectively always be overridden i.e. in typical usage ``p.name`` will not be the name of the overridden object but of the ParamOverrides object. That could be confusing. 

* Keep the overloads decorator or not (see #36)? If keeping, what about its name (and the name of "_overrides")?

* I haven't tested performance :( The following comment used to exist: ``we'd like __init__ to be fast because it's going to be called a lot``

* Fixes #134 and #173.

ceball:
@jlstevens:
> I'm not sure about the decorator and corresponding underscore attribute described in #36.

Yep, I agree. Wrote my comment above at the same time as yours. After initially liking the way it seems to make things look a bit nicer, I think I prefer the explicit ``p = ParamOverrides(...)``.

ceball:
Another note: #176 (issue #104) should be merged first or at the same time, or else holoviews will have spurious failing tests next time holoviews' travis runs (it takes param from github, I think).

For reference:

```
======================================================================
ERROR: Assert histogram shares colormapper
----------------------------------------------------------------------
  File "/Users/cball/code/ioam/holoviews/tests/testplotinstantiation.py", line 839, in test_side_histogram_cmapper
    plot = bokeh_renderer.get_plot(img.hist())
  File "/Users/cball/code/ioam/holoviews/holoviews/core/element.py", line 41, in hist
    individually=individually, dimension=d, **kwargs)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 1881, in __new__
    return inst.__call__(*args,**params)
  File "/Users/cball/code/ioam/holoviews/holoviews/core/operation.py", line 125, in __call__
    self.p = param.ParamOverrides(self, params)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 1758, in __init__
    Parameterized.__init__(self,**dict_)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 1041, in __init__
    self._setup_params(**params)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 978, in override_initialization
    fn(parameterized_instance,*args,**kw)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 1441, in _setup_params
    setattr(self,name,val)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 537, in __set__
    self._check_value(val)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 534, in _check_value
    raise ValueError("String '%s' only takes a string value: %s"%(self._attrib_name,val))
ValueError: String 'dimension' only takes a string value: None

======================================================================
ERROR: test_side_histogram_no_cmapper (tests.testplotinstantiation.TestBokehPlotInstantiation)
----------------------------------------------------------------------
  File "/Users/cball/code/ioam/holoviews/tests/testplotinstantiation.py", line 827, in test_side_histogram_no_cmapper
    plot = bokeh_renderer.get_plot(points.hist())
  File "/Users/cball/code/ioam/holoviews/holoviews/core/element.py", line 41, in hist
    individually=individually, dimension=d, **kwargs)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 1881, in __new__
    return inst.__call__(*args,**params)
  File "/Users/cball/code/ioam/holoviews/holoviews/core/operation.py", line 125, in __call__
    self.p = param.ParamOverrides(self, params)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 1758, in __init__
    Parameterized.__init__(self,**dict_)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 1041, in __init__
    self._setup_params(**params)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 978, in override_initialization
    fn(parameterized_instance,*args,**kw)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 1441, in _setup_params
    setattr(self,name,val)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 537, in __set__
    self._check_value(val)
  File "/Users/cball/code/ceball/param/param/parameterized.py", line 534, in _check_value
    raise ValueError("String '%s' only takes a string value: %s"%(self._attrib_name,val))
ValueError: String 'dimension' only takes a string value: None
```

ceball:
Testing with topographica has revealed two things missing from this implementation vs the previous one: (1) support for pickling, and (2) support for arbitrary (non parameter) attributes on parameterized function instances.

ceball:
Topographica's tests also show a use case I'd completely forgotten about: `p=ParamOverrides(...); f(**p)`. I think `p` has to *be* a dict for that to work (i.e. `p` can't just have the right methods). Or maybe it's just got to be a `Mapping` (abstract base class) - can't remember if that's in python 2.6 or not. Either way I think that makes it difficult for the implementation proposed in this PR to be backwards compatible.

E.g. from imagen:

```python
class GaussianCloud(Composite, TimeAware):
    ...
    def __call__(self,**params_to_override):
        p = ParamOverrides(self,params_to_override)
        ...
        return super(GaussianCloud,self).__call__(**p)
```

ceball:
Also, missing `param_keywords()` method.

ceball:
Summary so far! :(

Currently, we use ParamOverrides like this: `p = ParamOverrides(overridden, overrides)`, where `overridden` is a Parameterized instance and `overrides` is a dict. Typically we use this in the `__call__` method of `overridden`.

I think we want `p` to behave a lot like `overridden`. When you try to set things on `p`, you should get param's usual behaviors (e.g. value checking). When you get things from `p`, you should get param's usual behaviors (e.g. dynamic values). For example, this should work:

```python
class A(param.Parameterized):
    x = param.String(default='hi')
    y = param.Number(default=0.5, bounds=(0,1))

a = A()

try:
    param.ParamOverrides(a, dict(x=1))
except ValueError:
    pass
else:
    raise

param.ParamOverrides(a, dict(y=lambda:0.75)).y == 0.75

try:
    param.ParamOverrides(a, dict(y=lambda:1.5)).y
except ValueError:
    pass
else:
    raise
```

If ParamOverrides doesn't work just like param normally does I think it's confusing. We have at least one example of that already: "ParamOverrides stops Filename search_paths from working" (https://github.com/ioam/param/issues/134), but there'll be others. I've also found myself wanting to use the normal methods available on Parameterized, too. E.g. "ParamOverrides: support get_param_values()" (https://github.com/ioam/param/issues/173). How far should we go? I think ParamOverrides should be just like the overridden instance.

Do you agree with the above? If you think it's too much, and we should instead have ParamOverrides support only a specific, smaller set of things, then please say. 

I thought the simplest and easiest way to achieve the above would be for ParamOverrides to inherit from Parameterized. I felt that alternative suggestions would result either in ParamOverrides being some kind of reimplementation of Parameterized, or in ParamOverrides requiring more magic to access features of Parameterized via overridden than exists in the proposed ParamOverrides-inherits-from-Parameterized implementation.

I created what I think is a bit of an ugly solution, but it works (in that it fixes an important gap in parameter checking, particularly for parameterized functions). I knew I could easily add anything necessary to make the new implementation backwards compatible with the previous one (except for the part about catching parameter errors i.e. the reason for this PR), giving us something working right now (the priority), while hoping in the future we might be able to simplify things/find an entirely better solution. However, it turns out that there's one thing I probably can't do. As noted above (https://github.com/ioam/param/pull/92#issuecomment-328301714), you can only use a Mapping for kwargs. In topographica and imagen (but not holoviews), that appears be done with ParamOverrides a couple of times (e.g. in a `__call__` method: ``p=ParamOverrides(...); ...; super(...).__call__(**p)``. 

So what should I do? Add a method to get a dictionary for use in kwargs and update imagen and topographica and apologize that ParamOverrides may start generating two types of error for your old code (genuine parameter errors, and an error if you try to use it for kwargs without updating your code)? Or maybe make the new ParamOverrides even trickier than it already is by also inheriting from Mapping? Or have I gone down a miserable technical rabbit hole, and can someone suggest an altogether simpler approach?

Jim previously suggested this:

> maybe we should provide a totally different way to do this that's less awkward? E.g. can we offer a method decorator that pushes the current dictionary of parameter values onto a stack, lets the method do what it likes, then pops the original values back off when the method returns?

However, I would be worried about changing "the state of the parameters". You might look at a `__call__` method and decide it doesn't modify anything so you can use it "safely" from multiple threads (or in e.g. a generator), without realizing what's happening to the parameters. It might be particularly surprising for someone to find calling 'a function' with some arguments in one place could affect the function's parameters in another. E.g.

```python
class A(param.ParameterizedFunction):
    x = param.Number(default=0)

                   # pushes __dict__ onto stack and replaces with temp copy into which
    @override_dict # ** params are set; pops __dict__ back off stack on return
    def __call__(self, **params):
        # some long-running operation e.g. a network operation...

a = A.instance(x=0)

with concurrent futures ThreadPool()) as executor: # (I can't quite remember how that goes)
    executor.submit(a) # will likely have a.x = 0
    executor.submit(a, x=10)
    executor.submit(a) # will likely have a.x = 10 because a(x=10) will still be running
```

tests/testparameterizedobject.py
```diff
@@ -297,13 +344,109 @@ def test_missing_key(self):
...
+        
+    def test_param_keywords(self):
+        overrides = ParamOverrides(self.po,{'name':'B','an_extra_thing':1},allow_extra_keywords=True)
+        assert 'an_extra_thing' in overrides.extra_keywords()
+        assert 'an_extra_thing' not in overrides.param_keywords()
+
+    def test_pickle(self):
+        overrides = ParamOverrides(self.po,{'name':'B','an_extra_thing':1},allow_extra_keywords=True)
+        import pickle
+        pkl = pickle.dumps(overrides)
```
ceball: Should test pickle load also. (Note: pickle load seems to be broken for the current ParamOverrides anyway. Should investigate.)

tonyfast:
Closing this as stale. We are likely to harvest some code from here, but it would require significant work.

PR #93, opened 2015-02-21
Added seed parameter to RandomDistribution number generators

jlstevens:
Adds an explicit seed parameter for random number generators. This seed will now also control time-dependent random streams (previously you had to mangle the name with the seed yourself).

I have tested that this works with the updated SOM tutorial (time dependent streams for the input generator)  but I cannot guarantee that this won't break everything else!

### DIFF
--- a/numbergen/__init__.py
+++ b/numbergen/__init__.py
@@ -315,6 +315,9 @@ class TimeAwareRandomState(TimeAware):
         """
         if seed is None: # Equivalent to an uncontrolled seed.
             seed = random.Random().randint(0, 1000000)
+            suffix = ''
+        else:
+            suffix = str(seed)
 
         # If time_dependent, independent state required: otherwise
         # time-dependent seeding (via hash) will affect shared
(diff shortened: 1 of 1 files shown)


numbergen/__init__.py
```diff
@@ -399,9 +412,8 @@ def __init__(self,**params):
 
         Note that any supplied seed is ignored if time_dependent=True.
```
jbednar: Is that note still true?

jbednar:
Looks good to me.

numbergen/__init__.py
jlstevens: Nope. That can be easily fixed though.

jlstevens:
Other than updating that one docstring, do you see any reasons not to merge?

jbednar:
Seems ok to me.

jlstevens:
Merge when you are ready...this commit is needed by the updated version of the som_retinotopy notebook that I am about to commit.

PR #95, opened 2015-02-22
Added pprint function to parameterized.py

jlstevens:
This PR adds something I've long wished for in param: sensible, short reprs that actually work. :-)

Although I started implementing this in ``__repr__``,  I saw the note about ``repr`` needing to be fast and then decided against implementing it there once I realized what was required.

Instead of changing ``__repr__``, I've ended up making a more robust ``pprint`` function that can be used as follows:

```python
    
    a = param.Number(4, precedence=6)
    
    b = param.String('B', precedence=-9)
    
    c = param.Number(4, precedence=0)
    
    z = param.Number(4, precedence=0)
    
    def __init__(self, a, b, c=4, d=99, **kwargs):
        self.d = d
        super(Foo, self).__init__(a=a, b=b, c=c, **kwargs)
        
    def __repr__(self):
        return param.parameterized.pprint(self, {'d':self.d})
```

First the positional arguments are shown as their appropriate values. Then modified keyword arguments are shown in the order they are declared in. Then the ``varargs`` (e.g ``*varargs``) is shown (if present). Then if a keyword dictionary (i.e. **kwargs) is present, it is assumed that the modified parameters can then be listed in order of their precedence.

Here are some examples:

```python
>>> Foo(1,'C')
Foo(1, 'C')

# Positional arguments always shown even if at default values
>>> Foo(4,'B')  
Foo(4,'B')

# Keyword not shown if supplied at default value
>>> Foo(4,'B', c=4) 
Foo(4,'B')

>>> Foo(4, 'B', c=5)
Foo(4, 'B', c=5)

# 'd' is not a parameter (supplied at the default value)
>>> Foo(4,'B', c=5, d=99)   
Foo(4, 'B', c=5)

>>> Foo(4,'B', c=5, d=100)
 Foo(4,'B', c=5, d=100)

>>>  Foo(4,'B', z=44, d=100)
 Foo(4,'B', z=44, d=100)

>>>  Foo(4,'B', z=4, d=100)
Foo(4, 'B', d=100)
```

For comparison, here is an example of the default repr:

```python
>>>Foo(4,'B')
Foo(a=4, b='B', c=4, name='Foo00456', z=4)
```

My main concern with ``pprint`` is performance. It might be nice to cache some of the results of the introspection for performance reasons.

EDIT: If you are happy with the addition of this function, I'll make sure to write a good number of unit tests for it.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1188,8 +1188,8 @@ class Parameterized(object):
                 raise Exception("Argument %r is not a parameter "
                                 "and has an unknown value." % k)
 
-            # Explicit kwarg (unchanged)
-            if (k in kwargs) and kwargs[k] == value: continue
+            # Explicit kwarg (unchanged, known value)
+            if (k in kwargs) and (k in values) and kwargs[k] == values[k]: continue
 
             if k in posargs:
                 # The value repr is used for positional arguments
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -80,6 +80,68 @@ def classlist(class_):
     return inspect.getmro(class_)[::-1]
 
 
+def pprint(obj, including={}, exclude=['self', 'name']):
+    """
+    Pretty printed representation of a parameterized object that may
+    be evaluated with eval. Similar to repr except introspection of
+    the constructor (__init__) ensures a valid and succinct
+    representation is generated.
+
+    Note that any **kwargs argument is assumed to be used for setting
+    parameters and will therefore not be shown.
+
+    Positional arguments are always shown, followed by the explicitly
+    declared keyword arguments (that have been modified) followed by
+    modified parameters, sorted by precedence.
+
+    The including dictionary allows arguments to be supplied that
+    aren't parameters but that should also be pretty printed.
+
+    The exclude list is used to exclude specified parameters from
+    being included in the representation.
+    """
+    changed_params = dict(obj.get_param_values(onlychanged=True))
+    param_values = dict(obj.get_param_values())
+    spec = inspect.getargspec(obj.__init__)
+    args = spec.args[1:] if spec.args[0] == 'self' else spec.args
+
+    if spec.defaults is not None:
+        posargs = spec.args[:-len(spec.defaults)]
+        kwargs = dict(zip(spec.args[-len(spec.defaults):], spec.defaults))
+    else:
+        posargs, kwargs = args, []
+
+
+    ordering = sorted(changed_params.keys(),
+                          key=lambda k: obj.params(k).precedence)
+
+    values = dict(obj.get_param_values())
+    values.update(including)
+
+    arglist, defaults, processed = [], [], []
+    for k in args + ordering:
+        if k in processed: continue
+
+        if k in posargs:
+            # The value repr is used for positional arguments
+            arglist.append(repr(values[k]))
+        elif (k in kwargs) and kwargs[k] != values[k]:
+            # Explicit keyword arguments that have changed
+            arglist.append(k)
+            defaults.append(values[k])
+        elif spec.keywords is not None:
+            # Parameters ordered by precendence (if **kwargs present)
+            arglist.append(k)
+            defaults.append(values[k])
+        processed.append(k)
+
+    return obj.__class__.__name__ + inspect.formatargspec(arglist,
+                                                          varargs=spec.varargs,
+                                                          defaults=defaults)
+
+
+
+
 def descendents(class_):
     """
     Return a list of the class hierarchy below (and including) the given class.
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -129,7 +129,9 @@ def pprint(obj, including={}, exclude=['self', 'name']):
             # Explicit keyword arguments that have changed
             arglist.append(k)
             defaults.append(values[k])
-        elif spec.keywords is not None:
+        elif k in kwargs:
+            continue
+        elif (spec.keywords is not None):
             # Parameters ordered by precendence (if **kwargs present)
             arglist.append(k)
             defaults.append(values[k])
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1168,7 +1168,9 @@ class Parameterized(object):
 
         ordering = sorted(
             sorted(changed_params.keys()), # alphanumeric tie-breaker
-            key=lambda k: self.params(k).precedence)
+            key=lambda k: (- float('inf')  # No precedence is lowest possible precendence
+                           if self.params(k).precedence is None
+                           else self.params(k).precedence))
 
         values = dict(self.get_param_values())
 
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -804,6 +804,8 @@ def script_repr(val,imports, prefix="\n    ", settings=[],
     """
     # CB: doc prefix & settings or realize they don't need to be
     # passed around, etc.
+    # JLS: The settings argument is not used anywhere. To be removed
+    # in a separate PR.
     if isinstance(val,type):
         rep = type_script_repr(val,imports,prefix,settings)
 
@@ -811,10 +813,8 @@ def script_repr(val,imports, prefix="\n    ", settings=[],
         rep = script_repr_reg[type(val)](val,imports,prefix,settings)
 
     elif hasattr(val,'script_repr'):
-        rep=val.script_repr(imports=imports,
-                            prefix=prefix+"    ",
-                            qualify=True,
-                            unknown_value=unknown_value)
+        rep=val.script_repr(imports=imports, prefix=prefix+"    ",
+                            qualify=True, unknown_value=unknown_value)
 
     else:
         rep=repr(val)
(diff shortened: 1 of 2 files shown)


ceball:
> Although I started implementing this in __repr__, I saw the note about repr needing to be fast and 
> then decided against implementing it there once I realized what was required.

The comment is `` # CB: __repr__ is called often; methods it uses should not be too slow``. That comment's misleading, unfortunately (and is one of the ones listed to deal with during my 'comments cleanup' task...not that that will make you any happier :( ). It definitely used to be true that Parameterized's ``__repr__`` was called often by code outside param (e.g. Topographica), so for short-term practical reasons we might have kept ``__repr__`` fast, but the real fix is to stop calling ``__repr__`` so often.

The big issue we had in the past was not calling logging (``Parameterized.debug``, ``.verbose``, etc) methods correctly:

```python
# CEBALERT: designed to avoid any processing unless the print
# level is high enough (e.g. to avoid expensive str(Parameterized
# instance) calls). Not all callers are taking advantage of this
# (either calling str() themselves, resulting in potentially
# expensive operations for things that might never be printed, or
# redundantly using lambda functions to avoid the
# processing. Should fix that.
#
# Note that Python's logging module would simplify print
# statements still further (see "topographica's debug printing"
# emails between CB&JB).
def __db_print(self,level=INFO,*args):
```

I'm sure at one point we found a debug statement called so often that it slowed simulations down, or something like that. We fixed that (probably it happened several times...), but I don't know how much incorrect calling of the logging methods is left across ioam code.

ceball:
Oh dear, param itself is not setting a good example. For instance, a quick search reveals how we're calling ``__repr__`` multiple times every time we create a Parameterized instance:

```python
    def _setup_params(self,**params):
        [...]
        for name,val in params.items():
            desc = self.__class__.get_param_descriptor(name)[0] # pylint: disable-msg=E1101
            if desc:
                self.debug("Setting param %s=%s"% (name, val))
            else:
                self.warning("Setting non-parameter attribute %s=%s using a mechanism intended only for parameters" % (name, val))
            # i.e. if not desc it's setting an attribute in __dict__, not a Parameter
            setattr(self,name,val)
```
(https://github.com/ioam/param/blob/master/param/parameterized.py#L1288)

jlstevens:
Ok. Thanks for the background!

I've discussed this PR with Jim and I'm now working on integrating this new functionality with ``script_repr``. That avoids introducing a new concept while leaving the existing ``__repr__`` alone.

ceball:
I agree about not introducing the extra concept of pretty print. But if we weren't calling ``__repr__`` so often, wouldn't we want to improve it? Apart from speed, wouldn't it be nice to merge this code into ``__repr__``? Or is there something everyone prefers about the existing ``__repr__`` (e.g. how it prints all parameters, not just ones that have been overridden from the default)?

jbednar:
I'd rather the existing repr change to only print overridden parameters; I think that's much more useful and better from the command line and seems like an easy change to make.  But Jean-Luc's changes require the inspect module, which seems like overkill for repr() but is fine for script_repr.

jbednar:
Fixing the bad example from param above seems like an excellent idea; no reason to be slowing things down...

jlstevens:
I would also like this functionality for ``__repr__``. That said, it definitely makes sense for ``script_repr`` and now it is implemented switching ``__repr__`` as well would be as simple as:

```python
def __repr__(self):
    return self.script_repr()
```

My biggest objection now is that I would really like ``script_repr`` to be renamed to ``pprint``: even if you are printing everything, ``script_repr`` is always pretty printing due to nice indentation it introduces. I think pretty printing is a concept people will immediately understand as opposed to 'script repr' which will mean nothing to most people.

jbednar:
I'm fine with pprint, but don't personally want to deal with the headache of changing it.  repr() can be script_repr without the indentation, but probably best not to have indentation on for repr() by default...

ceball:
Ok, to summarize (tell me if I'm wrong):

* This PR could be merged as it is now to give us better ``script_repr``.
* We'd like to rename ``script_repr`` to ``pprint`` but that will be a pain (although isn't it possible to do everything as ``pprint`` and also leave the ``script_repr`` name around pointing at ``pprint`` for a couple of releases?).
* We'd like ``__repr__`` to be like the ``script_repr`` of this PR, but (a) Jim objects that using ``inspect`` is overkill for ``__repr__`` (but Jean-Luc and I don't see the problem) and (b) ``__repr__`` shouldn't have the indentation that ``script_repr`` has.

If that's right, what should we do? Merge this PR to get a better ``script_repr`` for the upcoming release and decide what to do about the other things afterwards? Or should we be more conservative (leave this PR for after the release), or more ambitious (make ``__repr__`` better for the upcoming release)?

jlstevens:
Here are my responses to each of your bullet points:

* Yes, I think this could be merged now. In some ways, the earlier we merge the better as that gives us more time to find potential issues before release.
*  I really would like ``pprint`` as a name. Aliasing ``script_repr`` to ``pprint`` for now does makes sense to me; it doesn't look like the public API is going to be cleaned up for this release so it isn't a terrible thing to do.
* a) My worry about using ``inspect`` (and the full ``pprint`` functionality) in ``__repr__`` is essentially down to how much that might affect performance b) I am happy leaving a ``script_repr`` function that has suitable indentation (and other options) for representing large objects.

I do want to see this merged before the next release and think adding ``pprint`` and aliasing ``script_repr`` to it would be ok. I probably wouldn't replace the existing ``__repr__`` before this release though...

ceball:
I'm fine with you merging (and hence having more chance to find potential issues), and doing the ``script_repr`` aliasing if you have time. We can then improve ``__repr__`` later - and hopefully by then the performance of ``__repr__`` won't matter (because we'll have checked for and fixed any bad calls of ``debug``, ``verbose`` etc from other ioam code).

jlstevens:
Ok..please merge this now, otherwise I'll click the button myself later!

param/parameterized.py
```diff
@@ -810,7 +813,8 @@ def script_repr(val,imports,prefix,settings):
         rep = script_repr_reg[type(val)](val,imports,prefix,settings)
 
     elif hasattr(val,'script_repr'):
-        rep=val.script_repr(imports=imports,prefix=prefix+"    ")
+        rep=val.script_repr(imports=imports, prefix=prefix+"    ",
+                            qualify=True, unknown_value=unknown_value)
```
ceball: Is that supposed to be ``qualify=qualify``?

jlstevens: Yes... I think so.

param/parameterized.py
```diff
@@ -1139,36 +1143,69 @@ def __repr__(self):
...
+                                re.match('^'+self.__class__.__name__+'[0-9]+$', values[k])):
+                continue
+
+            value = script_repr(values[k], imports, prefix=prefix,settings=[],
+                                unknown_value=unknown_value,
+                                qualify=qualify) if k in values else unknown_value
+            if value is None:
+                raise Exception("Argument %r is not a parameter "
+                                "and has an unknown value." % k)
+
```
ceball: Could you clarify this part? It seems to be saying that if ``script_repr(values[k])`` returns ``None``, that means ``k`` is not a parameter of ``self``? I don't understand why getting back ``None`` from ``script_repr`` means ``k`` isn't a parameter. Anyway, I do agree that getting back ``None`` from ``script_repr`` means ``k``'s value is "unknown" (if that means, "has unknown repr")...but I don't agree about raising an exception. Param has ``script_repr_reg``, a place to register handlers, and I can see some of the ones param itself supplies actually return ``None``, e.g.

```python
def empty_script_repr(*args):
    return None

script_repr_reg[random.Random] = empty_script_repr
```

So, I could alter your code above to be ``if value is None: continue``, and I think that would match the previous behavior of ``script_repr`` (untested). However, I think maybe we should change that behavior. We could give a warning whenever we deliberately suppress a ``script_repr``, because people would probably like to be aware of what's being missed from their ``script_repr`` (e.g. random state). This would change the previous behavior of param (which was silently suppressing), but I think I would prefer it. What do you think?

If we did the above, I'm not sure if we'd need to distinguish deliberate suppression (i.e. a deliberate ``None`` coming back from a handler) from the case of the ``script_repr`` function falling all the way through to ``repr`` and still getting back ``None``, because I'm not sure what kind of ``__repr__`` method would return ``None``. Basically, I'm not clear about when you thought ``value`` would be ``None``, above...

jlstevens: I admit I didn't read through all you wrote but I can probably clarify the some of the confusion anyway.

The ``value`` variable  is set with a conditional statement (``... if k in values else unknown_value``). It isn't  the ``script_repr(values[k])`` that will return ``None`` but the specified token for unknown value (default is ``'<?>'``` if I remember correctly) that may be set to ``None``.   In other words, if you don't specify a string token to denote unknown values, you'll get the exception.

Hope that clarifies it a bit.. I'll now try to get my head around the rest of what you wrote! ;-p

ceball: Oops, I totally misread that line and then ran with it and didn't look back! Maybe it's because I first read it on my phone and didn't see the end. I hope so!

So I could get the old behavior via unknown_value (setting it to an empty string if I really wanted to).

ceball: (I'll preface this by saying I'm looking at it on my phone again just in case, but...)

I think i didn't misread it, so I must be confused. If ``k`` is in ``values`` and ``script_repr`` returns ``None`` (which I think it could because of the handlers we have), ``value`` will be ``None``. 

So to me the error message seems wrong (about ``k`` not being a parameter) and raising an exception seems wrong (given that param itself has handlers that will presumably cause that exception).

If I'm not totally confused, it seems like our handlers (e.g. the empty repr one) should return the unknown_value rather than None. I hope it isn't just something like I lost your changes somewhere...I can't check because if I go to a new tab on my phone I'll lose this message ;)

Issue #97, opened 2015-02-24
Overriding parameter value with None in subclass ignored


philippjfr:
This behavior just bit me. Anyone have a clue why this would be happening?

Test case:

```
import param

class A(param.Parameterized):

    a = param.Number(default=10)

    def __init__(self, **params):
        super(A, self).__init__(**params)
        print self.a

class B(A):

    a = param.Number(default=None)

    def __init__(self, **params):
        super(B, self).__init__(**params)
        print self.a

a=A()
b=B()
```

Output:

```
10
10
10
```

philippjfr:
Ahh, seems this is actually expected behavior according to the docstrings.

```
        Ordinarily, when a Python object is instantiated, attributes
        not given values in the constructor will inherit the value
        given in the object's class, or in its superclasses.  For
        Parameters owned by Parameterized classes, we have implemented
        an additional level of default lookup, should this ordinary
        lookup return only None.

        In such a case, i.e. when no non-None value was found for a
        Parameter by the usual inheritance mechanisms, we explicitly
        look for Parameters with the same name in superclasses of this
        Parameterized class, and use the first such value that we
        find.
```

Not sure it is desirable though. Anyone have any suggestions how best to get the behavior I want. There should be some way of getting override with None behavior at least.

ceball:
Inheritance of default values has nagged away at me. For example:

```python
import 
class X(param.Parameterized):
    a = param.Number(default=5.0, bounds=(-10,10))

class Y(X):
    a = param.Number(bounds=(0,10)) # someone just wants smaller bounds

>>> print Y().a
0.0
```

Would it make more sense for the default value of ``Y.a`` to be inherited from ``X.a`` than to come from ``param.Number.default``?

Anyway, back to Philipp's problem. Could we change the logic of param inheritance for the ``default`` slot, so that instead of just looking for the first non-None value of ``default``, we also check to see if the value of ``default`` is different from that of the parent class?

jbednar:
I don't understand -- why does Y.a not already inherit from x.a?  I thought that was the crux of Philipp's problem, that the default value of None makes inheritance happen, but that he did not want that to happen in his case.  I'm getting very confused here.  Yes, we do want Y.a to inherit the value from X.a, even though it will cause Philipp's problem.  I don't understand why Philipp has a problem if we aren't doing that now.

ceball:
The inheritance mechanism assumes the default value of Parameter objects is ``None``.

[...]

PR #98, opened 2015-02-24
Fixed bug preventing overriding inherited parameter with None

philippjfr:
See Issue #97.

Test case:
```
import param

class A(param.Parameterized):

    a = param.Number(default=10)

    def __init__(self, **params):
        super(A, self).__init__(**params)
        print self.a

class B(A):

    a = param.Number(default=None)

    def __init__(self, **params):
        super(B, self).__init__(**params)
        print self.a

a=A()
b=B()
```

Output:
```
10
10
10
```

Fixed output:
```
10
None
None
```

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -5,6 +5,7 @@ messaging.
 
 import copy
 import re
+import sys
 
 from operator import itemgetter,attrgetter
 from types import FunctionType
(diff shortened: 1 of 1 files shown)


philippjfr:
This is the wrong fix so don't merge it. Will either find a way around this or implement it properly.

jbednar:
What does this commit have to do with the issue described in the comment above?

philippjfr:
It doesn't, my fork of param was still linked against this PR so when I pushed a change it automatically updated it. I've since closed it.

PR #105, opened 2015-03-06
Logging: make param's logging interface match python's logging interface.

ceball:


jbednar:
Seems like the right changes to me.

PR #114, opened 2015-04-07
Unicode support for param.String

philippjfr:
This commit adds unicode support for param.String, which has been requested on multiple occasions, see https://github.com/ioam/param/issues/73 and https://github.com/ioam/holoviews/issues/29. This is the simplest possible fix that will get param to support Unicode strings in Python 2.

To handle unicode correctly consistently we should consider importing: ```from future import unicode_literals``` everywhere, but I'm not sure what the implications of that change would be.

The fully verbose version providing compatibility for all string, bytes and unicode types is the following but that's probably overkill:

```
try:
    unicode = unicode
except NameError:
    # 'unicode' is undefined, must be Python 3
    str = str
    unicode = str
    bytes = bytes
    basestring = (str,bytes)
else:
    # 'unicode' exists, must be Python 2
    str = str
    unicode = unicode
    bytes = str
    basestring = basestring
```

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -5,6 +5,7 @@ messaging.
 
 import copy
 import re
+import sys
 
 from operator import itemgetter,attrgetter
 from types import FunctionType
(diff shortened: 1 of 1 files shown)


jbednar:
Seems ok to me.

jlstevens:
I think this is an important thing to support and there is very little code so I will merge it now. Obviously, we can continue to improve it (if necessary) on master.

PR #119, opened 2015-07-06
be compatible with wheel package format

stonebig:


jbednar:
Thanks!  Are the imports of find_packages and Extension necessary?  They don't seem to be used.

stonebig:
Well, maybe not, it was a simple copy/past.
I can rectify

philippjfr:
While almost everyone is guaranteed to have setuptools nowadays, in HoloViews we ended up wrapping the import in a try/except. Could consider doing the same here:

```python
try:
    from setuptools import setup
except ImportError:
    from distutils.core import setup
```

ceball:
I'm out of touch with packaging python projects (and I've always been very confused about it). Can someone persuade me that setuptools is now the right thing to use? E.g. I see that distutils is in the standard library (of python 2 and python 3), but setuptools is not. That doesn't necessarily mean it's a problem to use setuptools, of course, I'm just noting my level of knowledge.

Also, presumably only the person creating the wheel needs setuptools? To install the wheel, you'd just need pip - is that right?

And finally, seems like we should be uploading a wheel to pypi when we do a release - right?

Thanks!

stonebig:
argumentation here: https://github.com/winpython/winpython/issues/95#issuecomment-105617815

stonebig:
by the way I fail to reproduce the "mandelbrot" graphic, not having the mandelbrot.npy file.
How did you do it ? starting from my own sample mandelbrot (http://nbviewer.ipython.org/github/winpython/winpython_afterdoc/blob/master/docs/Winpython_checker.ipynb), I fail to do anything

````
from holoviews import *

%load_ext holoviews.ipython

fractal = image

((fractal * HLine(y=0)).hist() + fractal.sample(y=0))

TypeError                                 Traceback (most recent call last)
<ipython-input-7-625e4c62a556> in <module>()
      4 
      5 
----> 6 ((fractal * HLine(y=0)).hist() + fractal.sample(y=0))

TypeError: unsupported operand type(s) for *: 'int' and 'HLine'


````

jbednar:
Hopefully the replies at the holoviews site have shown how to make the .npy file; here it looks like "fractal" is an integer rather than an instance of hv.Image(), which of course won't work.  I'll delete these comments now as they are about HoloViews, not Param, but please file an issue at the HoloViews site if you have any further problems.

jbednar:
It sounds like we ought to support a wheel, even though it's a pure Python package, because that seems to be friendlier to Windows users.  So, I vote that we make the change that Philipp suggested.  If you submit that as a pull request, I'll merge it!

stonebig:
Do you speak about what I did in last patch proposal ? Or is it in another aspect (like Travis) ?

jbednar:
You can just update the current pull request if you like, because that will help tie the change to this discussion.  Otherwise, we'll just make the change ourself shortly.

stonebig:
Please do it yourself, I'm short on time now.

philippjfr:
Done. I'll close the issue now.

PR #121, opened 2015-08-09
Apply E231 (white space after comma) with autopep8.

ceball:
Not for merging - just demonstration.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -9,7 +9,7 @@ import sys
 import inspect
 import random
 
-from operator import itemgetter,attrgetter
+from operator import itemgetter, attrgetter
 from types import FunctionType
 from functools import partial, wraps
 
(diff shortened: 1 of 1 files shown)


ceball:
I'm not proposing to blindly apply autopep8 (unless autopep8 understands "know when to be inconsistent" from pep8 ;) ), but to apply autopep8 one rule at a time and inspect the results (and then commit, fix up and commit, or abandon and tell flake8 to ignore the rule). (It has to be one rule at a time anyway, because I don't have enough time in one go at the moment to do everything at once.)

philippjfr:
Alternatively I've just let PyCharm apply all the PEP8 rules to the code. You can see what that looks like [here](https://github.com/philippjfr/param/commit/a56118f8bbcaba1fda7cdcbd283df93e424dd17c). Looks sensible except for a few cases, but someone would have to thoroughly review it.

jbednar:
Looks fine to me, both using just E231 and philipp's full PEP8 conversion, but yes, someone would need to check each one.  Grueling, but would be nice to have it all cleaned up.  I personally don't like splitting things across lines so much, as it makes the files so much longer, but it does make sense to have *some* line length limit, and I guess people seem to think 80 characters is the right one.

ceball:
You didn't even give me a year to get this done before closing! ;)

jbednar:
:-)

PR #122, opened 2016-03-29
Enabled cythonizing of param modules if Cython is available

philippjfr:
This small PR enables cythonizing for the param ``__init__``, ``parameterized`` and ``version`` modules. The IPython module has too much magic in it to be supported. Overall this already provides roughly a 50% speedup when creating new parameterized objects (measured by creating 20k HoloViews Images). In future we could go further and actually rewrite some parts of param as actual Cython, but I suspect that may not be worth it given the decent speedup without it and that most of the remaining time is spent deep copying default parameter values, which can in some cases be avoided by setting ``instantiate`` to ``False`` on a parameter.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 python:
   - "2.6"
   - "2.7"
-  - "3.2"
   - "3.3"
   - "3.4"
   - "pypy"
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[build_ext]
+inplace=1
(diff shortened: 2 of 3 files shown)


philippjfr:
Seems Python 3.2 isn't supported by the coverage tool anymore so it's failing. Will have to decide whether we should drop support for 3.2 as well.

jbednar:
I think we can safely drop support for Python 3.2 from the latest Param release, or at least stop testing it.

jbednar:
I'm happy for this to be merged.

philippjfr:
Removed Python 3.2, tests are now passing. Ready to merge whenever.

PR #123, opened 2016-04-07
Add shared parameter context manager

philippjfr:
This PR adds the ``shared_parameters`` context manager. Here's the docstring, which should explain what it does:

Context manager to share parameter instances when creating
multiple Parameterized objects of the same type. Parameter default
values are instantiated once and cached to be reused when another
Parameterized object of the same type is instantiated.
Can be useful to easily modify large collections of Parameterized
objects at once and can provide a significant speedup.

I've done some basic profiling and this can be a significant boost when instantiating a lot of complex objects.

```python
%%timeit
with param.parameterized.shared_parameters():
    hmap = hv.HoloMap({i: hv.Image(np.random.rand(10,10)) for i in range(10000)})
```

```
1 loop, best of 3: 1.77 s per loop
```

vs.

```python
%%timeit
hv.HoloMap({i: hv.Image(np.random.rand(10,10)) for i in range(10000)})
```

```
1 loop, best of 3: 3.86 s per loop
```

There's many cases where this might be useful, particularly holoviews where many operations create a lot of identical objects (e.g. groupby). Sharing a parameter value can also be useful in it's own right since you could change the dimensions of all the elements at once (although I might not recommend it):

```python
with param.shared_parameters():
    hmap = hv.HoloMap({i: hv.Image(np.random.rand(10,10)) for i in range(10000)})
hmap.last.kdims[:] = [hv.Dimension('A'), hv.Dimension('B')]
```

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -22,6 +22,7 @@ from .parameterized import Parameterized, Parameter, String, \
      descendents, ParameterizedFunction, ParamOverrides
 
 from .parameterized import logging_level # pyflakes:ignore (needed for eval)
+from .parameterized import shared_parameters # pyflakes:ignore (needed for eval)
 
 
 # Determine up-to-date version information, if possible, but with a
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -539,6 +539,28 @@ class String(Parameter):
 
 
 
+class shared_parameters(object):
+    """
+    Context manager to share parameter instances when creating
+    multiple Parameterized objects of the same type. Parameter default
+    values are instantiated once and cached to be reused when another
+    Parameterized object of the same type is instantiated.
+    Can be useful to easily modify large collections of Parameterized
+    objects at once and can provide a significant speedup.
+    """
+
+    _share = False
+    _shared_cache = {}
+
+    def __enter__(self):
+        shared_parameters._share = True
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        shared_parameters._share = False
+        shared_parameters._shared_cache = {}
+
+
+
 class ParameterizedMetaclass(type):
     """
     The metaclass of Parameterized (and all its descendents).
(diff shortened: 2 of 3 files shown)


philippjfr:
Appears like ``assertIs`` doesn't work in Python2.6 unit tests. Will fix that tomorrow.

philippjfr:
Since not all parameters are shared something like ``shared_instances`` would probably be better.

jlstevens:
After some discussion, I think this is a useful utility that can really help performance but it needs to be used carefully!

Parameters with simple literals (ints, floats, strings, booleans) can have shared defaults without problem but you do need to be careful when handling things like lists of objects. In such cases, you have to be sure it is ok to have all the defaults change together as all the parameters will be pointing to the same object in memory. 

Previously, I understand that you could declare a default that creates instances (e.g instances in a list) which would result in *new* instances every time the parameterized class was instantiated. Using this context manager, all instances point to the same instances in a single list.

It is all rather subtle and you need to think carefully whether you'll get the behaviour you want when using this context manager. That said, used properly, it will be great to see the performance improvement!

jbednar:
Seems useful to me.  In practice, it seems problematic to apply, though.  I would have imagined doing this inside of HoloMap, but in the above example all of the Image objects get created *before* the HoloMap is constructed, at which point it's too late, hence the context manager you've used here.  Having users have to do this explicitly seems very annoying, but I don't see a way around it. :-(

philippjfr:
My suggestion for holoviews is that we would use it in a few select places where the generated objects are guaranteed to be identical except for their ``.data`` value. The main example being ``Columns.groupby``, where you're creating a lot of individual Elements with subsets of the data. In that example most parameters are already shared because the same parameter values are passed to all Element instances or they are inherited via clone, which also does not copy instances.

jbednar:
But what about the example above, which is a very common idiom (make a dictionary of Elements in user code, then make a HoloMap from it)?  In that case you know very clearly that they will share the same parameter values, yet the actual storage of those values can only be shared if done with the context manager?  Seems painful to systematically miss out on this opportunity, but it's also painful to ask people to use a context manager like this. :-(

philippjfr:
I think the only situations where this is likely to matter much, you'll be dealing with a very large dataset and the ``groupby`` method will be the preferred way to construct objects, e.g. via the conversion interface. In those cases you'll either have a very large table in a database or a dense n-dimensional array on disk (e.g. iris/xarray), which you can then flexibly expand into different Element types. Otherwise you should simply be using a DynamicMap. I also have ideas about making ``groupby`` optionally be dynamic, avoiding creating this many objects altogether.

jbednar:
Seems to me like it's easy enough to construct a large HoloMap just by exploring a parameter space using Elements directly, without creating a cube or other large data structure.  But I don't have any suggestions for making that situation work well, so I'll leave it for now. :-/

jbednar:
In any case, merging this is fine whenever you are ready.

philippjfr:
Just fixed the tests for Python 2.6, ready to merge now.

PR #124, opened 2016-04-07
Skip processing non-parameterized classes in setup_param

philippjfr:
Very minor addition, which gives an easy 5-10% speedup in creating Parameterized objects. Between this, the ``shared_defaults`` context manager, and Cython we can now squeeze out a 3x speedup when creating HoloViews Image Elements.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1392,6 +1392,8 @@ class Parameterized(object):
         #  a later-overridden parent class's parameter)
         params_to_instantiate = {}
         for class_ in classlist(type(self)):
+            if not issubclass(class_, Parameterized):
+                continue
             for (k,v) in class_.__dict__.items():
                 # (avoid replacing name with the default of None)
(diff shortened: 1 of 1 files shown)


jbednar:
This looks like a safe change to make, but I'm curious about when it would have any effect.  Is this designed to skip the "object" base class from all Parameterized classes, or is it for mixin non-Parameterized classes that are alongside the Parameterized ones in the hierarchy for this class?  Basically, when does a non-Parameterized class come up as a subclass of a Parameterized class?

philippjfr:
I'm not quite sure where the speedup is coming from when instantiating objects which just inherit from Parameterized but I can consistently reproduce it. There are also a few places in HoloViews where we have unparameterized baseclasses and mixins (e.g. AttrTree), which don't need to be processed.

philippjfr:
Straightforward optimization, which has been ready to merge for a long time. Merging.

PR #127, opened 2016-07-12
Ordered selection

jbednar:
Previously, options in a selection widget created from a Selector parameter would be in an arbitrary order, determined by the underlying Python dictionary type.  With this PR, the original supplied order of options should be preserved.

Under Python 2.6 requires installing the `ordereddict` package, but otherwise should support all the same Python versions as before.


jlstevens:
Looks good. Merging.

PR #130, opened 2016-08-13
Check there's a working compiler; see #129.

ceball:
Before:
```
ceball@spheroid:~$ docker run --rm continuumio/anaconda:2.3.0 bash -c 'pip install param'
Collecting param
  Downloading param-1.4.1.zip (472kB)
Building wheels for collected packages: param
  Running setup.py bdist_wheel for param
  Complete output from command /opt/conda/bin/python -c "import setuptools;__file__='/tmp/pip-build-6WGJpP/param/setup.py';exec(compile(open(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" bdist_wheel -d /tmp/tmprwtYGWpip-wheel-:
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.linux-x86_64-2.7
  creating build/lib.linux-x86_64-2.7/param
  copying param/__init__.py -> build/lib.linux-x86_64-2.7/param
  copying param/parameterized.py -> build/lib.linux-x86_64-2.7/param
  copying param/ipython.py -> build/lib.linux-x86_64-2.7/param
  copying param/version.py -> build/lib.linux-x86_64-2.7/param
  creating build/lib.linux-x86_64-2.7/numbergen
  copying numbergen/__init__.py -> build/lib.linux-x86_64-2.7/numbergen
  running build_ext
  building 'param.__init__' extension
  creating build/temp.linux-x86_64-2.7
  creating build/temp.linux-x86_64-2.7/param
  gcc -pthread -fno-strict-aliasing -g -O2 -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -I/opt/conda/include/python2.7 -c param/__init__.c -o build/temp.linux-x86_64-2.7/param/__init__.o
  unable to execute 'gcc': No such file or directory
  error: command 'gcc' failed with exit status 1
  
  ----------------------------------------
  Failed building wheel for param
Failed to build param
Installing collected packages: param
  Running setup.py install for param
    Complete output from command /opt/conda/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip-build-6WGJpP/param/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-8Rv9Wg-record/install-record.txt --single-version-externally-managed --compile:
    running install
    running build
    running build_py
    running build_ext
    building 'param.__init__' extension
[...]

PR #131, opened 2016-08-13
Skip flake8 tests on Python 2.6

jbednar:
flake8 uses OrderedDict, not present in 2.6, so we should skip it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,6 @@ install:
 
 script:
   - nosetests --with-doctest --with-coverage --cover-package=param
-  - if [[ $TRAVIS_PYTHON_VERSION != 2.6]]; flake8 --statistics --exit-zero --ignore=E231 param numbergen
+  - if [[ $TRAVIS_PYTHON_VERSION =~ 2.6]]; flake8 --statistics --exit-zero --ignore=E231 param numbergen
 
 after_success: coveralls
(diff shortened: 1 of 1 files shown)


jbednar:
@ceball; I give up!  Do you have any idea how to express "skip flake8 on Python 2.6"?  I tried a couple of guesses in the two commits in this PR, but neither is apparently valid syntax, and I can't tell what language this is supposed to be in...

ceball:
While I'm waiting for travis, what I didn't understand in my recent PR (#130) is why my change caused a previously passing thing to fail. 

I can only think that maybe flake8 changed (to require OrderedDict) between the previous commit and mine?

ceball:
> Do you have any idea how to express "skip flake8 on Python 2.6"?

No ;)

ceball:
I sidestepped the problem by specifying an older release of flake8 for python 2.6.

I feel like it might be better to run flake8 just for "current" python 2 and python 3, but I'm not sure. 

I also wonder if it would be reasonable to just drop python 2.6 tests now?

Of course, before any of that, we should make flake8's output useful at all by suppressing some messages...

jbednar:
I'm happy with this approach; seems reasonable.  But yes, I was hoping to run flake8 only on latest Python 2 and 3, because we really are never going to want to fix a flake problem found only on 2.6.

PR #135, opened 2016-10-10
Version fixes and improvements

jlstevens:
Fixes a bug in ``param.Version`` when using pip to install from archive i.e:

```bash
pip install https://github.com/ioam/holoviews/archive/master.zip
```

The SHA was being set correctly via the ``'$Format'`` string but wasn't displaying appropriately (unless the source happened to also be in a dirty git repository).

In addition, users are used to looking at ``holoviews.__version__`` directly instead of ``str(holoviews.__version__)`` which is why I've made the ``repr`` match the string representation.

### DIFF
--- a/param/version.py
+++ b/param/version.py
@@ -233,8 +233,10 @@ class Version(object):
         """
         if self.release is None: return 'None'
         release = '.'.join(str(el) for el in self.release)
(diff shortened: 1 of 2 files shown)


jlstevens:
Tests are failing so I'll now check whether there is still a bug or whether I just need to update the tests.

PR #143, opened 2017-02-13
Handled unicode values in ObjectSelector.get_range

philippjfr:
Correctly handle unicode ObjectSelector values in python2. Original code comes from https://github.com/ioam/paramnb/pull/24, but is required here because current unicode handling is broken.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -17,6 +17,7 @@ Parameters and Parameterized classes.
 """
 
 import os.path
+import sys
 import glob
 
 from .parameterized import Parameterized, Parameter, String, \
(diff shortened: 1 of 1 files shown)


philippjfr:
Previously broken example:

```python
import param

chars = [ch.decode('utf-8') for ch in ['❤' , '☀', '☆', '☂', '☻', '♞', '☯', '☭', '☢']]

    test = param.ObjectSelector(objects=chars)
    
Test().params('test').get_range()
```

jbednar:
Looks good, except for the failing test.  Looks like it's an issue of using ".major" under Python 2.6, and that you can get around it by accessing a fixed spot in the tuple instead:

https://docs.python.org/2/library/sys.html#sys.version_info

PR #144, opened 2017-02-25
Added Color, Date and Range parameters

philippjfr:
* Color: Expects an RGB hex string with optional # prefix
* Date: Has to be datetime or np.datetime64 type

The Date parameter could support bounds like Number but I'm not sure how important that is.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1494,18 +1494,28 @@ class MultiFileSelector(ListSelector):
         return abbreviate_paths(self.path,super(MultiFileSelector, self).get_range())
 
 
-class Date(Parameter):
+class Date(Number):
     """
     Date parameter of datetime type.
     """
 
+    def __init__(self, default=None, allow_None=False, **kwargs):
+        allow_None = True if default is None else allow_None
+        super(Date, self).__init__(default=default, allow_None=allow_None, **kwargs)
+        self._check_value(default)
+
     def _check_value(self,val):
+        """
+        Checks that the value is numeric and that it is within the hard
+        bounds; if not, an exception is raised.
+        """
+        if self.allow_None and val is None:
+            return
+
         if not isinstance(val, dt_types) and not (self.allow_None and val is None):
             raise ValueError("Date '%s' only takes datetime types."%self._attrib_name)
 
-    def __set__(self,obj,val):
-        self._check_value(val)
-        super(Date,self).__set__(obj,val)
+        self._checkBounds(val)
 
 
 class Color(Parameter):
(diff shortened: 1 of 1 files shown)


philippjfr:
@jbednar Added bounds validation for Date along with a Range parameter which also has bounds validation.

param/__init__.py
```diff
@@ -1483,3 +1494,117 @@ def get_range(self):
...
 
+class Date(Number):
+    """
+    Date parameter of datetime type.
+    """
+
+    def __init__(self, default=None, allow_None=False, **kwargs):
+        allow_None = True if default is None else allow_None
+        super(Date, self).__init__(default=default, allow_None=allow_None, **kwargs)
+        self._check_value(default)
```
jbednar: Isn't all of this the default behavior anyway?  I can't see what required there to be a constructor here, but there must be some part of this that doesn't happen already?

param/__init__.py
```diff
@@ -1483,3 +1494,117 @@ def get_range(self):
...
+
+    def _check_value(self,val):
+        if (self.allow_None and val is None):
+            return
+        if not isinstance(val, String.basestring):
+            raise ValueError("Color '%s' only takes a string value."%self._attrib_name)
+        if not re.match('^#?(([0-9a-fA-F]{2}){3}|([0-9a-fA-F]){3})$', val):
+            raise ValueError("Color '%s' only accepts valid RGB hex codes."
+                             % self._attrib_name)
```
jbednar: Would it be too confusing to accept a numeric triple when it's set, converting it to the RGB hex string required?  The value would change from what the user thought it was; maybe too tricky.

param/__init__.py
```diff
@@ -1483,3 +1494,117 @@ def get_range(self):
...
+
+
+
+class Range(NumericTuple):
+    """Range parameter represents a numeric range with option bounds and softbounds."""
```
jbednar: "A numeric range with optional bounds and softbounds"

param/__init__.py
philippjfr: > The value would change from what the user thought it was; maybe too tricky.

Right, thought about this but didn't want to magically change the value.

param/__init__.py
philippjfr: Yes, I think this isn't needed, I originally thought it was.

param/__init__.py
jbednar: Ok.  We can always extend it later to support this if we really need it somewhere.

param/__init__.py
philippjfr: It seems a default of None doesn't force allow_None to True on Number types.

philippjfr: Nevermind, Number defined a default of 0.0.

jbednar: I just checked, and I think it's all covered, except that in the superclass, _check_value is only run if the value isn't callable.  Avoiding that check is what allows the values to be dynamic, so I *think* deleting the constructor is the right thing to do here.

PR #146, opened 2017-02-27
ObjectSelector lists possible objects on error

philippjfr:
The fact that ObjectSelector doesn't report what the valid values are has annoyed me and would help greatly with usability in holoviews. Is reporting the first 20 items too many?

Here's an example from HoloViews:

>ValueError: left not in Parameter xaxis's list of possible objects, valid options include [top, bottom, bare, top-bare, bottom-bare, None]

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1029,8 +1029,15 @@ class ObjectSelector(Selector):
                 attrib_name = self._attrib_name
             except AttributeError:
                 attrib_name = ""
-            raise ValueError("%s not in Parameter %s's list of possible objects" \
-                             %(val,attrib_name))
+
+            items = self.objects
+            limiter = ']'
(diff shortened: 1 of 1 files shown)


jbednar:
Showing the items sounds fine to me, but I don't think the number of items would be the thing to limit. E.g. I have a ListSelector object in one nb that chooses between a bunch of Bokeh palettes, each of which is 256 colors long, so 20*256*8 characters will be over 40,000 characters long.  Maybe choose the first 20, then chose the first 200 characters or so?

philippjfr:
>Maybe choose the first 20, then chose the first 200 characters or so?

Seems like just basing it on the number of chars is the best options so I went with that.

jlstevens:
Very helpful! Thanks!

PR #149, opened 2017-04-22
Fixed ClassSelector error handling

philippjfr:
ClassSelector allows specifying multiple classes as a tuple because isinstance and issubclass support it. However the error handling code does not handle this condition correctly so you get a bad exception message. This PR fixes the exception message and adds unit tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ notifications:
 install:
   - pip install coveralls flake8
   - if [[ $TRAVIS_PYTHON_VERSION == 2.6 ]]; then pip install ipython==1.0 ordereddict flake8==2.6; fi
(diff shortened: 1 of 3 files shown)


jbednar:
Looks like the test should be skipped on python2.6?  Otherwise happy to merge it.

jlstevens:
I would appreciate it if this could be merged ASAP as I've written a few unit tests that rely on this being fixed.

philippjfr:
>as I've written a few unit tests that rely on this being fixed.

So you'd require a release to update travis?

jlstevens:
Good point. If my miscellaneous fixes PR for HoloViews were merged, I would want param master used on travis.

PR #150, opened 2017-04-26
Added working ParamOverrides get and contains methods

philippjfr:
The fact that these two methods don't work caught me out and is currently causing a bug in the released HoloViews 1.7.0. Ideally we'd merge this and release param 1.5.1 immediately.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1802,6 +1802,15 @@ class ParamOverrides(dict):
         else:
             dict.__setattr__(self,name,val)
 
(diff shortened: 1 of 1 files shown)


jlstevens:
Those two methods look very reasonable to me and I would be happy to release param now if @jbednar and @ceball  support the idea. The changelog would be small but for param it normally is - and that is a good thing!

philippjfr:
@jbednar Would you mind merging asap?

PR #153, opened 2017-06-12
Serialize / deserialize param objects: yaml and json

PeterDSteinberg:
*EDIT: A couple minor changes to the snippets below*

This PR does JSON / YAML IO for `param` objects.  I'll add some tests shortly.  In the meantime, to try this PR out:

Make `param` objects like this:
```
import json

import param
import yaml


class ServiceMetadata(param.Parameterized):
    display_name = param.String()
    description = param.String()
    geographical_area = param.String()
    bounding_boxes = param.String()
    service_type = param.String()
    geom_type = param.String()
    datatype = param.String()
    file_format = param.String()
    parameters = param.List()
    unmapped_parameters_available = param.Boolean()


class Features(param.Parameterized):
    file = param.String()
    format = param.String()


class Datasets(param.Parameterized):
    mapping = param.String()
    save_folder = param.String()


class Service(param.Parameterized):
    service_folder = param.List()
    service_id = param.String()
    metadata = param.ObjectSelector(default=ServiceMetadata())
    features = param.ObjectSelector(default=Features())
    datasets = param.ObjectSelector(default=Datasets())


class Organization(param.Parameterized):
    abbr = param.String()
    name = param.String()


class ProjectMetadata(param.Parameterized):
    display_name = param.String()
    description = param.String(allow_None=True)


class ConfigSpec(param.Parameterized):
    metadata = ProjectMetadata()
    env = param.Dict()
    cmd_args = param.Dict()
    services = param.List(class_=Service, instantiate=True)
```

Make this `yaml` file:
```
metadata:
  description: Local Copy of Terrain VITD Data for Iraq
  display_name: Iraq Terrain VITD Data
  organization: {abbr: NGA, name: National Geospatial Agency}
name: iraq-vitd
services:
- datasets: {mapping: <feature>.vpf, save_folder: iraq-vitd}
  features: {file: mbr.txt, format: mbr-csv}
  metadata:
    bounding_boxes:
    - [38.782, 28.833, 49.433, 37.358]
    datatype: vitd
    description: null
    display_name: VITD Datasets
    file_format: vpf
    geographical area: Iraq
    geom_type: polygons
    parameters: [vitd]
    service_type: geo-discrete
    unmapped_parameters_available: false
  service_folder: [VITD/obs, VITD/sdr, VITD/slp, VITD/smc, VITD/trn, VITD/veg]
  service_id: all

```
Name that file `example.yml`, then do:
```
import glob
import param.io
fnames = glob.glob('example.yml')
c = param.io.from_yaml(ConfigSpec(), fnames[0])
param.io.to_yaml(c, 'output_check.yml')
```

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -27,7 +27,7 @@ from .parameterized import Parameterized, Parameter, String, \
 
 from .parameterized import logging_level # pyflakes:ignore (needed for eval)
 from .parameterized import shared_parameters # pyflakes:ignore (needed for eval)
-
+from .io import *
 try:
    from collections import OrderedDict
 except ImportError:
@@ -746,7 +746,7 @@ class Integer(Number):
 
     def __init__(self,default=0,**params):
        Number.__init__(self,default=default,**params)
-    
+
     def _check_value(self,val):
         if self.allow_None and val is None:
             return
@@ -1150,6 +1150,7 @@ class List(Parameter):
         self.bounds = bounds
         Parameter.__init__(self,default=default,instantiate=instantiate,
                            **params)
+        #self.class_ = class_
         self._check_bounds(default)
 
     # Could add range() method from ClassSelector, to allow
--- a/param/io.py
+++ b/param/io.py
@@ -75,7 +75,7 @@ def serialize(parameter):
                         d[key] = {}
                     d = d[key]
                 d[current_key[-1]] = parameter
-            return output
+            yield output
 
     list(switch_data_types(parameter, ()))
     return dict(output)
--- a/param/io.py
+++ b/param/io.py
@@ -1,0 +1,94 @@
+from functools import partial
+import json
+
+import param
+
+__all__ = ['serialize', 'deserialize',
+           'to_yaml', 'to_json',
+           'from_yaml', 'from_yaml',]
+
+def from_params_dict(parameter, obj=None):
+    if hasattr(parameter, 'params'):
+        for k, p in parameter.params().items():
+            if k in obj:
+                v = obj[k]
+                args = (p, v)
+            else:
+                continue
+            if isinstance(p, param.List):
+                if getattr(p, 'class_'):
+                    setattr(parameter, k, [from_params_dict(p.class_(), vi) for vi in v])
+                else:
+                    p = v
+            elif hasattr(p, 'params'):
+                setattr(parameter, k, from_params_dict(*args))
+            else:
+                p = v
+    else:
+        parameter = obj
+    return parameter
+
+
+def deserialize(parameter, fname, ftype='yaml'):
+    with open(fname) as f:
+        if ftype == 'yaml':
+            import yaml
+            obj = yaml.safe_load(f.read())
+        elif ftype == 'json':
+            obj = json.load(f)
+        else:
+            raise ValueError('Expected ftype in ("yaml", "json")')
+    from_params_dict(parameter, obj)
+    return parameter
+
+from_yaml = partial(deserialize, ftype='yaml')
+from_json = partial(deserialize, ftype='json')
+
+def serialize(parameter):
+
+    output = {}
+    def handle_dict_type(parameter, current_key):
+        for key in parameter.params():
+            value = getattr(parameter, key)
+            for item in switch_data_types(value, current_key + (key,)):
+                yield (current_key, item)
+
+    def handle_list_type(parameter, current_key):
+        for item in parameter:
+            for item in switch_data_types(item, current_key):
+                yield (current_key, item)
+
+    def switch_data_types(parameter, current_key):
+        if hasattr(parameter, 'params'):
+            for item in handle_dict_type(parameter, current_key):
+                yield item
+        elif isinstance(parameter, (list, param.List)):
+            for item in handle_list_type(parameter, current_key):
+                yield item
+        elif isinstance(parameter, param.Parameterized):
+            tuple(switch_data_types(parameter, current_key))
+        else:
+            if current_key:
+                d = output
+                for key in current_key[:-1]:
+                    if key not in d:
+                        d[key] = {}
+                    d = d[key]
+                d[current_key[-1]] = parameter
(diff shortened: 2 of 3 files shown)


PeterDSteinberg:
@jbednar Should `yaml` be an optional requirement?  (Writing some tests now and will make sure `yaml` is optional in tests if optional in code.)

ceball:
Thanks for submitting this. I'm currently reviewing the code, but doing it in bits (I'm just returning to work after a baby :) ). I hope you'll be able to tell when the review's done (I haven't used github code review before).

> Should yaml be an optional requirement

Yes

jbednar:
> Should yaml be an optional requirement

We definitely cannot add any new required dependencies to param, so that it can be used without any external libraries beyond what's in the standard library.  So if it's not in the stdlib, then it has to be imported just when that bit of code is actually used.

PeterDSteinberg:
I'll work on this tomorrow more - I'll start by making `yaml` optional (optional in tests as well).  There are a few other issues as well I should resolve and add about 2 more small tests.  @ceball Congrats on the baby! It may be more efficient for your time to wait about 2 days to review this PR.

ceball:
Ah, ok, I didn't realize you were still working on it. (I also didn't realize my in-progress review would show up before I click on submit!) I'll stop looking for now, but here are some things that occurred to me as I read through the code:

1. I'd prefer not to add a new file to param if possible. Ideally we would be able to shrink the code in this PR so that it'll fit into the existing file :)

2. You have a big 'switch data type' section. Maybe it would be clearer to use a registry, as is done for e.g. ``script_repr`` (a feature you are unlikely to be aware of - sorry). See https://github.com/ioam/param/blob/master/param/parameterized.py#L820 (``script_repr``, ``script_repr_reg``).

3. I noticed you handle lists, but maybe you need to consider handling containers more generally? Or maybe I missed the point of that code (I just skimmed it - I might have been confused).

Thanks

PeterDSteinberg:
@ceball - Good feedback.  Regarding your points.  1) I'll work on making it more succinct and consistent with existing styles, 2) I will look into the registry functions you mention - still coming up to speed on this project, 3) I agree container types need to be handled more generally and will figure out the least brittle way of checking for the container types when recursing and handling them as needed.

ceball:
(I just submitted the review comments that I had from before you asked me to wait.)

param/__init__.py
```diff
@@ -1150,6 +1150,7 @@ def __init__(self,default=[],class_=None,instantiate=True,
         self.bounds = bounds
         Parameter.__init__(self,default=default,instantiate=instantiate,
                            **params)
+        #self.class_ = class_
```
PeterDSteinberg: Right - thanks!

param/__init__.py
```diff
@@ -27,7 +27,7 @@
 
 from .parameterized import logging_level # pyflakes:ignore (needed for eval)
 from .parameterized import shared_parameters # pyflakes:ignore (needed for eval)
-
+from .io import *
```
PeterDSteinberg: My opinion on * imports is that they are okay if the module being imported uses `__all__`, but if not using `__all__` then * imports are a bad idea.  I can see how they lead to less clear code and am happy to defer - I'll fix them.  I will try to get this new material of the PR as small as possible.  Are you sure we want to get rid of the new `io.py` file entirely?  It seems like a logical separation of concerns to put IO assets in a separate file.

param/io.py
```diff
@@ -0,0 +1,99 @@
...
+                    d = d[key]
+                d[current_key[-1]] = parameter
+            yield output
+
+    list(switch_data_types(parameter, ()))
+    return dict(output)
+
+
+def to_yaml(parameter, fname):
+    import yaml
```
PeterDSteinberg: Good idea - that will make it more clear.  On `ImportError` I'll probably define `yaml` to be `None` in case a function has to check for whether it is available/imported.

param/io.py
```diff
@@ -0,0 +1,99 @@
...
+    converted = serialize(parameter)
+    with open(fname, 'w') as f:
+        f.write(yaml.safe_dump(converted))
+
+def to_json(parameter, fname):
```
PeterDSteinberg: Good idea - I'll make the following functions:
 * `yaml_loads` - make a Parameter object out of yaml string
 * `yaml_load` - same but a yaml file name or file handle
 * `json_loads` - make a Parameter object out of a json string
 * `json_load` - same but a json file name or file handle

param/__init__.py
jbednar: Yes, it's a logical separation of concerns to put IO in a separate file, but param is a bit of a special case, because it's a library that we expect at least some users to include directly into their repo, and I think having a small total number of files will help that.  If there were lots of independent chunks that could be mixed and matched, having separate files would help that, but I don't think there are, so making IO separate would just make things less obvious for users who want to do that.

PeterDSteinberg:
@ceball @jbednar Currently this IO PR is written in a way where the user calls standalone functions to dump param objects to file.  I'll also add dumping to string in `yaml` or `json` format as discussed above.  Should I also add methods of param objects so one can do this:

```
class ServiceMetadata(param.Parameterized):
    display_name = param.String()
    description = param.String()
    unmapped_parameters_available = param.Boolean()

s = ServiceMetadata()
s.yaml_dumps()
s.yaml_dump('metadata.yml')
s.yaml_dump(open('metadata.yml', 'w'))
s.json_dumps()
s.json_dump('metadata.json')
s.json_dump(open('metadata.json', 'w'))
```
Let me know if you think any or all of the bottom 6 lines of that snippet should be possible and if you have any naming changes in mind.

param/__init__.py
PeterDSteinberg: @jbednar That makes sense to me.  I didn't realize we planned around that style of usage (putting `param` in a dif't repo directly), but I can see why people will use it.  I'll get rid of `io.py`

tonyfast:
closing in favor #414

PR #163, opened 2017-06-24
Allow get_param_values() to work on class (fixes #162).

### LINKED ISSUE #162: Allow get_param_values() to work for Parameterized classes?
I think it would be reasonable for [Parameterized.get_param_values()](https://github.com/ioam/param/blob/master/param/parameterized.py#L1444) to work for a class as well as an instance. What do you think?

```
>>> class A(param.Parameterized):
...    x = param.Number(1)

>>> a = A()
>>> a.get_param_values()
[('name', 'A00009'), ('x', 1)]
```

```
>>> A.get_param_values()
[...]
TypeError: get_param_values() missing 1 required positional argument: 'self'
```

Should be a straightforward change to use ``@bothmethod``. I guess ``onlychanged=True`` would only make sense for an instance, but I don't think it matters much.

ceball:


jbednar:
Looks good, thanks.

PR #168, opened 2017-07-05
Added DateRange parameter.

ceball:


param/__init__.py
```diff
@@ -1630,3 +1630,29 @@ def _checkBounds(self, val):
...
+                    raise ValueError("DateRange '%s' only takes datetime types: %s"%(self._attrib_name,val))
+
+            start, end = val
+            if not end >= start:
+               raise ValueError("DateRange '%s': end date %s is before start date %s."%(self._attrib_name,val[1],val[0]))
+        
+        # Calling super(DateRange, self)._check(val) would also check
+        # values are numeric, which is redundant, so just call
+        # _checkBounds().
+        self._checkBounds(val)
```
ceball: There'd be no harm in calling ``super(DateRange, self)._check(val)`` so I wouldn't be against changing this.

param/__init__.py
```diff
@@ -1630,3 +1630,29 @@ def _checkBounds(self, val):
...
                     raise ValueError("Parameter '%s' %s bound must be in range %s"
                                      % (self._attrib_name, bound, self.rangestr()))
+
+
+class DateRange(Range):
+    """
+    A date range specified as (start_date, end_date).
+
+    Dates must be specified as datetime-like types (see
+    param.dt_types).
```
jbednar: Should specify that times are also allowed, right?

param/__init__.py
```diff
@@ -1630,3 +1630,29 @@ def _checkBounds(self, val):
...
+    """
+    def _check(self,val):
+        if self.allow_None and val is None:
+            return
+        
+        if not (self.allow_None and val is None):
```
jbednar: Seems like this if statement is always going to be true when it's reached, and can be removed, unindenting the rest of this function.

param/__init__.py
jbednar: No opinion.

param/__init__.py
ceball: Ha, true. Just testing.

param/__init__.py
ceball: I'm going to deal with naming/doc of date vs datetime in a separately (because applies to existing Date parameter also).

jbednar:
Ready to merge when tests pass.

param/__init__.py
ceball: Hmm, turns out that by not calling super(), tuple length isn't checked - so super() ought to be called. Probably like this:

```
    def _check(self,val):
        if self.allow_None and val is None:
            return

        super(DateRange, self)._check(val)
       
        for n in val:
            if not isinstance(n, dt_types):
                raise ValueError("DateRange '%s' only takes datetime types: %s"%(self._attrib_name,val))

        start, end = val
        if not end >= start:
           raise ValueError("DateRange '%s': end date %s is before start date %s."%(self._attrib_name,val[1],val[0]))
```

However, I implemented DateRange by inheriting from Range, which inherits from NumericTuple. A datetime is not a number according to param (``_is_number()``) and python (``numbers.Number``). I guess I was thinking of 'simple' time since epoch when I was doing this.

ceball: Hmm, param also has a Date parameter, which expects a datetime but is a subclass of Number.

Would it be a bad idea to modify param's ``is_number()`` to accept datetime as a number, or should we rethink datetime handling in param?

Issue #170, opened 2017-07-08
Why does set_hook exist on Number?


ceball:
If having it at all, why not have for other parameter types?

ceball:
#314

Issue #160, opened 2017-06-16
Expressing dependencies between parameters


ceball:
(Attempt to capture notes/plan from PR, JB, JLS, CB - feel free to edit and/or comment.)

Users of param GUI widget toolkits often want to be able to control the range or type of some widgets based on the values of others.

For example, consider specification of a date range: one might want to allow a user either to select start date and end date, or to select a start date and duration. Or the value in one numeric widget might affect the range of another slider. Or a widget should be greyed out or not depending on the value of another widget.

Beyond GUI toolkits, being able to express dependencies between parameters seems like something that might be useful to any user of param. Many of us have occasionally felt the need for linking parameters while using param in the real world, but we've never actually needed it enough to justify the added complexity, and have solved our problems a different way.

We should therefore consider at what level we should support expressing dependencies between parameters.

# Solve at param level?

It's likely that many of the dependencies desired by GUI widget users could be expressed at the domain level (i.e. using param), so it could be valuable for param to support expressing those dependencies in a general way, such as by having parameters register that they subscribe to other parameters' values (being careful to avoid a vicious circle) and allowing code to be registered to respond to such changes. Solving this at the param level would handle a large class of problems across all the widget toolkits, so could be very beneficial. However, it has the danger of making Param impossible to understand (which data flows                                                            where, when?) and maintain, if not done well.

# Solve at param GUI toolkit level?

Since it's not yet clear if the above can be done well, or would inherently have the problems described above when sufficiently advanced to cover the typical real-world cases where these issues arise, an alternative solution would be to extend the toolkit-specific classes like paramnb.Widgets() to allow an update method to be registered that gets called whenever a parameter is changed. Such a method would then check for changes in parameters and apply changes to others that depend on it. This seems straightforward to implement, and would allow us to discover exactly what it is we might want to support in param.

However, this solution has several disadvantages over the param-level solution:

* Doesn't help non-gui users and would have to be implemented for each widget toolkit.

* The GUI code would have to know about the underlying objects. Although a convention could be adopted whereby if specific parameter names (e.g. start_date, end_date, duration) are detected, the appropriate widgets and logic are automatically used (ensuring                                                           that all three of those values are set consistently regardless of                                                           how the data was input by the user).

 # Proposal

We propose to solve at the GUI toolkit level, gather real-world use cases together as a set of demos, and then consider again about implementing at the param level. We take the risk that we end up just using the GUI level solutions forever, and code based on them spreads everywhere.

## Additional note: megawidgets

[...]

PR #178, opened 2017-07-20
[WIP] Defined pre_set_hooks and post_set_hooks for Parameters

jlstevens:
This PR is a prototype that replaces ``set_hook`` on ``Number`` with a more general system of pre- and post- setting hooks. This may be useful for things like linking streams or eventual unit support in param.

An example of what this might be useful for is shown in  ioam/holoviews#1740.

I'm not quite sure about the appropriate signatures for the hooks right now and this PR will probably need quite a bit of discussion before we are agreed on the appropriate approach.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -573,7 +573,7 @@ class Number(Dynamic):
       AB = Number(default=0.5, bounds=(None,10), softbounds=(0,1), doc='Distance from A to B.')
     """
 
-    __slots__ = ['bounds','_softbounds','inclusive_bounds','set_hook']
+    __slots__ = ['bounds','_softbounds','inclusive_bounds']
 
     def __init__(self,default=0.0,bounds=None,softbounds=None,inclusive_bounds=(True,True),**params):
         """
@@ -582,8 +582,6 @@ class Number(Dynamic):
         Non-dynamic default values are checked against the bounds.
         """
         super(Number,self).__init__(default=default,**params)
-
-        self.set_hook = identity_hook
         self.bounds = bounds
         self.inclusive_bounds = inclusive_bounds
         self._softbounds = softbounds
@@ -609,8 +607,6 @@ class Number(Dynamic):
         Also applies set_hook, providing support for conversions
         and transformations of the value.
         """
-        val = self.set_hook(obj,val)
-
         if not callable(val): self._check_value(val)
         super(Number,self).__set__(obj,val)
 
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -373,7 +373,8 @@ class Parameter(object):
     # persistent storage pickling); see __getstate__ and __setstate__.
     __slots__ = ['_attrib_name','_internal_name','default','doc',
                  'precedence','instantiate','constant','readonly',
-                 'pickle_default_value','allow_None']
+                 'pickle_default_value','allow_None',
+                 'pre_set_hooks', 'post_set_hooks']
 
     # When created, a Parameter does not know which
     # Parameterized class owns it. If a Parameter subclass needs
@@ -411,6 +412,8 @@ class Parameter(object):
         self._set_instantiate(instantiate)
         self.pickle_default_value = pickle_default_value
         self.allow_None = (default is None or allow_None)
+        self.pre_set_hooks = []
+        self.post_set_hooks = []
 
 
     def _set_instantiate(self,instantiate):
@@ -474,6 +477,10 @@ class Parameter(object):
         """
         # NB: obj can be None (when __set__ called for a
         # Parameterized class)
+
+        for hook in self.pre_set_hooks:
+            val = hook(obj, val)
+
         if self.constant or self.readonly:
             if self.readonly:
                 raise TypeError("Read-only parameter '%s' cannot be modified"%self._attrib_name)
@@ -490,6 +497,9 @@ class Parameter(object):
             else:
                 obj.__dict__[self._internal_name] = val
 
+
+        for hook in self.post_set_hooks:
+            hook(obj, val)
 
     def __delete__(self,obj):
         raise TypeError("Cannot delete '%s': Parameters deletion not allowed."%self._attrib_name)


jlstevens:
I should also note that this PR relates to issue #170.

ceball:
> I should also note that this PR relates to issue #170.

Also: #160

jbednar:
Seems like a very reasonable extension, but note that it adds two slots to every parameter, and it seems unlikely that this mechanism will be used often, so that seems like a bit of a waste.  It would be awkward to do it in a single parameter (e.g. a tuple of the pre,post lists) but wouldn't be *that* bad; is it worth it?

jlstevens:
Tuples are fine with me. That said, what makes slots expensive to use? Memory consumption?

jbednar:
Isn't each one given a fixed amount of reserved memory space, per parameter?

ceball:
I quickly scanned the diff in hv. I may well be misreading that diff (and I don't know the classes there generally), but it appears to be effectively adding an 'instance-level' update method into the 'class-level' hook list (https://github.com/ioam/holoviews/pull/1740/files#diff-deb386f2061a9a3f41d4982d6a4f7926R203). 

So, every time you make a new instance, won't a new instance-level update method be added to the class-level hook list, and then when the parameter is set (on the class or on any instance), all those update methods will be called?

ceball:
Oh, actually I see it's making a new list each time - but in that case, when the parameter is set on the class or on any instance, it'll be the most recently created instance's update method that'll be called. (Assuming I read it right.)

ceball:
> Seems like a very reasonable extension,

Many things seem like they would be reasonable extensions to param. Poor param! ;)

I've often vaguely wanted something like this, but never actually needed it in the end. (Or maybe I gave up on doing something cool?) I.e. I want to add this feature, and I don't want to add it!

ceball:
> It would be awkward to do it in a single parameter (e.g. a tuple of the pre,post lists) but wouldn't be that bad; is it worth it?

> Isn't each [slot] given a fixed amount of reserved memory space, per parameter?

Ah, __slots__. It's been a long time since I was working on any of this, so don't necessarily believe anything I say, but...

I'd guess it would be 64 bits per slot per parameter object for most people (one python reference - i.e. one pointer - per slot). And maybe the same again for the list it points to? However, most uses of param don't result in many parameter objects - right? Parameter objects are created by Parameterized classes. Usually the number of instances of classes is much bigger than the number of classes. (And really we don't expect there to be all that many Parameterized instances, either.)

I think the benefit of slots in Parameter comes from e.g. not allowing mistakes in Parameter declarations (``defualt=4``)...because Parameter doesn't have Parameters ;) Oh, and also it usually prevents anyone summoning the willpower to add more stuff to Parameters! ;)

Slots do save memory (vs object with __dict__), and they do make attribute access faster (vs regular attribute lookup, or because they allow optimized access based on memory address - which could additionally avoid the GIL) but I doubt any of that helps us much. Slots definitely don't save my memory or time whenever I work on param...

jlstevens:
> So, every time you make a new instance, won't a new instance-level update method be added to the class-level hook list, and then when the parameter is set (on the class or on any instance), all those update methods will be called?

That is right for what I want in HoloViews, though really I only want this hook called when instance parameters are set. This is actually a bit of an annoyance - once you set the hook in the constructor once, it exists at the class level so if you make a new instance, the hook then exists and is called in the constructor of the second object you created, stopping you from setting parameters in the constructor.

I think I could easily update this PR to only apply hooks at the instance level, just by checking whether ``obj`` is a class or not before applying the hooks. Does this also seem more useful to you?

jbednar:
Adding slots for this sounds fine in terms of memory; I was thinking they would take space per instance, but if it's per class, I don't think it's worth worrying about.  So never mind on the tuple idea; that's just making things harder for ourselves.  I'm happy to add this feature as long as we either have found a very clear use case for it by the time of the next release, or we mark it experimental until we have done so.

philippjfr:
I'm +1 on this. It can also eventually be used to replace View parameters I added to paramnb and paramBokeh or at least unify them in some way.

jbednar:
> It can also eventually be used to replace View parameters I added to paramnb and paramBokeh or at least unify them in some way.

It would be great to eliminate those, so that objects can be specified using Param only.  View parameters work well, but they violate separation of content and presentation, so I'm very much in favor of this PR if indeed it does allow the specification of View parameters to be done at the Param level.

ceball:
Do the votes in favor mean positive reviews of how the implementation works now and you want to merge, or of the idea of pre- and post-set hooks in general?

Can someone point me to an example using View parameters?

philippjfr:
>Do the votes in favor mean positive reviews of how the implementation works now and you want to merge, or of the idea of pre- and post-set hooks in general?

My +1 was mostly about the idea, I'll look at the implementation now.

>Can someone point me to an example using View parameters?

There are some examples here: https://github.com/ioam/paramnb/blob/master/doc/AdditionalFeatures.ipynb

In essence View parameters declare a post-set hook, which are used to update the display output whenever you set the parameter.

param/__init__.py
```diff
@@ -609,8 +607,6 @@ def __set__(self,obj,val):
         Also applies set_hook, providing support for conversions
```
philippjfr: Should update this docstring since ``set_hook`` has now been removed.

philippjfr:
Okay I see the implementation is very simple. The main question I have is about instance level callbacks. I know you encountered the same issue when you prototyped this so I want to establish what the recommended approach will be. The problem is that you often want to define a callback per instance, e.g. when you create two parameterized instances to use as widgets or two stream instances you don't want setting the parameter on one instance to trigger the set hook installed by another instance. In paramnb/parambokeh I used the object id to trigger only the set hooks for a particular instance, how would you recommend this should be handled now?

ceball:
Thanks for the link about View parameters - will check those out.

> Okay I see the implementation is very simple.

Unfortunately I still don't understand it :(

> The main question I have is about instance level callbacks

Yes, I'm confused about what the desire/intent is here.

@jlstevens asked:
> I think I could easily update this PR to only apply hooks at the instance level, just by checking whether obj is a class or not before applying the hooks. Does this also seem more useful to you?

I don't think so, no. But I'm not sure about the intent. Can I just double check again that I understand what you are trying to achieve? You want to have something happen to all instances of a class when a parameter is set on any one of them?

And can I double check what your implementation does/what you want it do?

1. Right now in your implementation, the hook list you've created is at the parameterized class level. You are storing one single list of things to call when a parameter is set, whether that parameter is set via the class or it's set on any instance. If the parameter is set on the class, each thing in the parameter's hook list will be called with the class and the new value. If the parameter is set on instance x, each thing in the parameter's hook list will be called with x and the new value. If the parameter is set on instance y, each thing will be called with y and the new value.

2. Your proposed modification is to still have the hook list stored in the parameter (i.e. at the parameterized class level), but when the parameter is set on the class you will skip calling the hook list functions - you will only call them when the parameter is set on an instance (any instance). So again when the parameter is set on instance x, each function in the parameter's hook list will be called with x and the new value. If the parameter is set on instance y, each thing will be called with y and the new value.

If all of the above is right (is it?), I think I'd instead prefer the same kind of semantics as for setting and getting parameter values in general. That is, I'd want to be able to set pre-/post-set hooks per instance or per class, and to know which I'm doing. I would want to be able to say something like these things for class ``C`` and instances of it:

i. Whenever parameter ``a`` is set on instance ``x``, I want function ``f`` to be called (as ``f(x,new_val)`` plus *possibly* including other things like the parameter name but let's discuss that later). Setting ``a`` on instance ``y`` would do nothing. Setting ``a`` on class ``C`` would do nothing.

ii. Whenever parameter ``a`` is set (whether on the class ``C`` or on an instance), I want function ``f`` to be called (with the class/particular instance and new value as arguments). Setting ``a`` on instance ``x`` would result in ``f(x,new_val)``. Setting ``a`` on instance ``y`` would result in ``f(y,new_val)``. Setting ``a`` on class ``C`` would result in ``f(C,new_val)``.

I don't fully understand your hv example (sorry I still have not tried it out yet!), but if you wanted something to happen to all instances whenever the value of a parameter is set on one instance, you could achieve that by adding one function per instance (e.g. an instance method from each instance) to the class-level hook list.


> In paramnb/parambokeh I used the object id to trigger only the set hooks for a particular instance, 

@philippjfr could you point me to that? Sorry to be lazy and keep asking for pointers but it's tricky for me to keep track of stuff right now.

> how would you recommend this should be handled now?

I would expect the class-level hook list to be overridden and stored at the instance level instead.

Unfortunately I still haven't got round to gathering/trying out/understanding the various supplied examples yet to see what it is that actually needs to be supported; I don't want to hold up this PR and be an annoying bottleneck, but I also want to try to protect param a bit.

philippjfr:
>Unfortunately I still haven't got round to gathering/trying out/understanding the various supplied examples yet to see what it is that actually needs to be supported; I don't want to hold up this PR and be an annoying bottleneck, but I also want to try to protect param a bit.

No I think this an important discussion since I expect this functionality to be widely used by both holoviews streams and paramnb/parambokeh and we really want to get it right at the param level. Thanks for outlining the various issues so neatly. I agree with everything you've said and setting the hooks at the class/instance level makes sense. Here is what I did in paramnb: [view.py#L22](https://github.com/ioam/paramnb/blob/master/paramnb/view.py#L22)

In that version I use a dictionary of ``callbacks`` indexed by the instance/class object id, which I store on the parameter itself. I then manually register a callback for each instance. Storing and setting it on the parameterized class/instance as you suggest is probably preferable though if we add this functionality to param itself.

jbednar:
What Chris outlined makes very good sense to me.  Chris, would you be able to implement that, or would you want Philipp (as someone who has already done the same thing one level up, and has a specific use case for it) or Jean-Luc (as the author of this PR, and also has a specific use case) to do it?

ceball:
Yes, I'd be happy to continue the implementation once I've understood Jean-Luc's and Philipp's examples (they are likely to have thought of doing things that wouldn't occur to me, and so far I've only looked at the examples superficially). So I'll study the examples mentioned in the comments in more detail (along with another widget example that came up recently in another project) and then implement the above if it seems to fit.

jlstevens:
> You want to have something happen to all instances of a class when a parameter is set on any one of them?

That was my original goal but as Philipp points out, hooks specific to each instance would also be useful.

> Right now in your implementation, the hook list you've created is at the parameterized class level. You are storing one single list of things to call when a parameter is set, whether that parameter is set via the class or it's set on any instance. If the parameter is set on the class, each thing in the parameter's hook list will be called with the class and the new value. If the parameter is set on instance x, each thing in the parameter's hook list will be called with x and the new value. If the parameter is set on instance y, each thing will be called with y and the new value.

That sounds about right. As mentioned earlier, the problem with setting hooks at the class level is you trigger the hooks when trying to set parameters via constructors. There must be a better way than the simple thing I attempted to do here.

> Your proposed modification is to still have the hook list stored in the parameter (i.e. at the parameterized class level), but when the parameter is set on the class you will skip calling the hook list functions - you will only call them when the parameter is set on an instance (any instance). So again when the parameter is set on instance x, each function in the parameter's hook list will be called with x and the new value. If the parameter is set on instance y, each thing will be called with y and the new value.

Correct.

Your parts i and ii outline the per-instance approach (which I didn't attempt).

> I don't fully understand your hv example (sorry I still have not tried it out yet!), but if you wanted something to happen to all instances whenever the value of a parameter is set on one instance, you could achieve that by adding one function per instance (e.g. an instance method from each instance) to the class-level hook list.

That is fine, all instance can point to the same function but with the updated suggestion each instance *could* point to a *different* function.

So in summary, I am in favor of per-instance hooks which isn't currently in this PR. My goal was to do something simple to start a discussion and I think it has now done that job! :-)

Issue #179, opened 2017-07-22
all_equal() is broken


ceball:
Unless I'm missing something?

```
>>> from param.parameterized import all_equal
>>> all_equal([1],[])
True
>>> all_equal([1],[0])
False
>>> all_equal([1,2],[1])
True
>>> all_equal([1],[1,2])
True
```

```
def all_equal(arg1,arg2):
    """
    Return a single boolean for arg1==arg2, even for numpy arrays
    using element-wise comparison.
    Uses all(arg1==arg2) for sequences, and arg1==arg2 otherwise.
    If both objects have an '_infinitely_iterable' attribute, they are
    not be zipped together and are compared directly instead.
    """
```

Why ``return all(a1 == a2 for a1, a2 in zip(arg1, arg2))``?

```
commit d767150659da4e526d53b10c011deb0a15ed3f97
Author: Dobromir Stefanov <d.stefanov@sms.ed.ac.uk>
Date:   Tue Mar 27 23:07:54 2012 +0000

    API cleanup for param and topo.base
```

```
-        return all(arg1==arg2)
+        return all(a1 == a2 for a1, a2 in zip(arg1, arg2))
```

https://github.com/ioam/svn-history/commit/d767150659da4e526d53b10c011deb0a15ed3f97

jbednar:
Is your objection that it should be verifying that the lengths are the same, because zip does not do that?

ceball:
Yes, mismatched lengths is one objection. I'd probably expect the shapes to be the same. Another (from #180) is that it doesn't work for >1d arrays:

```
>>> a = np.array([[1,2],[3,4]])
>>> all_equal(a,a)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/cball/code/ioam/param2/param/parameterized.py", line 144, in all_equal
    return all(a1 == a2 for a1, a2 in zip(arg1, arg2))
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

Is there some reason it was intended to only check elements up to the shortest one?

jbednar:
I'm sure that would have been an oversight, not a design feature.

ceball:
One feature provided by this all_equal function is to suppress Parameterized names almost all the time when doing script_repr.

```
>>> param.parameterized.all_equal('Parameterized00001','Parameterized')
True
```

means

```
>>> param.parameterized.all_equal(param.Parameterized().name,param.Parameterized.name)
True 
```

[...]

PR #180, opened 2017-07-22
Demonstrate problem with pprint() and numpy arrays.

ceball:
Not for merging yet, just for demonstrating problem (should show up on travis). Will fix in a future commit, and probably improve this test too.

The problem comes from parameterized.all_equal(): see #179.

### DIFF
--- a/tests/testparameterizedrepr.py
+++ b/tests/testparameterizedrepr.py
@@ -3,6 +3,8 @@ Unit test for the repr and pprint of parameterized objects.
 """
 
 import unittest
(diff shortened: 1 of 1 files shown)


tonyfast:
closing this pr as stale. the issue still persists.

ceball:
Would you accept this PR with the test marked "known fail"? (if that's possible - can't remember how param's test suite is run)

Or the test here is actually stale in some way?

jbednar:
The PR claims that it's demonstrating a problem, not that it's fixing one, so what is there to be accepted?

Issue #167, opened 2017-07-04
Clean up release procedure


ceball:
https://github.com/ioam/param/wiki/Release-instructions is out of date, and contains code fragments that could potentially be added to the repository (e.g. conda package info). Also, current param conda packages contain various extraneous files and data.

ceball:
Param's packaging should be straightforward because it's a pure python package with no dependencies. However, need to consider conda and constructor. A selection of the platforms param is available for on various channels (anaconda/defaults/main, conda-forge, ioam):

<img width="173" alt="screen shot 2017-08-20 at 4 06 15 pm" src="https://user-images.githubusercontent.com/1929/30252034-487ab43c-9663-11e7-9289-d84041e79724.png">

<img width="183" alt="screen shot 2017-08-20 at 4 05 59 pm" src="https://user-images.githubusercontent.com/1929/30252022-0f80e49e-9663-11e7-98af-f0a2ab7d60d4.png">

<img width="176" alt="screen shot 2017-08-20 at 4 08 48 pm" src="https://user-images.githubusercontent.com/1929/30252025-1506a656-9663-11e7-9c7c-2f341f04c13c.png">

(not in order)

ceball:
See also: #189, #192

Issue #166, opened 2017-07-04
Decide about cythonizing


ceball:
Currently, pip install/python setup.py attempts to cythonize. This may bring performance improvements (e.g. when creating many parameterized objects), although we have no performance tests running to back that up (https://github.com/ioam/param/issues/165). However, there are a number of potential downsides of cythonizing:

(1) source code is obscured/people might think it's not purely python e.g. if they try to open the file they've imported;

(2) possible errors/incompatibilities introduced by cython we're not aware of (although you could also argue cythonizing helps catch a class of problems, and also cythonizing by default means we're all testing cython's effects all the time);

(3) possible installation failures (e.g. https://github.com/ioam/param/issues/129) or confusion (e.g. https://github.com/ioam/param/issues/161);

(4) just more stuff to think about overall.

Incidentally, conda packages have never included cythonized param. People would only have the cythonized version by pip install/python setup.py from pypi/github on a machine with cython and a working compiler.

ceball:
pip-installed (i.e. cythonized) param causes the following error somewhere in topographica's tests:

```
$ TOPO_SKIP_RUNTESTS_GUI_IF_NO_DISPLAY=1 DISPLAY= ./topographica -t all
[...]
Traceback (most recent call last):
  File "/Users/cball/code/ioam/topographica/topo/command/__init__.py", line 867, in __call__
    save_script_repr()
  File "/Users/cball/code/ioam/topographica/topo/command/__init__.py", line 396, in save_script_repr
    script = header+topo.sim.script_repr()
  File "/Users/cball/code/ioam/topographica/topo/base/simulation.py", line 1537, in script_repr
    sorted(self.objects().values(), cmp=lambda x, y: cmp(x.name,y.name))]
  File "/Users/cball/code/ioam/topographica/topo/base/simulation.py", line 240, in script_repr
    super(EventProcessor,self).script_repr(imports=imports,prefix=prefix)
  File "param/parameterized.py", line 1236, in param.parameterized.Parameterized.script_repr
    return self.pprint(imports,prefix, unknown_value=None, qualify=True,
  File "param/parameterized.py", line 1285, in param.parameterized.Parameterized.pprint
    value = pprint(values[k], imports, prefix=prefix,settings=[],
  File "param/parameterized.py", line 891, in param.parameterized.pprint
    rep=val.script_repr(imports, prefix+"    ")
  File "param/parameterized.py", line 1236, in param.parameterized.Parameterized.script_repr
[...]

PR #194, opened 2017-09-10
Remove cython

### LINKED ISSUE #161: ``python setup.py develop`` is silently broken for param
Now that the default of param's setup.py is to cythonize, ``python setup.py develop`` also cythonizes and you end up importing from the generated .so files. If you then modify the python source of param, your changes don't show up unless you redo the ``python setup.py develop`` step (and wait for re-compilation).

Maybe we could add something to param's setup.py so that the develop command does not cythonize unless explicitly requested?

Until then, you either have to use PYTHONPATH or delete the .so files.

ceball:
My opinion about #166. Note that the only people to have been getting cythonized param would be those installing from pip while also having cython and a functioning c compiler.

* Closes #161 (makes it redundant).
* Hopefully allows noarch conda packages, making #167 simpler.

If merged, I would open an issue about considering cythonizing once: 
* there's higher test coverage of param (e.g. resurrecting topographica's tests picked up a problem with cythonized param)
* there are performance tests

Meanwhile, it might be useful to cythonize param as part of testing (e.g. can be useful to catch certain classes of error/bad practice). We could maybe add a command/argument to setup.py to optionally cythonize param.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,25 +7,14 @@ try:
 except ImportError:
     from distutils.core import setup
 
-try:
-    # check we can compile on this machine
-    import cython; cython.inline("return 1;")
-    
-    from Cython.Build import cythonize
-    ext_modules = cythonize("param/*.py", exclude=['param/ipython.py'])
-except:
-    ext_modules = []
 
-setup_args = {}
-
-setup_args.update(dict(
+setup_args = dict(
     name='param',
     version="1.5.1",
     description='Declarative Python programming using Parameters.',
     long_description=open('README.rst').read() if os.path.isfile('README.rst') else 'Consult README.rst',
     author= "IOAM",
     author_email= "developers@topographica.org",
-    ext_modules=ext_modules,
     maintainer="IOAM",
     maintainer_email="developers@topographica.org",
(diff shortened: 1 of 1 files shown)


ceball:
> Note that the only people to have been getting cythonized param would be those installing from pip while also having cython and a functioning c compiler.

Or people installing from http://www.lfd.uci.edu/~gohlke/pythonlibs/#param

philippjfr:
Looks fine to me. We should go back and look at what benefits cythonizing actually provides. I think it may be worth it in some cases but the current situation where certain installs get the cythonized version and others do is clearly not right.

>resurrecting topographica's tests picked up a problem with cythonized param

Could you remind me what that issue was?

jlstevens:
I think I approve of this change as well. Maybe @jbednar would like to merge?

jbednar:
I guess I'll go ahead and merge. I generally think that if there are performance problems caused by param then people are not using param the way it is intended, and this should simplify things overall.

ceball:
@philippjfr:

> Could you remind me what that issue was?

https://github.com/ioam/param/issues/166#issuecomment-325225105

PR #200, opened 2017-10-04
Added argument to param.Version to support dev versions

jlstevens:
Added ``dev`` argument that allows dev versions to be handled like regular release versions. I'm happy to add more unit tests to make sure everyone is satisfied with this addition.

### DIFF
--- a/param/version.py
+++ b/param/version.py
@@ -107,10 +107,12 @@ class Version(object):
     compute the number of commits since the last version tag, the
     short commit hash, and whether the commit is dirty (has changes
     not yet committed). Version tags must start with a lowercase 'v'
-    and have a period in them, e.g. v2.0, v0.9.8, v0.1a, or v0.2beta.
-    Note that any non-numeric portion of the version ("a", "beta",
-    etc.)  will currently be discarded for the purposes of numeric
-    comparisons.
+    and have a period in them, e.g. v2.0, v0.9.8 or v0.1.
+
+    Development versions are supported by setting the dev argument to an
+    appropriate number. The corresponding tag then takes the form
+    v0.1dev3, v1.9.0dev2 etc and the version string will be formatted in
+    the corresponding manner.
 
     Also note that when version control system (VCS) information is
     used, the comparison operators take into account the number of
@@ -126,12 +128,14 @@ class Version(object):
       __init__.py export-subst
     """
 
-    def __init__(self, release=None, fpath=None, commit=None, reponame=None, commit_count=0):
+    def __init__(self, release=None, fpath=None, commit=None,
+                 reponame=None, dev=None, commit_count=0):
         """
         :release:      Release tuple (corresponding to the current VCS tag)
         :commit        Short SHA. Set to '$Format:%h$' for git archive support.
         :fpath:        Set to ``__file__`` to access version control information
         :reponame:     Used to verify VCS repository name.
+        :dev:          Development version number. None if not a development version.
         :commit_count  Commits since last release. Set for dev releases.
         """
         self.fpath = fpath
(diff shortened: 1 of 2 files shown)


jlstevens:
I'll definitely add enough unit tests to make sure test coverage increases instead of decreasing.

param/version.py
```diff
@@ -233,6 +243,7 @@ def __str__(self):
         """
         if self.release is None: return 'None'
         release = '.'.join(str(el) for el in self.release)
+        release = '%sdev%d' % (release, self.dev) if self.dev else release
```
jbednar: Shouldn't there be a "." before "dev" by PEP440?

jbednar:
Looks fine by me apart from comment above.

param/version.py
jlstevens: I think it is fine, this is a snippet Bryan posted:

```python
In [1]: from packaging.version import Version as V

In [2]: V("0.12.10dev3")
Out[2]: <Version('0.12.10.dev3')>

In [3]: v = V("0.12.10dev3")

In [4]: v.is_prerelease
Out[4]: True

In [5]: v.base_version
Out[5]: '0.12.10'

In [6]: v > V("0.12.11")
Out[6]: False
```

That said, I am happy to consider adding the dot - even though our existing dev tags don't have the dot, they have all been lightweight tags so far (and therefore have no effect on ``param.Version``).

@philippjfr Do you want the dot?

jlstevens: If we aren't worried about people referencing the old tags (I'm not - they are dev versions!) then I could go back and replace the old dev tags with the new format.

jlstevens: Alternatively, we can delete the tags for old dev versions that have been followed by a release. Then we can use the new format going forwards.

philippjfr: I'd prefer to leave out the dot personally, bokeh does too.

ceball:
I am also ok with this.

Thanks a lot for doing it. Despite the jokes, I appreciate it. (I will try to come up with new versions of the jokes.)

Seems unfortunate that it's necessary to support both with and without the dot
  * to match PEP 440 (dot MUST be there) and its implementation in packaging ("normalized" form has dot, but the dot is optional)
  * to match what's out there (see below)

-----

Examples off the top of my head...

No dot:
* holoviews
* bokeh
* datashader

Dot:
* packaging
* pandas
* numba

screen shot 2017-10-04 at 10 36 50 pm

jlstevens:
@philippjfr @ceball @jbednar 

I added a pep440 flag we can use to stick to the ``devX`` format instead of the PEP440 ``.devX`` format. If you agree with this approach, I'll add more unit tests and update the docstrings to explain this option.

param/version.py
```diff
@@ -233,6 +249,8 @@ def __str__(self):
         """
         if self.release is None: return 'None'
         release = '.'.join(str(el) for el in self.release)
+        release = ('%s%sdev%d' % (release, '.' if self.pep440 else '', self.dev)
+                   if self.dev else release)
```
ceball: Do people do `dev0` (thinking along the lines of 0 being false above)? Actually, I realize now I have seen e.g. `1.x.dev` (numba, pandas). But we can't support everything, and I think just `.dev` doesn't comply with PEP 440 ("The developmental release segment consists of the string .dev, followed by a non-negative integer value").

This is all getting a bit crazy, sorry...

jlstevens:
I think I'll go with a suggestion from @jbednar - support tags with and without the dot and return the version string with the dot. Then you have the flexibility for tagging while using the PEP440 compliant version strings.

ceball:
Looks ok to me, thanks :)

PR #202, opened 2017-10-26
Streamline releasing and testing

ceball:
This is a PR about streamlining param releases (fixes #167), and 
improving testing.

1. Added conda recipe and package upload

    Builds and uploads conda packages for (non-PR/simulated) commits to master.  I.e. latest master is always available as a conda package from anaconda.org/cball/param.

2. Added pypi package building and uploading

    For tags, builds sdist and universal wheel, and uploads to test.pypi (simulating actual releases to pypi).

3. Enabled testing on Windows via Appveyor.

    Param is pure python, but it's useful to test on windows to catch accidental platform specific content (whether that's code, developer workflow, testing, or whatever). Fixes #195.

4. Use tox for testing

    There is now very little scripting in the travis and appveyor config files (except for conda, which has no integration with travis). The same tests can be run locally as on CI. Nothing is platform specific.

    Tox not only tests 'python setup.py develop', but also creates an sdist, installs it, and tests it. Fixes #192.

    Changes include:

      * Moved dependencies out of platform-specific travis file and into setup.py (if genuine dependency) or tox (if test dependency).

      * Now tests that optional dependencies are actually optional. Previously, many other packages were present in the test environment. Have also started to test that specific functionality works when the optional dependencies are present. Fixes #189.

      * Fixed running of flake8.

      * Warnings during tests are now displayed (fixes #86).

    Being python only, param can take advantage of the standard python ecosystem, so unfortunately not all of this approach is suitable for conda-dependent packages.

Also, minor updates to README, including coverage badge (not sure why hiding it makes sense; closes #65).

PR #203, opened 2017-10-28
Drop support for EOL Python 2.6

### LINKED ISSUE #175: Drop support for python 2.6?
I struggle to remember what's valid in all the different pythons, and thinking about it takes up time. It used to be worthwhile to support multiple versions of Python 2 while people moved between versions and when people relied on their system's python, but those days seem to be gone with everyone converging on Python 2.7 (Python 2.6 hasn't had an update since 2013; https://www.python.org/download/releases/2.6.9/). And it's not like we've kept supporting every version of Python 2 we ever used (we happily dropped 2.5, 2.4, ... etc in the past).

But please comment if you need param to keep working on Python 2.6...

hugovk:
Fixes #175.


ceball:
Thanks for following up comments with code :) The delay in responding is down to PR #202, opened just two days before this one, which ought to be merged first. Once that's finished, we'll move onto this...

hugovk:
PR #202 is now merged, I've rebased this on master.

ceball:
Thanks for updating your PR. I'd be happy for this to be merged, and I think it's ok for the next release of param not to support 2.6. 

(@jbednar, ok to merge, or do you see any problems?)

jbednar:
Looks good; happy to see the end of 2.6!

PR #207, opened 2018-01-11
Upload "latest master" conda package to pyviz/label/dev

ceball:
When merged, will make a conda package appear at pyviz/label/dev for any commit to master. The conda package will have the version style shown below:

screen shot 2018-01-11 at 5 31 06 pm

Note: 
  * before merging, should delete the test packages
  * needs to be squash merged
  * the contained python package `__version__` will be what's hardcoded in the param source code i.e. the most recent release, I think (something that could be addressed in the future)

Another note: I don't have a strong opinion about the version format we use. Also, various similar forms seems to be accepted and 'canonicalized':
```
>>> from packaging.version import Version as V
>>> V('1.2.3.post1.dev0')
<Version('1.2.3.post1.dev0')>
>>> V('1.2.3.post1.dev')
<Version('1.2.3.post1.dev0')>
>>> V('v1.2.3.post1.dev')
<Version('1.2.3.post1.dev0')>
>>> V('v1.2.3.post1dev')
<Version('1.2.3.post1.dev0')>
```

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: param
-  version: {{environ['GIT_DESCRIBE_TAG'].lstrip('v')}}.post{{GIT_DESCRIBE_NUMBER}}.dev+{{GIT_DESCRIBE_HASH}}
+  version: {{ os.environ.get("VERSIONHACK") }}
 
 source:
   path: ..
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: param
-  version: {{environ['GIT_DESCRIBE_TAG'].lstrip('v').split('.dev')[0]}}.dev{{GIT_BUILD_STR }}
+  version: {{environ['GIT_DESCRIBE_TAG'].lstrip('v')}}.post{{GIT_DESCRIBE_NUMBER}}.dev+{{GIT_DESCRIBE_HASH}}
 
 source:
   path: ..
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -143,7 +143,7 @@ Official releases of Param are available on
 ``conda install -c ioam param``, ``pip install --user param``, or 
 ``pip install param``.
 
-The very latest changes can be obtained via ``conda install -c cball
+The very latest changes can be obtained via ``conda install -c pyviz
 param`` or ``pip install
 https://github.com/ioam/param/archive/master.zip``.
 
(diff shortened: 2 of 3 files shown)


jbednar:
I don't think ".dev" is legal PEP440, without the numeric digit, so I don't think we should be using that form.  All things considered, I think I'm now favoring "next predicted release" as the base, with ".devN" and no ".post".

doc/index.rst
```diff
@@ -143,7 +143,7 @@ Official releases of Param are available on
 ``conda install -c ioam param``, ``pip install --user param``, or 
 ``pip install param``.
 
-The very latest changes can be obtained via ``conda install -c cball
+The very latest changes can be obtained via ``conda install -c pyviz
```
jbednar: Shouldn't it say /label/dev if it's the "very latest"?

conda.recipe/meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: param
-  version: {{environ['GIT_DESCRIBE_TAG'].lstrip('v').split('.dev')[0]}}.dev{{GIT_BUILD_STR }}
+  version: {{environ['GIT_DESCRIBE_TAG'].lstrip('v')}}.post{{GIT_DESCRIBE_NUMBER}}.dev+{{GIT_DESCRIBE_HASH}}
```
jbednar: Hmm.  If it's .yml, I don't know how to do the arithmetic necessary to use the "predicted next release form"; how will it increment the result of GIT_DESCRIBE_TAG?

doc/index.rst
ceball: Yes, well spotted, that's a mistake.

conda.recipe/meta.yaml
ceball: If setup.py had the correct version, I would instead use that. Something like:

```
{% set setup_data = load_setup_py_data() %}

package:
  name: nbsite
  version: {{ setup_data.get('version') }}
```

The version in setup.py can be computed by python. (So in projects where I've used a versioning tool to ensure the version's always up to date from git, there's no need to do the "manual version construction" I've done in this PR.)

ceball:
> I don't think ".dev" is legal PEP440, without the numeric digit, so I don't think we should be using that form. 

I think it's equivalent to `.dev0` (I edited my original comment probably after you read it), although yes I agree from the regex in PEP440 it isn't. Anyway, I'm definitely not tied to how we do it. Plus...

> All things considered, I think I'm now favoring "next predicted release" as the base, with ".devN" and no ".post".

Heh, I think that's what I originally suggested in the pyver issue, so at one point I must have thought it was better too. But I can't remember any more.

> not sure if the "predicted next release" form is feasible?

I could compute the version number using python and pass it to conda build via an environment variable.

jbednar:
Ok, I vote for the predicted next version then.

PR #208, opened 2018-01-22
Fixed checks for None

philippjfr:
Checking for ``if not obj`` is wrong for objects that define a custom ``__nonzero__`` or ``__len__`` method and this appears to be the cause of multiple bugs in HoloViews. It's particularly bad because this happens during Parameterized instantiation, in which case ``__nonzero__`` and ``__len__`` may not yet work because the required attribute may not be set up yet.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -449,7 +449,7 @@ class Dynamic(Parameter):
 
         dynamic = callable(val)
         if dynamic: self._initialize_generator(val,obj)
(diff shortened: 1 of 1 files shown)


ceball:
Thanks. We should merge this.

I think there are likely to be other similar cases (i.e. of assuming things about objects, rather than just explicitly checking for None). We should go through all similar cases and check if the intention is actually to check for None.

Issue #210, opened 2018-01-24
Search code for cases where intention is to check for None, and make explicit


ceball:
See #208.

tonyfast:
Maybe some of these cases will fall out or be revealed in the 2.0 release.

maximlt:
This issue has been opened for 5 years now, I'm going to close it. We can re-open it if we find the cases it indicated.

PR #211, opened 2018-01-24
Enable more flake8 checks.

ceball:
Presumably won't catch cases like in #208, but might catch some others...we'll see.


ceball:
Doesn't address #210 (I don't see how it could), but still seems useful and not too aggressive/annoying.

For reference...
 * Flake errors (all enabled - unchanged): http://flake8.pycqa.org/en/latest/user/error-codes.html
 * pep8 style conventions (enabled some): https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes

PR #214, opened 2018-03-05
Switch to autover

ceball:
Changes:
* `param.__version__` now complies with pep440, and is a string
* `param.version.Version` uses a different versioning scheme from before; see autover for more details.


param/__init__.py
```diff
@@ -36,10 +36,13 @@
 # only two required files.
 try:
     from .version import Version
-    __version__ = Version(release=(1,5,1), fpath=__file__,
-                          commit="$Format:%h$", reponame="param")
+    _version = Version(release=None, fpath=__file__,
```
jbednar: Will autover accept (1,5,1) and do something reasonable with it, so that people who have been using the old param.version.Version won't have problems?

jlstevens: I *think* so as ``autover.Version`` is based on the old ``param.Version``. The issues are:

1) Supporting the old behavior properly makes things more complicated.
2) This old approach is not properly tested.

I'm pretty sure it will be accepted but I can't guarantee it will do the right thing! I think the original unit tests were moved to autover but I would need to check how many remain.

jlstevens: I should add that we are the only people who used ``param.Version`` as far as I know (unlikely autover it wasn't a separate project and it wasn't advertised).

jlstevens: One option would be to explicitly *disallow* these tuples making sure that projects depending on the next param release have to update to the new approach.

jbednar: We haven't advertised it recently, but it has always been available for use, the code at the top of the old Version file explained exactly how to use it in external projects, and though it's been years now, I definitely recommended it in various talks to audiences I have now forgotten.

jlstevens: I'm not sure what to do about this. On one hand I *think* this is mostly backwards compatible (other than the deprecated ``commit_count`` argument that wasn't often used) but I am not convinced that it is worth the effort to promise backwards compatibility add new tests etc. Even if it an explicit release tuple is fine, the actual version string format has changed and is not PEP440 compliant.

I think the honest thing to do here would be to issue a warning if ``release`` is not ``None`` or if deprecated arguments like ``commit_count`` are used. That way if any projects are using ``param.Version`` that we aren't in control of *also* upgrade their param version, they get a warning so they know that they should at least check that ``Version`` still works appropriately for their projects (and to prompt them to know that there is a new recommended workflow that is easier, purely based on tags).

jbednar: That sounds fine to me.

jbednar: Shouldn't release be None by default, then?

jlstevens: It is: https://github.com/ioam/param/pull/214/files#diff-0187a55516aa9995e9033ad5f0d98b33R170

jlstevens: The way it is specified in the line above is explicit but not strictly necessary. Explicit is good though!

setup.py
```diff
@@ -14,17 +14,43 @@
...
     """
     Helper to get the current version from either git describe or the
     .version file (if available).
     """
     basepath = os.path.split(__file__)[0]
     version_file_path = os.path.join(basepath, reponame, '.version')
-    try:
-        version = importlib.import_module(reponame + ".version") # Bundled
-    except:
-        version = None
```
jlstevens: I'll just note that the shortened version looks fine! We can return to it later but for now, it is safest to stick to what is properly tested in autover.

jlstevens:
One suggestion might be to shorten the functions in ``setup.py`` after checking autover is working nicely by doing a quick param dev release. :-)

Then if it all works well, we can shorten things again before the actual param release.

ceball:
I just unshortened the setup.py functions after your recommendation. I can't tell what all the code paths do, but I think at this stage I would just like to use it and move on. Can we merge this?

jlstevens:
I'm happy to see this merged but I would recommend that you try doing a param dev release before doing the actual release. I'm guessing @jbednar should be the one to merge or not?

jlstevens:
I can do the quick change suggested by Jim on both the autover and param repos after this is merged.

param/version.py
```diff
@@ -177,6 +177,10 @@ def __init__(self, release=None, fpath=None, commit=None, reponame=None,
         """
         self.fpath = fpath
         self._expected_commit = commit
+
+        if release is not None or 'commit_count' in kwargs:
+            print('WARNING: param.Version now supports PEP440 and a new tag based workflow. See param/version.py for more details')
+
```
jlstevens: I would like to use ``param.warning`` but that will probably cause cyclic import issues unless I inline the import? Is a print statement ok in this case and do you have any suggestions for a better message?

param/__init__.py
jbednar: But since this is code that gets copied into every new project, I'd rather have it as short as possible here.

param/version.py
jbednar: Using print is ok by me.

jbednar:
I'm happy for it to be merged as soon as release=None is deleted, to keep the boilerplate as simple as possible.

param/__init__.py
jlstevens: Fixed in 0954adf. The version declaration shouldn't be copied verbatim between projects but I agree it is worth making this short as it is the reference example!

param/version.py
jlstevens: This is so simple I didn't test it locally. I would wait for travis here and also on autover where I'll port this change across now. The only thing I'm wondering about is whether this print will be triggered in some of the unit tests.

jlstevens:
I've made the requested changes both here and to the autover repo. Params tests are green (above) and so are [autover's](https://travis-ci.org/ioam/autover/builds/349482622).

PR #216, opened 2018-03-19
Remove trailing whitespace

jbampton:


ceball:
Thanks! This change will be appreciated by members of ioam :)

Unfortunately, I'm a frequent trailing whitespace offender. I did set up my editor for python/pep8, and wanted to just pep8 all the code, but then it turned out pep8 is rejected by the majority of ioam in favor of custom/varying rules that I've never quite learned, so I went back to being an offender...

jbampton:
Hi @ceball I did some linting for ioam once before and learned a bit about their views on pep 8.  And tonight on another issue here -> https://github.com/ioam/holoviews/issues/2457  @jlstevens said it was worth while to remove the trailing whitespace so I went ahead and did it myself here :)

PR #217, opened 2018-04-03
setup.py cleanup

ceball:
Declare that only python 2.7+ is supported (true since #203)

Be explicit about using setuptools (pip always uses setuptools, even if you don't import setuptools).

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -14,6 +14,7 @@ build:
 requirements:
   build:
     - python
+    - setuptools
   run:
     - python
 
--- a/setup.py
+++ b/setup.py
@@ -3,16 +3,10 @@ import sys
 import json
 import importlib
 
-try:
-    from setuptools import setup
-except ImportError:
-    from distutils.core import setup
+from setuptools import setup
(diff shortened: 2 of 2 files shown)


ceball:
I guess one issue with this PR is that pip didn't become part of python.org pythons until 3.4 and 2.7.9. So people using python.org python 2.7.0-2.7.8 or python 3.0-3.3 would now need to install pip or setuptools to be able to install param (they'd install pip via their package manager or https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py).

CI systems that provide python 3.3 include pip+setuptools, which is I guess why our tests keep working.

Anaconda currently supports python 3.4+. Old versions of anaconda that did include python 3.3 (e.g. 2.3) also include pip+setuptools.

ceball:
I guess python 3.3 users will be used to having to get pip and setuptools for various packages by now, though.

And, "[As of 2017-09-29, Python 3.3.x reached end-of-life status](https://www.python.org/dev/peps/pep-0398/#lifespan)".

jlstevens:
Looks good to me!

PR #218, opened 2018-04-03
Drop python 3.2 support.

ceball:


jbednar:
Python 3.2 is still working as far as we know, but it was causing some problems with the packages we use for testing, so we'll turn it off and no longer consider it supported now.

ceball:
For the record, we could probably still test 3.2 if we didn't set up the test environment using pip/virtualenv, maybe. The actual error while creating the test environment was:

screen shot 2018-04-03 at 10 04 46 pm

https://travis-ci.org/ioam/param/jobs/361810275

PR #220, opened 2018-04-03
Simplifications related to packaging and testing.

ceball:
* specify test dependencies in setup.py
  * simplify running of test tools
  * reduce duplication of packaging metadata


ceball:
Some bits (package building & uploading) will only be tested after merging to master.

Issue #215, opened 2018-03-06
Simplify version-related code in setup.py and __init__.py?


ceball:
Not sure I've got this right; do any of the following changes make sense?

## setup.py embed_version() 

Will never be called (because version.py is bundled): remove this function?

## setup.py get_setup_version()
  * two tries to import `param.version`
  * `version` will never be None (as version.py is bundled)
  * `version_file_path` will never be used

So could it be simplified to something like this?

```python
def get_setup_version(reponame):
    import param.version
    return param.version.Version.setup_version(os.path.split(__file__)[0], reponame, dirty='raise', archive_commit="$Format:%h$")
```

(Unrelated note: changed `dirty` from `strip` to `raise`. I think `report` would be ok too. I'm not sure `strip` should be an option, let alone the default - what am I missing?)

## `__init__.py`

Currently:

```python
# Determine up-to-date version information, if possible, but with a
# safe fallback to ensure that this file and parameterized.py are the
# only two required files.
try:
    from .version import Version
    _version = Version(fpath=__file__, archive_commit="$Format:%h$", reponame="param")
    __version__ = str(_version)
except:
    import json
    __version__ = json.load(open(os.path.join(os.path.split(__file__)[0],
                                              '.version'), 'r'))['version_string']
```

* For the comment to still be true, would need an "if exists .version" or one more except.
* Should probably delete names once done with them (`Version`, `json`)?
* Will the current except ever be used? Doesn't Version itself look for the .version file? I might have misunderstood...

Could it be changed to something like the following?

```python
# Determine up-to-date version information, if possible, but with a
# safe fallback to ensure that this file and parameterized.py are the
# only two required files.
try:
    from .version import Version
    _version = Version(fpath=__file__, archive_commit="$Format:%h$", reponame="param")
    __version__ = str(_version)
except:
    __version__ = '0.0.0+unknown'
```

(where 0.0.0+unknown could be whatever makes most sense for indicating version unknown)

jbednar:
That all sounds good to me.  Do we *need* to keep _version around? It would be nice to make this as simple as possible...

jlstevens:
This should be fine for ``__init__.py``:

```python
from .version import Version
__version__ = str(Version(fpath=__file__, archive_commit="$Format:%h$", reponame="param"))
```

For ``setup.py`` I agree we shouldn't need ``embed_version()`` but otherwise I cannot tell offhand if your suggested change is fine. I suggest trying it out and testing it.

PR #222, opened 2018-04-04
Update version.py from autover, and simplify usage (see #215).

ceball:


param/__init__.py
```diff
@@ -36,12 +36,9 @@
...
 try:
     from .version import Version
-    _version = Version(fpath=__file__, archive_commit="$Format:%h$", reponame="param")
-    __version__ = str(_version)
+    __version__ = str(Version(fpath=__file__, archive_commit="$Format:%h$", reponame="param"))
 except:
-    import json
-    __version__ = json.load(open(os.path.join(os.path.split(__file__)[0],
-                                              '.version'), 'r'))['version_string']
+    __version__ = "0.0.0+unknown"
```
ceball: Can't remember what we decided would be best here. `"None"`? Something else?

ceball:
After review, presumably the way to test this is to merge to master, and then:
  * Manually trigger an autover build on travis (one of autover's tests fetches latest master param and uses that).
  * Try getting autover from param in other projects? (After merging, a conda package will appear automatically at pyviz/label/dev; a new pip package will appear at https://testpypi.python.org/pypi/param/ whenever an alpha tag is pushed. Or could use param master from github.)

param/__init__.py
philippjfr: ``"0.0.0+unknown"`` seems good. It's PEP440 compliant, and if for whatever reason someone installs it any real version number will take precedence.

jlstevens: Agreed. I think this was what we decided on as being better than ``"None"``.

setup.py
```diff
@@ -82,7 +43,6 @@ def get_setup_version(reponame):
     url='http://ioam.github.com/param/',
     packages=["param","numbergen"],
     provides=["param","numbergen"],
-    package_data = {'pkg_bundle': ['.version']},
```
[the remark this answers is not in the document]
ceball: It's already in MANIFEST.in; combined with `include_package_data=True`, this is enough to ensure .version ends up in packages, and that .version ends up in the package when installed.

(Note that things have been working even though it says `pkg_bundle` above rather than `param`...)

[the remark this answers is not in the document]
jlstevens: Ok. I've never been clear on this so I tend to  try everything to make sure!

jlstevens:
Actually, I thought it would be ``autover.py`` which is aliased to ``version`` in ``__init__.py``?

jlstevens:
I see it hasn't been renamed to ``autover.py`` in autover itself either...

ceball:
From param's point of view, since the decision was that autover should continue to be available via param.version whether or not version.py is renamed, I decided to keep things simple.

ceball:
@jbednar I merged this before your review so that I could keep testing. But please still feel free to review :)

jbednar:
> From param's point of view, since the decision was that autover should continue to be available via param.version whether or not version.py is renamed, I decided to keep things simple.

Ok.  The only reason to rename version.py to autover.py is if it would mean that the various packages that rely on it would then be able to be agnostic about whether it came from Param or from a directly installed source.  If the code that is checking for this file (in `__init__.py` and in `setup.py`) is different from other autover-using code anyway because it's importing from param, it's fine to keep calling this file version.py.

Issue #224, opened 2018-04-06
param 1.6.0 prints warning message when used with libraries that use param.version and haven't been updated to the way of using the new param.version


ceball:
<img width="1093" alt="screen shot 2018-04-06 at 7 48 27 pm" src="https://user-images.githubusercontent.com/1929/38438851-a1e1fe4a-39d3-11e8-8ac3-dcd7a29f1cad.png">

jbednar:
It would be nice if only developers would see this message, not users, but I don't know any way to achieve that, so I'm not sure there is anything to fix...

ceball:
I meant, the warning should be pink and blinking, like other warnings ;)

I think we should not be displaying this kind of confusing warning to users. It's just a print...how would a user even turn it off? I actually thought the new version.py was backwards compatible - I must have confused myself. What if I were to make param.version backwards compatible and release 1.6.1? Any objections? (Or hints that it's in some way not possible...)

jbednar:
The new version is *mostly* backwards compatible, but the presence of the arguments that the print statement is testing for indicates that the corresponding calls need updating, because it's going to ignore them.  It would be great if the message would only appear for developers, not users, but as usual I have no idea how that could happen. (I guess we could be very fancy and detect .git in the project directory that's including version, but that's quite a bit magical!)  In any case I'm fine with turning off the message for now, while our own libraries catch up, but in the long run it does seem like something to warn about.

PR #225, opened 2018-04-21
Provide old Version class if called in old way

ceball:
From #224. I've pasted in the old Version class; it gets used if Version() is called in the old way.

Is this kind of thing just a terrible idea/weekend joke, or could it fly? I guess others would rather just forget about the old class and move on...?

### DIFF
--- a/param/version.py
+++ b/param/version.py
@@ -1,121 +1,19 @@
-"""Apart from this first paragraph, this version.py file is an exact
-copy of https://github.com/ioam/autover/blob/ad22b3fcbcdaa40570c3fc8eedb3aeed23b5cdcf/autover/version.py.
-Autover's version.py is included inside the param package only to make
-version.py available to various other projects that use version.py and
-already depend on param, thus saving them from bundling version.py or
-dependending on autover. Otherwise, version.py would only be in the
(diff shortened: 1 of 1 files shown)


jbednar:
Is there a way to do a search of Github to see if anything's actually using it the old way?

ceball:
Not all projects are on github.

And are we ourselves even planning to update all our own projects that use the old way?

Having the option to clean up the new Version class without worrying about projects that are using the old one seems good to me.

However, as I said above, I don't mind just forgetting about it and moving on...

jbednar:
A github search wouldn't be exhaustive, but it would be indicative: if we found dozens of projects using it, I'd say it's worth the effort and confusion of including the duplicative class.  If we found zero (in a fair search), I'd say it's not worth it.

In this case, yes, I do want all of our projects updated to use the new way.

jlstevens:
> In this case, yes, I do want all of our projects updated to use the new way.

That was my thought too. Doesn't mean this PR might not be worth merging though...

It might make sense to have *some* way to warn developers that this is deprecated. E.g when using the methods used for packaging so users aren't affected. Then the old class can be removed eventually...

Issue #233, opened 2018-06-14
API discussion


jlstevens:
PR #230 gives us an opportunity to clean up the namespace of parameterized objects but it also gives us a chance to tidy up the rest of the API: we can define the API we want for the future on ``.param`` while retaining backwards compatibility.

This issue is intended to start some discussion. Here are some of my own notes:

## Notes

* Some API is staying on parameterized objects e.g ``pprint``, ``script_repr``, ``instance`` (for parameterized functions) and ``.param`` itself of course. There is also a ``.initialized`` attribute that I think should be private. What should be exposed at this level?
* Do we want to keep ``script_repr`` in future at the parameterized level?
* The way ``.params`` sometimes returns a dictionary and sometimes does not has always seemed odd to me.
* ``set_default`` seems odd as you might want to set other parameter attributes too. Shouldn't there be a more general way to set *any* parameter attribute?
* Generally there is a fair bit of redundancy I think could be eliminated.

What are your own thoughts on this issue?

jbednar:
1. I vote strongly for retaining `.param()` and `.instance()`, and for making `.initialized` private unless someone finds some clear reason not to.  I don't have any better suggestion than `.param`, so I vote to keep that as well.  I vote weakly to keep `.pprint()` and `.script_repr()`, but would be happy to rename them if a better name is available.
2. I'd be happy to hide `script_repr` under `.param`, though the implementation would need to stay on Parameterized (e.g. as `._script_repr`).
3. For me to have an opinion about `.params`, I'd have to see a proposed alternative (e.g. two methods?).
4. I don't know where `set_default` is used or what it's used for, but yes, it does seem like a more general mechanism should be available.
5. Presumably someone would need to have a task of going through and looking for such redundancy and proposing how to eliminate it.

jlstevens:
> I vote strongly for retaining .param() ...

Do you mean ``.params(...)`` or ``.param`` (the namespace sub-object, no parentheses)? Or are you suggestion that by making ``Parameters`` callable it could offer a useful API that way?

jbednar:
Under 1, I'm talking about the namespace subobject, as I think you were.  Under 3. I'm talking about `.params()`, where we should drill down a bit with two possible replacements for `.param.params()`'s functionality of returning one parameter by name or a dictionary of all parameters:

3a. Rename `.param.params()` to just `.param()` but otherwise leave it unchanged (returning a parameter if one name is specified, and otherwise the full dictionary)
[...]

PR #235, opened 2018-06-25
param namespace fixups

ceball:
* Restores `_set_name()` to Parameterized (still deprecated).
* Restores `state_push()` and `state_pop()` to Parameterized - and don't deprecate - because subclasses may be overriding. 

If we want to change/remove `Parameters._set_name()` and/or `Parameterized.state_push/pop()`, that should be discussed separately (e.g. #233).


jlstevens:
Looks good to me. Ready to merge?

ceball:
Yes, I'm finished

Issue #240, opened 2018-07-13
Regular expression for 'String' Parameter?


thomasdeneux:
Hello,
- First the 'String' Parameter sub-class is not present in the doc, and i initially thought that there is no such sub-class because i was looking in param/__init__.py instead of param/parameterized.py
- Second, a nice idea would be to add regular expression matching to the class. In fact I am trying to add this right now, will send a pull request.

thomasdeneux:
I have now addressed the second bullet point by adding regular expression testing and submitting a pull request.

jbednar:
Thanks!  We should make String more visible in the docs; yes, it's hard to notice in the code, because it's one of the few most basic Parameter types that we had to define in parameterized.py rather than postponing to `__init__.py`.

RegEx matching is a great feature to have, thanks!

ceball:
Thanks for addressing this in #241.

Yes, param's docs need to be improved. It should be happening soon as part of all pyviz project documentation being improved/unified.

String's there, but hidden away: http://param.pyviz.org/Reference_Manual/param.html#param.parameterized.String. You're right that it's missing from the class diagram at the top, though.

PR #241, opened 2018-07-13
Added regular expression to 'String' Parameter

thomasdeneux:
Hello,

As mentioned in issue #240, i have added regular expression matching to the String Parameter sub-class.
I also added relative API0 and API1 tests, but was only able to run the API0 tests.

(In addition i corrected typos in the Number Parameter sub-class)

param is a great package so I was happy to add this small contribution

Best, 
Thomas Deneux

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -546,7 +546,7 @@ class Number(Dynamic):
     When not being dynamically generated, bounds are checked when a
     Number is created or set. Using a default value outside the hard
     bounds, or one that is not numeric, results in an exception. When
-    being dynamically generated, bounds are checked when a the value
+    being dynamically generated, bounds are checked when the value
     of a Number is requested. A generated value that is not numeric,
     or is outside the hard bounds, results in an exception.
 
@@ -563,7 +563,7 @@ class Number(Dynamic):
     allows, for instance, a GUI to know what values to display on
     sliders for the Number.
 
-    Example of creating a Number::
+    Example of creating a Number:
       AB = Number(default=0.5, bounds=(None,10), softbounds=(0,1), doc='Distance from A to B.')
     """
 
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -519,24 +519,36 @@ class Parameter(object):
 # Define one particular type of Parameter that is used in this file
 class String(Parameter):
     """
-    A simple String parameter.
+    A String Parameter, with a default value and optional regular expression.
+
+    Example of creating a String:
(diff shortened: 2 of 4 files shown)


jlstevens:
Allowing a regular expression to further validate allowable strings does make sense to me!

param/__init__.py
```diff
@@ -563,7 +563,7 @@ class Number(Dynamic):
     allows, for instance, a GUI to know what values to display on
     sliders for the Number.
 
-    Example of creating a Number::
+    Example of creating a Number:
```
jbednar: @jlstevens, @ceball, @philippjfr, does anyone know if the :: syntax is still required here?  I don't think that was a typo; it was originally for a docstring interpreted as reStructuredText, where :: introduces a code example. But are our docstrings currently being interpreted as rst, or is this a holdover from some long-ago doc building system?

param/parameterized.py
```diff
@@ -519,26 +519,38 @@ def __setstate__(self,state):
 # Define one particular type of Parameter that is used in this file
 class String(Parameter):
     """
-    A simple String parameter.
+    A String Parameter, with a default value and optional regular expression.
+
+    Example of creating a String:
+      ip_regexp = '^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$'
+      IP = String(default='0.0.0.0', regexp=ip_regexp, doc='An IP address.')
```
jbednar: Good example, though in practice I'd probably recommend making a subclass in this case if you are using IP addresses in many locations.  E.g.:

```
class IPAddress(String):
    "IPV4 address as a string"
    def __init__(self, default="0.0.0.0", **kwargs):
       super(IPAddress, self).__init__(default=default, regexp=ip_regexp, **kwargs)
```

That way you declare that your String is actually an IP address, which won't be evident otherwise. Plus it lets you define what the default value is.

thomasdeneux:
Thanks. Regarding the :: i assumed that it was a typo and have no idea about the correct syntax!

param/__init__.py
jlstevens: Personally I know nothing about using :: so it looks like a typo to me...

jbednar: It's definitely [valid rst](http://docutils.sourceforge.net/docs/user/rst/quickref.html#literal-blocks); I just don't know if any of our tools are interpreting it as rst. Hopefully @ceball knows that.

jbednar: Ok, from discussing with @ceball it's clear that at least some tools are indeed able to process the rst, so please back out this one-character change.

ceball:
Thanks for your contribution, @thomasdeneux !

PR #253, opened 2018-08-25
Add simplified param.depends and watch mechanisms

philippjfr:
Supersedes https://github.com/ioam/param/pull/250 by making a number of simplifications:

* No ``viewable`` argument for ``param.depends``
* No ``cache`` argument for ``param.depends`` (can be added back later)
* Renamed ``eager`` keyword ``watch`` in ``param.depends``

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1102,7 +1102,7 @@ class Parameters(object):
             return MInfo(inst=inst,cls=cls,name=attr,mthd=getattr(src,attr))
 
 
-    def watch(self_,parameter_name,parameter_attribute=None,fn=None):
+    def _watch(self_,action,fn,parameter_name,parameter_attribute=None):
         #cls,obj = (slf_or_cls,None) if isinstance(slf_or_cls,ParameterizedMetaclass) else (slf_or_cls.__class__,slf_or_cls)
 
         assert parameter_name in self_.cls.params()
@@ -1116,19 +1116,25 @@ class Parameters(object):
                 subscribers[parameter_name] = {}
             if parameter_attribute not in subscribers[parameter_name]:
                 subscribers[parameter_name][parameter_attribute] = []
-            subscribers[parameter_name][parameter_attribute].append(fn)
+            getattr(subscribers[parameter_name][parameter_attribute],action)(fn)
         else:
             subscribers = self_.cls.params(parameter_name).subscribers
             if parameter_attribute not in subscribers:
                 subscribers[parameter_attribute] = []
-            subscribers[parameter_attribute].append(fn)
+            getattr(subscribers[parameter_attribute],action)(fn)
 
+    def watch(self_,fn,parameter_name,parameter_attribute=None):
+        self_._watch('append',fn,parameter_name,parameter_attribute)
 
+    def unwatch(self_,fn,parameter_name,parameter_attribute=None):
+        self_._watch('remove',fn,parameter_name,parameter_attribute)
+
+    # TODO: now unused?
     # TODO: event_type (e.g. set, change)
     def subscribe(self_,mthd_name,*callbacks):
         for p in self_.self_or_cls.param.params_depended_on(mthd_name):
             for c in callbacks:
-                (p.inst or p.cls).param.watch(p.name,p.what,c)
+                (p.inst or p.cls).param.watch(c,p.name,p.what)
 
 
     # Instance methods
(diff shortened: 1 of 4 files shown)


param/parameterized.py
```diff
@@ -192,6 +193,66 @@ def __get__(self, obj, type_=None):
...
 
 
+def _getattrr(obj, attr, *args):
+    def _getattr(obj, attr):
+        return getattr(obj, attr, *args)
+    return reduce(_getattr, [obj] + attr.split('.'))
+
+
+# (thought I was going to have a few decorators following this pattern)
+def accept_arguments(f):
```
jbednar: Would be helpful to have docstrings on some of these new functions

param/parameterized.py
```diff
@@ -424,6 +487,35 @@ def _set_instantiate(self,instantiate):
...
+
+        super(Parameter, self).__setattr__(name, value)
+
+        if old is not NotImplemented:
+            for subscriber in self.subscribers[name]:
+                subscriber(Change(what=name,attribute=self._attrib_name,obj=None,cls=self._owner,old=old,new=value))
+
+    # TODO: this is a python 3.6+ thing; we can presumably do this in
+    # Parameterized.__new__ for older pythons. (And merge with objtype
+    # slot.)
```
jbednar: @ceball, can you work out what to do here for when not in 3.6?

param/parameterized.py
```diff
@@ -424,6 +487,35 @@ def _set_instantiate(self,instantiate):
...
+    #
+    # (not specific to this change, but regarding attrib_name, owner:
+    # what if someone re-uses a parameter object across different
+    # classes? maybe we should raise an error if attrib name,owner
+    # already set)
```
jbednar: Seems like that would be polite of us but is not required for merging.

param/parameterized.py
```diff
@@ -472,24 +564,38 @@ def __set__(self,obj,val):
         object stored in a constant or read-only Parameter (e.g. the
         left bound of a BoundingBox).
         """
+        # TODO: simplify this method!
```
jbednar: Or document it, at least.

param/parameterized.py
```diff
@@ -979,6 +1085,65 @@ def inspect_value(self_,name): # pylint: disable-msg=E0213
 
         return value
 
+
+    def params_depended_on(self_,name):
```
jbednar: This and several more below need docstrings...

param/parameterized.py
```diff
@@ -1553,6 +1718,19 @@ def __init__(self,**params):
...
         object_count += 1
 
+        # TODO: should move to param namespace? (like _param_value
+        # etc)
+        self._param_subscribers = {}
+
+        # add watched dependencies
+        #
+        # TODO: This isn't great. Also note anything here will happen
+        # for every instantiation.
```
philippjfr: This bit seems to be pretty expensive, it seems to take about 4x longer than before to instantiate a parameterized class.

ceball:
With #255, does this cover what yuo need for now?

When merged, I'll make an issues about things that must be cleaned up before release, and things that could be improved/added in the future.

philippjfr:
>With #255, does this cover what yuo need for now?

Yes I think so. Looking forward to seeing this merged.

ceball:
Who's handling that - me or you? I don't mind either way, just let me know if I should do it...

philippjfr:
>Who's handling that - me or you? I don't mind either way, just let me know if I should do it...

Missed your question about subscribe on the other PR. Happy to merge both once that's deleted.

param/parameterized.py
ceball: Did it get any better? I'd expect it to be a bit better than before (and minimal impact when there's no watching), but there are still improvements I plan to make (future PR).

param/parameterized.py
jlstevens: I'm also not sure about the names yet but we can iterate on that outside this PR.

ceball: I agree - will be part of an issue covering things that need to be sorted out before the next release.

jlstevens:
Looking at the diff I do have a number of suggestions to make. To avoid slowing things down, I won't write anything down just yet so that we can merge sooner rather than late.

ceball:
> Looking at the diff I do have a number of suggestions to make

I have some suggestions too. But yes, I think until we've settled on the functionality we want, we should hold off. Unless you're talking about something that's an error...

philippjfr:
I just double checked and after updating ``panel`` and my holoviews PR everything still works great.

PR #254, opened 2018-08-28
Dependencies: support more pythons and begin storing dependency info in more convenient way

ceball:
* Support python 2.7 and python3 <3.6
* Keep list of watched methods on class (avoids having to search all attributes at instantiation time)

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1281,6 +1281,7 @@ class ParameterizedMetaclass(type):
         for param_name,param in parameters:
             mcs._initialize_parameter(param_name,param)
 
+        # retrieve depends info from methods and store more conveniently
         dependers = [(n,m._dinfo) for (n,m) in dict_.items()
                      if hasattr(m,'_dinfo')]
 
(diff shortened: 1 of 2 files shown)


philippjfr:
This looks great, ready to merge?

ceball:
yes (the TODOs are part of the overall PR; will turn them into issues when that PR is merged)

param/parameterized.py
```diff
@@ -437,10 +431,15 @@ class Foo(Bar):
...
-    # (which will be filled in by ParameterizedMetaclass)
+    # Note: When initially created, a Parameter does not know which
+    # Parameterized class owns it, nor does it know its names
+    # (attribute name, internal name). Once the owning Parmaeterized
+    # class is created, _owner, _attrib_name, and _internal name are
+    # set.
+
+    # TODO regarding _attrib_name, _owner: what if someone re-uses
+    # a parameter object across different classes? we should raise
+    # an error if attrib name,owner already set
```
jbednar: Raising an error sounds good, and the message could tell people how to avoid the message (by cloning the Parameter object somehow before using it in a new class.  

I'm trying to figure out how to make it reasonable to add an _owner_instance slot, but it ties my brain in knots.  Basically, I want to be able to ask a Parameterized instance for an object that has all the information in the Parameter, plus a pointer to the actual instance so that the value can be accessed.  Yet there's only one Parameter object per class, not per instance, so I guess it would have to be a copy of the Parameter object?  And then it wouldn't be the one that actually applies to this object.  So I guess that before having _owner_instance, we'd first have to set up allowing a local instance copy of the Parameter to override the class's Parameter, at which point the real Parameter just for this instance could be returned with _owner_instance filled in?

Issue #251, opened 2018-08-20
Enhancement: allow instantiation of new Parameterized instances with other instances


sdrobert:
Hello,

First off, thanks, devs, for a really cool package.

In my workings with ``param``, I've found it really useful to separate common parameters into multiple ``Parameterized`` classes and instances, then combine them later. A specific use case is when working with training, development, and test sets in machine learning. You wouldn't expect them to have the same number of samples, but they likely have a number of shared parameters (imagine images or something). You could create three instances of a ``Parameterized`` class and configure them separately, but if I'm somehow exposing those instances to the end user, I'd like to ensure that the common parameters remain the same across instances. Instead, I separate the parameters into two subclasses: one containing population (common) parameters; the other containing set-specific parameters, like so:

``` python
class PopulationParams(param.Parameterized):
    image_width = param.Number(10, bounds=(1, None))
    image_height = param.Number(10, bounds=(1, None))
    # etc...

class SetParams(param.Parameterized):
    num_samples = param.Number(1000, bounds=(1, None))
    # etc...

class SampleParams(PopulationParams, SetParams):
     pass
```

Where the last class can be used to combine parameters back together. I prefer this pattern because it has safety-by-design. AFAICT there's no convenient way of getting the parameters of the instantiations of each uppermost class into the lowermost class when instantiating it. Right now, I resort to something like the following
``` python
param_list = []
for params in (pop_params, set_params):
    param_list += ((k, getattr(params, k)) for k in params.params() if k != 'name')
samp_params = SampleParams(**dict(param_list))
```
Which is serviceable, but not elegant.

I think it would be nice to pass a variable number of ``Parameterized`` instances as positional arguments to any ``Parameterized`` initializer to have it populate its values with those from the instances. The above code could be then simplified to
``` python
samp_params = SampleParams(pop_params, set_params)
```
Overlapping entries can be clobbered in the order the instances were passed to the initializer.

Let me know what you think!

Thanks for the time,
Sean

jbednar:
Glad you find it useful!  From your description it sounds like what you want is just the normal way that parameters already are inherited by subclasses, but maybe I'm just not reading closely enough:

```
class PopulationParams(param.Parameterized):
    image_width = param.Number(10, bounds=(1, None))
    image_height = param.Number(10, bounds=(1, None))
    # etc...

class SetParams(PopulationParams):
    num_samples = param.Number(1000, bounds=(1, None))
    # etc...
```

[...]

Issue #159, opened 2017-06-15
Add json/yaml serialization


ceball:
Recently, PR #153 was opened, showing the desire for json/yaml serialization of parameterized objects.

My reading of #153's serialization: 
1. does not appear to include information about the class (when you deserialize, I think you have to pass in the class)
2. does not include non-parameter attributes
3. does not provide a way to control serialization
4. assumes you only want to write to a file

I think 1, 3, and 4 are necessary.

jbednar:
1. I think the YAML/JSON file definitely needs to include the class name somehow for any subobject; not sure how to best express that in that format. 
2. I'm not certain whether it's useful to include non-parameter attributes, but probably not -- this mechanism is not for serializing the state of a live object, but for saving a specification by which that object could be instantiated just the same as if that object had been specified using a Python constructor call (which would also not currently allow the attributes to be set by the user, only within the constructor.
3. I'm not sure what you mean?
4. The code should definitely support streams, not just named files.

ceball:
Re. 3: e.g. a class can control how it's pickled by implementing ``__getstate__()`` (see https://github.com/ioam/param/wiki/Serialization#controlling-serializationdeserialization).

jbednar:
Ah, you mean that 3. it does not provide classes a way to control their own serialization, or for users to control serialization of specific classes.  

Ok, yes, that's needed for picking; not sure if it's needed for JSON/YAML.  Seems like something that can be added once it's clearly required.

ceball:
Guess I'm thinking about e.g. numbergen.RandomDistribution, which will contain a random.Random object. How would you expect that to appear in json? I think we'll have to provide some way to customize json serialization from the outset just to support what's already in param.

ceball:
Hmm, sticking with random generators as an example to illustrate something I'm not clear about...what exactly is serialization?

Pickle presumably saves a random generator's state (where possible) so it carries on generating where it left off. 

Right now script_repr omits random generators. But maybe it should (I think?) give a representation that would start the same sequence from the beginning again (i.e. capture the parameters of the random generator like seed etc), because that seems most useful given the aim of script_repr. However, that doesn't seem like serialization (it's not capturing the object's state).

[...]

PR #256, opened 2018-08-31
Added param_union, unit tests

sdrobert:
This is the results of the discussion from #251. A helper function ``param_union`` was added to ``param/__init__.py`` that allows new parameter objects to be easily initialized from older ones. The behaviour is mostly the same to that discussed, except

- Supports API 1
- Overwrites the value instead of preserving the first

Also added some basic unit tests.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -103,14 +103,19 @@ def named_objs(objlist):
     return objs
 
 
-def param_union(*parameterizeds, warn=True):
+def param_union(*parameterizeds, **kwargs):
     """
     Given a set of Parameterized objects, returns a dictionary
     with the union of all param name,value pairs across them.
-    If warn is True, warns if the same parameter has been given
-    multiple values; otherwise uses the last value
+    If warn is True (default), warns if the same parameter has
+    been given multiple values; otherwise uses the last value
     """
-    d = {}
+    warn = kwargs.pop('warn', True)
+    if len(kwargs):
+        raise TypeError(
+            "param_union() got an unexpected keyword argument '{}'".format(
+                kwargs.popitem()[0]))
+    d = dict()
     for o in parameterizeds:
         for k, p in o.param.params().items():
             if k != 'name':
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -21,6 +21,7 @@ import sys
 import glob
 import re
 import datetime as dt
+import warnings
 
 from .parameterized import Parameterized, Parameter, String, \
      descendents, ParameterizedFunction, ParamOverrides
@@ -101,6 +102,22 @@ def named_objs(objlist):
         objs[k] = obj
     return objs
 
+
+def param_union(*parameterizeds, warn=True):
+    """
+    Given a set of Parameterized objects, returns a dictionary
+    with the union of all param name,value pairs across them.
+    If warn is True, warns if the same parameter has been given
+    multiple values; otherwise uses the last value
+    """
+    d = {}
+    for o in parameterizeds:
+        for k, p in o.param.params().items():
+            if k != 'name':
+                if k in d and warn:
+                    warnings.warn("overwriting parameter {}".format(k))
+                d[k] = getattr(o, k)
+    return d
 
 
 class Infinity(object):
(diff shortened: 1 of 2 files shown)


param/__init__.py
```diff
@@ -102,6 +103,27 @@ def named_objs(objlist):
     return objs
 
 
+def param_union(*parameterizeds, **kwargs):
+    """
+    Given a set of Parameterized objects, returns a dictionary
+    with the union of all param name,value pairs across them.
+    If warn is True (default), warns if the same parameter has
+    been given multiple values; otherwise uses the last value
+    """
```
ceball: Maybe: "If the same parameter is given multiple times, uses the last value (warns by default in this case; pass warn=False to overwrite silently)." ?

param/__init__.py
```diff
@@ -102,6 +103,27 @@ def named_objs(objlist):
...
+    if len(kwargs):
+        raise TypeError(
+            "param_union() got an unexpected keyword argument '{}'".format(
+                kwargs.popitem()[0]))
+    d = dict()
+    for o in parameterizeds:
+        for k, p in o.param.params().items():
+            if k != 'name':
+                if k in d and warn:
+                    warnings.warn("overwriting parameter {}".format(k))
```
ceball: My understanding is that this would show a warning only one time ever (i.e. only the first time any overwrite happens). Unless you definitely know otherwise, could you check?

ceball: By the way, param has its own messaging/warning system. While I don't particularly want to recommend it because it's annoying in many ways, it avoids problems like the above (and using it would be consistent with the rest of param, in general). So maybe `Parameterized(name="param_union").warning("overwriting parameter %s",k)` (untested).

sdrobert: It happens once (and only once) for each parameter the overwrite happens with. So yes, switching to param's warning system makes a lot of sense (I just took precedent from ``version.py``). I'll also update the test for posterity. Is there any way to avoid the 'root' part of the message using your proposed method?
```
WARNING:root:param_union: overwriting parameter a
WARNING:root:param_union: overwriting parameter b
```
I don't know what purpose 'root' serves in general. Please let me know what you think :)

ceball: > It happens once (and only once) for each parameter the overwrite happens with

Thanks for checking and for staying engaged through these seemingly trivial details!

> I just took precedent from version.py

Sorry to confuse you! That file is for versioning rather than being related to parameters; I'm not familiar with its internals.

> Is there any way to avoid the 'root' part of the message using your proposed method
> I don't know what purpose 'root' serves in general

I'm sure there's a way to avoid the 'root' part of the message, but it wouldn't be specific to this work on `param_union()`. I actually wasn't (consciously) aware that 'root' is shown, nor do I know what purpose it serves. param's logging system uses Python's logging module (see https://github.com/ioam/param/blob/master/param/parameterized.py#L32); I think everyone would be happy to see it simplified and/or clarified, but nobody's had time to work on it.

jbednar: 'root' doesn't serve any important purpose; it arises simply because the logging system is designed to log warnings and errors for particular Parameterized classes and instances, where "root" would be replaced with the actual class or instance raising the issue.  Here, it's the "root" or base class that's raising it, as a degenerate case, and we should probably just remove the object name altogether in that case.  That would be a backwards-incompatible change for anyone who searches or filters the log messages, but it's something that could be done in our upcoming shift to Param 2.0.

jbednar: Oops; you're right; I wasn't looking closely enough and I was talking about a completely different problem, i.e. that to get a message not tied to an existing Parameterized, one has to use a dummy instance of Parameterized(), which then shows up in the messages inappropriately.  "root" must be from the logger, as you say, in which case should we be using that bit for the Parameterized instance or class information (collapsing "root" and "param_union" above)?

jbednar: Or maybe we should just declare that the root logger has the name "param", to give people an idea where these messages come from?

sdrobert: Any decisions, gentleman?

Also, anyone feel like taking a look on my comment in #159?

jbednar: I've moved this discussion to https://github.com/ioam/param/issues/267 so that I could merge this PR.

PR #258, opened 2018-08-31
Allow ObjectSelector classes to accept a dictionary with names

jbednar:
Right now, ObjectSelector accepts a list of objects from which to select, which works well when the objects have a meaningful `name` parameter or `__name__` attribute.  However, one often needs to switch between objects for which you want to use a different representation.  For instance if one defines a selector like this: 

```
   video = param.ObjectSelector(objects=["k27MJJLJNT4","aZ1G_Q7ovmc", "BjlQdOK3n5w"])
```

The user will be confronted with a bewildering set of meaningless choices.  Instead, it would be nice to provide a list of readable names associated with each object, as in:

```
    video = param.ObjectSelector(objects=dict(AnacondaCon18="k27MJJLJNT4",
                                              SciPy18="aZ1G_Q7ovmc", 
                                              ERDC17="BjlQdOK3n5w"))
```

That way a GUI selection widget can let the users choose between the readable names while internally selecting between the underlying objects.  

This PR adds this capability.  Now, if a user passes in a dict or an ordereddict for the objects (which previously has been a list), then ObjectSelector stores this dict in a `names` slot, then pulls out the values of the dictionary to use as the normal objects list as before.  The `names` dictionary is then consulted for `get_range()`, which returns the `named_objs` using the names dictionary entries where present, and otherwise the name or other repr.

This approach of keeping a separate `names` dictionary was chosen to improve backwards compatibility, so that the behavior of the existing slots is the same apart from in `get_range`.

This approach has been tested with ParamNB:

image

And with ParamBokeh:

image

However, it looks like Panel currently uses its own machinery for coming up with the names list, and ignores this support in Param:

image

So Panel will need updating to use this functionality, but the other packages will not.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -21,6 +21,7 @@ import sys
 import glob
 import re
 import datetime as dt
+import collections
 
 from .parameterized import Parameterized, Parameter, String, \
      descendents, ParameterizedFunction, ParamOverrides
@@ -85,14 +86,22 @@ def as_unicode(obj):
     return unicode(obj)
 
 
-def named_objs(objlist):
+def named_objs(objlist, namesdict=None):
     """
     Given a list of objects, returns a dictionary mapping from
-    string name for the object to the object itself.
+    string name for the object to the object itself. Accepts
+    an optional name,obj dictionary, which will override any other
+    name if that item is present in the dictionary.
     """
     objs = OrderedDict()
+
+    if namesdict is not None:
+        objtoname = {v: k for k, v in namesdict.items()}
+
     for obj in objlist:
-        if hasattr(obj, "name"):
+        if namesdict is not None and obj in objtoname:
+            k = objtoname[obj]
+        elif hasattr(obj, "name"):
             k = obj.name
         elif hasattr(obj, '__name__'):
             k = obj.__name__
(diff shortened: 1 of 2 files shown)


philippjfr:
Would be nice to see some tests here, otherwise this looks good. I'll leave it up to you or @ceball to merge.

jbednar:
Good point; added some tests.

param/__init__.py
```diff
@@ -85,14 +86,22 @@ def as_unicode(obj):
...
     Given a list of objects, returns a dictionary mapping from
-    string name for the object to the object itself.
+    string name for the object to the object itself. Accepts
+    an optional name,obj dictionary, which will override any other
+    name if that item is present in the dictionary.
     """
     objs = OrderedDict()
+
+    if namesdict is not None:
+        objtoname = {v: k for k, v in namesdict.items()}
```
ceball: This is slightly confusing to me: this is "invert the dict"? Should I worry about a dictionary value being unsuitable for use as a key? (Or multiple keys pointing to the same object? Although that one seems a bit weird if I think of a selector in a gui, maybe it'd be ok in general?)

jbednar: Yes, this is inverting the dict, because it needs to find the name for a given object, but the function accepts a mapping from name to object, which is what the user would expect.

Yes, this will be limited to hashable objects; I'll update the docstring.  If the user supplies multiple keys pointing to the same object, I'd say that's their own problem. :-)

PR #259, opened 2018-09-10
Batched parameter setting for watch mechanism

jlstevens:
This PR is *very* WIP and I don't think it should be merged without a lot of work. All the first commit shows is that this should be possible, *hopefully* without requiring very deep changes to how watching works.

Major Issues that need to be addressed:

- [x] Uses a ``groupby`` on the callback identity: this is not a safe assumption (I believe this is fixed)
- [ ]  The `DISABLE` global variable should be moved the the namespace object and renamed.
- [ ] I'm assuming the 'what' is 'value' which I *think* is ok for `set_param` (as you can only set value there)
- [ ] Probably others I haven't thought of yet!

Instance example:

```python
import param
   a = param.Number(default=1)
   b = param.Number(default=3)

def foo(*la): print('CHANGES>' , la)
f = Foo()
f.param.watch(foo, ['a','b'])

f.set_param(a=32, b=33)
CHANGES> (Change(what='value', attribute='a', obj=Foo(a=32, b=33, name='Foo00002'), cls=<class '__main__.Foo'>, old=1, new=32), Change(what='value', attribute='b', obj=Foo(a=32, b=33, name='Foo00002'), cls=<class '__main__.Foo'>, old=3, new=33))
```

Class example:

```python
import param
   a = param.Number(default=1)
   b = param.Number(default=3)

def foo(*la): print('CHANGES>' , la)
Foo.param.watch(foo, ['a','b'])

Foo.set_param(a=455, b=33)
CHANGES> (Change(what='value', attribute='a', obj=<class '__main__.Foo'>, cls=<class '__main__.Foo'>, old=455, new=455), Change(what='value', attribute='b', obj=<class '__main__.Foo'>, cls=<class '__main__.Foo'>, old=33, new=33))
```

I'm also not a huge fan of these `Change` objects. I think they should be an option you explicitly enable to get as much information as possible: I think the default should be to just supply the new values on a change as kwargs.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -248,6 +248,7 @@ def _m_caller(self,n):
 PInfo = namedtuple("PInfo","inst cls name pobj what")
 MInfo = namedtuple("MInfo","inst cls name mthd")
 Change = namedtuple("Change","what attribute obj cls old new")
+Subscriber = namedtuple('Subscriber', 'batched fn')
 
 
 class ParameterMetaclass(type):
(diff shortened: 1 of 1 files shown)


jlstevens:
In a8f83d0 I define a `Subscriber`  named tuple which should avoid the problem with the ambiguous `groupby`.

jlstevens:
This PR was a useful prototype but it probably isn't worth resolving the merge conflicts here now that the watch mechanism has been further improved. In particular the semantics of ``watch`` has now been tightened so I think I'll close this PR and make a new PR based on those improvements.

PR #260, opened 2018-09-10
param_depends_on returns all parameters for methods

philippjfr:
This PR ensures that undecorated methods by default depend on all parameters. This PR will ensure that ``panel`` will listen to all parameter changes by default while holoviews will continue to only listen to changes that have been explicitly declared.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -230,7 +230,8 @@ def depends(func, *dependencies, **kw):
 
 
 def _params_depended_on(mthing,params):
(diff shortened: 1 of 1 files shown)


jbednar:
Looks good, thanks!

kcpevey:
@philippjfr is there an example of the depends behavior somewhere?

philippjfr:
@kcpevey Sorry missed your comment, I'll add some docs for panel shortly.

PR #263, opened 2018-09-13
WIP/demo: Make subscribers/watchers wait

ceball:
I started this some time ago at the end of doing the depends work to handle batching, but only just now made it into enough of a demo. The work in this PR is really not thought through (does the idea even make sense?), and isn't how I'd want it to be even if the idea makes sense, but the example might be enough to see if the idea is worthwhile pursuing or not...

Just make watchers wait (see "subscribers_wait" context manager at the end):

screen shot 2018-09-13 at 9 43 20 pm

Make watchers wait and process all changes together:

screen shot 2018-09-13 at 9 43 47 pm

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1133,8 +1133,19 @@ class Parameters(object):
 
     # Instance methods
 
+    # TODO: instance only for now (just a demo)
     @contextmanager
-    def watchers_wait(self_):
+    def subscribers_wait(self_,batch_fn=None):
+        """
+        no batch_fn: subscribers don't get called until the end, and each gets
+                     only the last relevant change
+
+        batch_fn: subscribers don't get called at all; instead,
+                  batch_fn(subscribers,changes) is called and can do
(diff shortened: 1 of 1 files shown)


jlstevens:
This idea is closely related to something I need for #259 in `set_param` and I suspect this proposal is syntactic sugar for the same concept (personally, I would prefer parallel setting via `set_param` than using a context manager).

ceball:
Yes, I guess I was kind of dumping this earlier start on you, for you to figure out and decide about ;)

ceball:
(The idea came from seeing what traitlets and/or traits does, kind of, I think.)

jlstevens:
This PR has some useful ideas to consider that can be implemented on top the initial batched watching  mechanism implemented here in #281. @ceball if you are happy with closing this PR, I can continue to refer to it as I further extend batched watching...

philippjfr:
I think we can safely close. Can always resurrect any ideas from the branch.

PR #264, opened 2018-09-18
Param depends improvements

philippjfr:
Refactor of param depends and added support for listening to all subobject parameters with ``subobject.param`` spec.

- [x] Add unit tests


philippjfr:
The new tests seem to be relying on non-deterministic ordering, will fix tomorrow.

philippjfr:
@ceball A quick review would be appreciated. Otherwise ready to merge.

PR #266, opened 2018-09-18
Fixed Parameter __setstate__ bug

philippjfr:
This is a somewhat subtle bug and I have no minimal example except for trying to run through the panel Param notebook and trying to set ``Example.unbounded_float = 3.14``.

- [x] Fixes https://github.com/ioam/param/issues/265

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -498,7 +498,16 @@ class Parameter(object):
     # to the Parameterized instance, so no per-instance subscription.
 
     def __setattr__(self,name,value):
-        old = getattr(self,name) if (name!="default" and hasattr(self,'subscribers') and name in self.subscribers) else NotImplemented
+        implemented = (name!="default" and hasattr(self,'subscribers') and name in self.subscribers)
+        try:
+            old = getattr(self,name) if implemented else NotImplemented
+        except AttributeError as e:
+            if name in self.__slots__:
+                # If Parameter slot is defined but an AttributeError was raised
+                # we are in __setstate__ and subscribers should not be triggered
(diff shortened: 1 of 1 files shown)


ceball:
I can't do anything about this right now or review in detail, so excuse the
brevity, but...

(a) That method is just a hack to allow watching of parameter metadata
(i.e. things other than value) - do we still need that? (If not, we could
remove it and leave metadata watching as something to consider in the
future.)

(b) I imagine there are various times when watching should be suspended.
Pickling, copying, initialization, etc. Respecting the 'initialized' flag
or whatever it is is one of the to-do items, and will appear in the
"depends cleanup" issue. (However, no doubt the __setattr__ method is bad
anyway...)

(c) I can't immediately think what the problem would be with your minimal
example, i.e. how it would lead to the traceback you describe in the
corresponding issue. I'd need to run it (or if you have a session where you
already ran it, maybe you could copy/paste that...?)

philippjfr:
>That method is just a hack to allow watching of parameter metadata
(i.e. things other than value) - do we still need that? 

Yes, I use it in panel and it works well.

>I'd need to run it (or if you have a session where you
already ran it, maybe you could copy/paste that...?)

The example above leads directly to the traceback, it occurs when setting the parameter value on a subclass which did not define the parameter, because this triggers a ``copy`` operation on the parameter which in turn triggers ``__setstate__``.

ceball:
By the way, I'd be fine with this being merged and a new dev release being made so you can proceed with panel. 

We can discuss more if necessary when reviewing the whole mechanism later on.

PR #268, opened 2018-09-20
ObjectSelector support for selecting lists and dicts

jbednar:
See https://github.com/pyviz/panel/pull/50 for more details.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -96,12 +96,15 @@ def named_objs(objlist, namesdict=None):
     """
     objs = OrderedDict()
 
+    def hashable(x):
+        return tuple(x) if isinstance(x, collections.Sequence) else x
+            
     if namesdict is not None:
-        objtoname = {v: k for k, v in namesdict.items()}
+        objtoname = {hashable(v): k for k, v in namesdict.items()}
 
     for obj in objlist:
-        if namesdict is not None and obj in objtoname:
-            k = objtoname[obj]
+        if namesdict is not None and hashable(obj) in objtoname:
+            k = objtoname[hashable(obj)]
         elif hasattr(obj, "name"):
             k = obj.name
         elif hasattr(obj, '__name__'):
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -87,6 +87,14 @@ def as_unicode(obj):
     return unicode(obj)
 
 
+def hashable(x):
+    if isinstance(x, collections.MutableSequence):
+        return tuple(x)
+    elif isinstance(x, collections.MutableMapping):
+        return tuple([(k,v) for k,v in x.items()])
+    else:
+        return x
+            
 def named_objs(objlist, namesdict=None):
     """
     Given a list of objects, returns a dictionary mapping from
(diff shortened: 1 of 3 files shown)


param/__init__.py
```diff
@@ -87,6 +87,22 @@ def as_unicode(obj):
...
+    dictionaries converted to tuples.  Allows mutable objects to be
+    used as a lookup key in cases where the object has not actually
+    been mutated. Lookup will fail (appropriately) in cases where some
+    part of the object has changed.  Does not (currently) recursively
+    replace mutable subobjects.
+    """
+    if isinstance(x, collections.MutableSequence):
+        return tuple(x)
+    elif isinstance(x, collections.MutableMapping):
+        return tuple([(k,v) for k,v in x.items()])
```
ceball: I know what you're doing (I think), but the terminology might be confusing in general.

> An object is hashable if it has a hash value which never changes during its lifetime (it needs a __hash__() method), and can be compared to other objects (it needs an __eq__() method). Hashable objects which compare equal must have the same hash value.

(https://docs.python.org/3/glossary.html#term-hashable)

E.g. could presumably have two dicts x and y which compare equal but for which hash(hashable(x)) is not the same as hash(hashable(y)). You could sort, maybe. Can't quite think through all the consequences for what you're doing or in general either wya!

jbednar: The reason for the name `hashable()` is simply that when applied, something that Python says is not hashable becomes hashable:

```
>>> i=0
>>> t=(1,2,3)
>>> l=[1,2,3]
>>> d=dict(one=1,two=2,three=3)
>>> reverse_items = {v:k for k,v in {"i":i, "t":t}.items()}
>>> reverse_items = {v:k for k,v in {"i":i, "t":t, "l":l}.items()}
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 1, in <dictcomp>
TypeError: unhashable type: 'list'
>>> reverse_items = {v:k for k,v in {"i":i, "t":t, "d":d}.items()}
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 1, in <dictcomp>
TypeError: unhashable type: 'dict'
>>> from param import hashable
>>> reverse_items = {v:k for k,v in {"i":i, "t":t, "l":hashable(l), "d":hashable(d)}.items()}
>>>
```
So I don't think there can be any disagreement that the result of `hashable()` is in fact hashable. 

But I think your point is that it may not work well as a hash *for the original object* in all cases, which is true (and which is presumably why Python doesn't already do this in general!).  For the special but very common case of a list of literals, I don't think there is any issue; the ordering should be the same regardless, and if any item in the list is added, removed, or change it won't match, as designed.  For dictionaries you're right that any ordering changes will make the comparison fail.  I don't know if there are cases where the ordering will change without at least *some* change to the object (e.g. adding then deleting an item), and I think it's fair for people to assume that if they have changed the object in any way it will no longer match its name (in which case they will get a default str-based representation).  So I'm not sure it's worth the extra step of adding sorting, but if you want to do so, then I certainly won't object.

PR #270, opened 2018-09-20
Renamed attributes of named tuples used by watch

jlstevens:
Simple PR that fixes what I consider incorrect terminology relating to `Change` objects and the other new named tuples.


jbednar:
As long as those new names are accurate, I greatly prefer them to the old. Thanks!

jlstevens:
I would recommend merging this PR now. I would be very grateful if @ceball could review this to make sure that the new names are indeed accurate. If not, I'm happy to rename again!

PR #272, opened 2018-09-21
Added watch_values and unwatch_values methods

jlstevens:
Using `watch_values` is useful when you only care about value changes and don't want to handle `Change` objects. In this case, the changed values are passed as `**kwargs`.

Ready for review/merge once the tests are green.


philippjfr:
Is ``unwatch_values`` definitely necessary? Seems to me ``unwatch`` could handle both.

PR #274, opened 2018-09-21
Added utility to test logging output

jlstevens:
This PR moves ``MockLoggingHandler`` from HoloViews to param where it is more appropriate. One example unit test is also added here to show how it is used. Once a new version of param is release, I'll be able to remove the equivalent code from HoloViews.


jlstevens:
Ready to merge if you are happy with it!

PR #277, opened 2018-09-24
Watch changed comparison only compares safe types

philippjfr:
For safety the `_changed` value comparison which checks whether a watcher should be called only attempts to compare safe types.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -8,6 +8,7 @@ import re
 import sys
 import inspect
 import random
+import numbers
 
 from collections import namedtuple
 from operator import itemgetter,attrgetter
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good to me!

param/parameterized.py
```diff
@@ -804,19 +817,19 @@ def inner(*args, **kwargs):
...
-
     @classmethod
     def _changed(cls, change):
         """
         Predicate that determines whether a Change object has actually
         changed such that old!=new.
         """
         try:  # To be improved by adding better machinery to test equality for complex types
+            if not is_comparable(change.new) or not is_comparable(change.old):
+                return True
```
jbednar: Hmm.  Maybe instead of having a separate is_comparable test and a hope of adding better machinery, those two things should be combined so that this class has a class attribute that stores the mechanism used to decide if it's changed, which currently would use is_comparable + the bool() test, but which someone could replace with a more full-featured test?

philippjfr: If we decide to implement a more full-featured approach wouldn't we be adding it to param anyway and replacing this code?

jbednar: Eventually, sure, but it would (a) help us develop it without having to tightly link the param version with other code versions (which has been painful over the past few weeks), and (b) this would allow other people who are doing this to develop and use their own such code if it came up.  My point is just that if we are already abstracting this, may as well go all the way and make it pluggable.

philippjfr: I have no objection. I was just thinking that implementing a way to declare equality functions for different types as I suggested in https://github.com/ioam/param/issues/269 would not take too long to implement and would offer a pluggable mechanism.

jbednar: Also fine.

philippjfr:
I've opened https://github.com/ioam/param/pull/279 to supersede this PR, if we decide to go with the approach implemented there.

jbednar:
Superseded by #279.

PR #278, opened 2018-09-24
Added onlychanged flag for watch method

jlstevens:
We decided we wanted this boolean option independently of whether we implement the `trigger` method or not.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -249,7 +249,7 @@ def _m_caller(self,n):
 PInfo = namedtuple("PInfo","inst cls name pobj what")
 MInfo = namedtuple("MInfo","inst cls name method")
 Change = namedtuple("Change","what name obj cls old new")
-Watcher = namedtuple("Watcher","fn mode")
+Watcher = namedtuple("Watcher","fn mode onlychanged")
 
 class ParameterMetaclass(type):
     """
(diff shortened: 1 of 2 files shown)


param/parameterized.py
```diff
@@ -1167,31 +1167,32 @@ def _watch(self_,action,watcher,parameter_name,parameter_attribute=None):
...
-            watcher = Watcher(fn=fn, mode='args')
-            self_._watch('remove',watcher,parameter_name,parameter_attribute)
-            unwatched = True
-        except: pass
-        try:
-            watcher = Watcher(fn=fn, mode='kwargs')
-            self_._watch('remove',watcher,parameter_name,parameter_attribute)
-            unwatched = True
-        except: pass
+        for onlychanged in [True, False]:
```
philippjfr: Are the watchers hashed based on their onlychanged value or why is this necessary?

philippjfr: Okay I see it, I'm happy to merge now but maybe we could consider changing the ``watchers`` to be a nested `dict` with keys of the form ``watchers[parameter_name][(parameter_attribute, fn)]``. That way you don't need to replicate the whole watcher state on unwatch.

jlstevens: This is so named tuples match when removing from a list. If the value is different for `onlychanged` then it won't match. The alternative is to specify `onlychanged` as an argument which will only remove that type.

PR #279, opened 2018-09-25
Add Comparator to detect changes in arbitrary objects

philippjfr:
Implements what I suggested in https://github.com/ioam/param/issues/269 and supersedes https://github.com/ioam/param/pull/277 if we decide to go with this approach. I haven't yet defined any comparisons for types beyond Python literals. Here's what a comparison function definition for a NumPy array would look like (without explicitly importing numpy and registering the equality by type):

```python
def is_numpy(obj):
    if 'numpy' not in sys.modules: return False
    import numpy as np
    return isinstance(obj, np.ndarray)

def array_equal(obj1, obj2):
    import numpy as np
    return np.array_equal(obj1, obj2)

Comparator.equalities[is_numpy] = array_equal

# If NumPy can be imported it's much simpler
Comparator.equalities[np.ndarray] = np.array_equal
```

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -8,6 +8,8 @@ import re
 import sys
 import inspect
 import random
+import numbers
+import operator
 
 from collections import namedtuple
 from operator import itemgetter,attrgetter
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good to me as well. Merging.

PR #280, opened 2018-09-26
Renamed 'parameter_attribute' to 'what'

jlstevens:
I believe this variable rename makes things a little less confusing and a little more consistent.


philippjfr:
LGTM, will merge when tests pass.

PR #281, opened 2018-09-27
Implementation of batched watching

jlstevens:
At the time of creation, this PR only implements some (hopefully!) non-controversial improvements. The rest of it will implement parameter subset semantics (we called this 'option 3') for batching.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -900,15 +900,16 @@ class Parameters(object):
         settings in kwargs using the queued Change and watcher objects.
         """
         change_dict = OrderedDict([(c.name,c) for c in self_.cls.param._changes])
-        for watcher in self_.cls.param._watchers:
+        watchers = self_.cls.param._watchers[:]
+        self_.cls.param._changes = []
+        self_.cls.param._watchers = []
+        for watcher in watchers:
             changes = [change_dict[name] for name in watcher.parameter_names if name in change_dict]
             if watcher.mode == 'args':
                 watcher.fn(*changes)
             else:
                 watcher.fn(**{c.name:c.new for c in changes})
 
-        self_.cls.param._changes = []
-        self_.cls.param._watchers = []
         self_.cls.param._BATCH_WATCH = False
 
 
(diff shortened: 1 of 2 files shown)


param/parameterized.py
```diff
@@ -874,18 +876,48 @@ def _changed(cls, change):
...
         Invoke the given the watcher appropriately given a Change object.
         """
-        if watcher.onlychanged and (not cls._changed(change)):
+        if watcher.onlychanged and (not self_._changed(change)):
             return
-        if watcher.mode == 'args':
+
+        if self_.cls.param._BATCH_WATCH:
+            self_._changes.append(change)
+            self_._watchers.append(watcher)
```
philippjfr: Maybe:

```python
if watcher not in self_._watchers:
    self_.watchers.append(watcher)
```

Then you don't have to compute an ordered set later.

param/parameterized.py
```diff
@@ -874,18 +876,47 @@ def _changed(cls, change):
...
+
+        for watcher in watcher_set:
+            changes = [change_dict[name] for name in watcher.parameter_names if name in change_dict]
+            if watcher.mode == 'args':
+                watcher.fn(*changes)
+            else:
+                watcher.fn(**{c.name:c for c in changes})
+
+        self_.cls.param._changes = []
+        self_.cls.param._watchers = []
```
philippjfr: I think the queues needs to emptied before you start dispatching since a watch callback may in turn call ``.set_param``, which would mean the queue is reused and the callback would recursively call itself.

jlstevens: Good point.

param/parameterized.py
jlstevens: Good idea...assuming I only ever need a set (which I currently think is true)

jlstevens: Addressed in 7c4b4ce.

param/parameterized.py
jlstevens: Addressed in 75325ae

jlstevens:
@philippjfr There is always more do do but I think this PR gets batching on value working the way we want. If you are happy with it, please merge this PR so we can test it properly.

I'll be working on more PRs to extend/refactor this mechanism (e.g support batched processing of parameter attributes).

philippjfr:
Looks good to me. I'll merge now and will start using batching in panel.

PR #285, opened 2018-10-01
Added pandas DataFrame and Series parameter types

jlstevens:
Currently WIP. As of the first commit, these two parameter types follow the implementation of `param.Array` which is just syntactic sugar for a `ClassSelector`. I'll now be adding options for additional validation.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1329,7 +1329,7 @@ class DataFrame(ClassSelector):
             difference = set(self.columns) - set([str(el) for el in val.columns])
             if difference:
                 msg = 'Provided DataFrame columns {found} does not contain required columns {expected}'
-                raise Exception(msg.format(found=list(val.columns), expected=self.columns))
+                raise Exception(msg.format(found=list(val.columns), expected=sorted(self.columns)))
 
         if self.ordered:
             if list(val.columns) != list(self.columns):
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1314,15 +1314,18 @@ class DataFrame(ClassSelector):
 
     def _check_value(self,val,obj=None):
         super(DataFrame, self)._check_value(val, obj)
+
+        if isinstance(self.columns, set) and self.ordered is True:
+            raise Exception('Columns cannot be ordered when specified as a set')
+
         if self.columns is None:
             pass
         elif (isinstance(self.columns, tuple) and len(self.columns)==2
               and all(isinstance(v, (type(None), numbers.Number)) for v in self.columns)): # Numeric bounds tuple
             self._length_bounds_check(self.columns, len(val.columns), 'columns')
-        elif isinstance(self.columns, list):
-            self.ordered = True if self.ordered is None else self.ordered
 
         if isinstance(self.columns, (list, set)):
+            self.ordered = isinstance(self.columns, list) if self.ordered is None else self.ordered
             difference = set(self.columns) - set([str(el) for el in val.columns])
             if difference:
                 msg = 'Provided DataFrame columns {found} does not contain required columns {expected}'
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1373,14 +1373,17 @@ class Series(ClassSelector):
     __slots__ = ['rows']
 
     def _length_bounds_check(self, bounds, length, name):
-        message = 'Length of {name} outside declared bounds of {bounds}'
+        message = '{name} length {length} does not match declared bounds of {bounds}'
+        if not isinstance(bounds, tuple):
+            if (bounds != length):
+                raise ValueError(message.format(name=name, length=length, bounds=bounds))
+            else:
+                return
         (lower, upper) = bounds
-        if lower is not None:
-            if length < lower:
-                raise ValueError(message.format(name=name, bounds=bounds))
-        if upper is not None:
-            if length > upper:
-                raise ValueError(message.format(name=name, bounds=bounds))
+        failure = ((lower is not None and (length < lower))
+                   or (upper is not None and length > upper))
+        if failure:
+            raise ValueError(message.format(name=name,length=length, bounds=bounds))
 
     def __init__(self, rows=None, **params):
         from pandas import Series as pdSeries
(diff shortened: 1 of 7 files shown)


jlstevens:
It looks like pandas isn't playing well with the pypy build. Any suggestions for what I can do about it?

Removing pypy builds is one obvious option, but I would rather not do that.

param/__init__.py
```diff
@@ -1287,6 +1287,73 @@ def __init__(self, **params):
...
+
+        if self.columns is None:
+            pass
+        elif isinstance(self.columns, list):
+            if len(val.columns) != len(self.columns):
+                raise Exception(length_error.format(found=len(val.columns), expected=len(self.columns)))
+            difference = set(val.columns) ^ set(self.columns)
+            if difference:
+                msg = 'Provided DataFrame columns {found} differ from expected columns {expected}'
+                raise Exception(msg.format(found=list(val.columns), expected=self.columns))
```
jbednar: I would argue for a different approach to validating DataFrame columns.  If someone provides a list of the columns they are expecting from the DataFrame, I think we should only validate that those columns are present.  E.g. if `['x','y']` is specified, it should be treated as a declaration that we will be accessing columns `x` and `y` from the provided dataframe, and so we ensure that those columns exist.  Conversely, if someone specifies cols to be `3`, it's a declaration that we will be accessing columns by number, and that there will need to be three of them, regardless of their names.  

For the numeric case it's not clear to me whether 3 should mean "at least 3", or "exactly 3", as there are arguments both ways. The reason for "at least 3" is that it's expressing specifically what we actually require, i.e. it's a declaration that we will access the first three columns by number; any additional columns won't hurt us but won't be accessed unless we first check that they exist. The reason for "exactly 3" would be to catch errors where someone supplies a larger one but doesn't realize that only the first 3 columns will be used. One could also imagine accepting a tuple to make it feasible to express precisely what you mean (like numeric bounds in Param in general, i.e.. with lower and upper bounds specified explicitly).  Unfortunately, a tuple like that could be ambiguous with the list of columns, e.g. when serialized to JSON (losing the distinction between tuples and lists), so I think I'd favor just picking whether a single numeric argument means "exactly" or "at least".  I think my vote is for "at least".

param/__init__.py
```diff
@@ -1287,6 +1287,73 @@ def __init__(self, **params):
...
+        else:
+            if len(val.columns) != self.columns:
+                raise Exception(length_error.format(found=len(val.columns), expected=self.columns))
+
+        if self.rows is None:
+            pass
+        elif len(val) != self.rows:
+            msg = 'Provided DataFrame has {found} rows which does not match the expected {expected} rows'
+            raise Exception(msg.format(found=len(val), expected=self.rows))
```
jbednar: For rows, I can easily imagine wanting to express a minimum, maximum, or exact number, which if implemented for rows may as well be implemented for columns with the same syntax if it doesn't conflict with the specific list of columns support.  The current "exactly 3" approach for rows seems unlikely to be sufficient in general.

param/__init__.py
jlstevens: The columns suggestion is now implemented in 3cf180b.

jlstevens:
@ceball I tried editing tox.ini as you suggested in 0c1bab5 but without success. What am I missing?

ceball:
pandas will only be available in matrix entries that include pandas. I.e. the test suite also needs to be able to run without pandas present, by skipping the pandas-requiring tests.

See e.g. https://github.com/ioam/param/blob/master/tests/API1/testnumpy.py#L10

The additional `PARAM_TEST_NUMPY` environment variable is the best hack I could think of at the time to ensure that the numpy tests don't get skipped silently by mistake when they are supposed to be running.

ceball:
You also unfortunately need to add a corresponding entry to travis: https://github.com/ioam/param/blob/master/.travis.yml#L63

This is the kind of pain that inspired pyctdev to start. (So that there could be a different type of pain...)

jlstevens:
@ceball Thanks for that! I think I got things working correctly between bdb4226 and 05caa10...

jlstevens:
Other than adding a few more unit tests, we just need to resolve the semantics we want for numeric rows/columns values. Or naming 'rows' to length. I'm happy to implement things whichever way as long as @jbednar and @philippjfr are in agreement.

param/__init__.py
jbednar: Looks good, thanks.  @philippjfr?

param/__init__.py
```diff
@@ -1287,6 +1287,112 @@ def __init__(self, **params):
         super(Array,self).__init__(ndarray, allow_None=True, **params)
 
 
+class DataFrame(ClassSelector):
+    """
+    Parameter whose value is a pandas DataFrame.
```
jbednar: Docstring explaining the arguments accepted?

param/__init__.py
```diff
@@ -1287,6 +1287,112 @@ def __init__(self, **params):
...
+                raise Exception(message.format(name=name, length=length, bounds=bounds))
+            else:
+                return
+        (lower, upper) = bounds
+        if lower is not None:
+            if length < lower:
+                raise Exception(message.format(name=name, length=length, bounds=bounds))
+        if upper is not None:
+            if length > upper:
+                raise Exception(message.format(name=name,length=length, bounds=bounds))
```
jbednar: Probably clearer if expressed a single if statement, with one test expression?

param/__init__.py
```diff
@@ -1287,6 +1287,112 @@ def __init__(self, **params):
...
+
+    def _check_value(self,val,obj=None):
+        super(DataFrame, self)._check_value(val, obj)
+
+        if isinstance(self.columns, set) and self.ordered is True:
+            raise Exception('Columns cannot be ordered when specified as a set')
```
jbednar: ValueError

param/__init__.py
jlstevens: Addressed in 3dca9a1

param/__init__.py
jlstevens: Addressed in 97c2169

param/__init__.py
jlstevens: Addressed in 70642bc

philippjfr:
Looks good.

PR #289, opened 2018-10-03
Fixed bug when watching class parameter

philippjfr:
In the previous PR which fixed batching I assumed that the ``obj`` would always be provided but if you look at ``Parameterized.__set__`` it is not provided if ``self._owner`` is the object.


philippjfr:
Actually this isn't sufficient, the relevant code here is:

The relevant code here is:

```
            if owning_class != mcs:
                type.__setattr__(mcs,attribute_name,copy.copy(parameter))
            mcs.__dict__[attribute_name].__set__(None,value)
```

The ``_owner`` on the parameter needs to be updated after the copy for this to work.

philippjfr:
Still not sufficient, have to think about this more.

philippjfr:
Okay, the new approach works.

philippjfr:
Had to update the test so that it would actually catch the problem I'm fixing. Ready to merge once passing.

PR #291, opened 2018-10-05
Introduced types field for watch Event objects

jlstevens:
This PR lets `Event` objects declare their type: either 'set', 'changed' or 'triggered'. If 'set', then the event was due to anything where `onlychanged=False`, if 'changed' then `onlychanged=True` and 'triggered' means the `trigger` methods was used.

- [x] Add unit tests


jlstevens:
@philippjfr If you are happy with it (and does what we need), this PR can be merged.

philippjfr:
Looks good to me and worked against my holoviews PR on the first try.

PR #299, opened 2018-11-05
Add output decorator and outputs lookup method

philippjfr:
Adds an ``output`` decorator which allows declaring output methods on a Parameterized object, which declare their output type as a parameter.

The simplest declaration simply declares the method returns an object of without any Parameter type guarantees, e.g.:

```
@output()
```

If a specific parameter type is specified this is a declaration that the method will return a value of that type, e.g.:

```
@output(param.Number())
```

To override the default name of the output the type may be declared as a keyword argument, e.g.:

```
@output(custom_name=param.Number())
```

``output`` also accepts Python object types which will be upgraded to a ClassSelector, e.g.:

```
@output(int)
```

Finally it adds a ``outputs`` method on the ``Parameterized.param`` accessor, which returns the outputs as a dictionary indexed by name and with tuples containing the declared parameter type and the decorated method.


jbednar:
> If a specific parameter type is specified ... `@output(param.Number())`

Here "type" is presumably not used in the Python sense of "class", because you are providing an object (instance).  Are instances required, encouraged, or just possible?  `param.Number` seems more natural unless I want to provide bounds or other data inside the parentheses.

param/parameterized.py
```diff
@@ -211,14 +211,21 @@ def _f(*args, **kwargs):
 
 @accept_arguments
 def depends(func, *dependencies, **kw):
+    """
+    depends allows annotating a method on a Parameterized class to
+    express dependencies on Parameter or Parameter metadata on the
+    same class or one of its Parameters if it is also Parameterized.
```
jbednar: Hmm; confusing.  Maybe:

```
    Annotates a Parameterized method to express its dependencies.
    The specified dependencies can be either be Parameters of this
    class, or Parameters of subobjects (Parameterized objects that
    are values of this object's parameters).  Dependencies can either
    be on Parameter values, or on other metadata about the Parameter.
```

philippjfr:
>Are instances required, encouraged, or just possible?

Instances are not required but internally it will always make an instance for consistency and so that we can eventually enable output type validation if the user requests it.

param/parameterized.py
jbednar: BTW, is there a way to do multi-line replacements in GitHub's new Suggestions feature?  So far I can only see how to edit a single line.

param/parameterized.py
```diff
@@ -231,6 +238,70 @@ def _depends(*args,**kw):
...
+    """
+    output allows annotating a method on a Parameterized class to
+    declare that it returns an output of a specific type. The outputs
+    of a Parameterized class can be queried using the
+    Parameterized.param.outputs method. By default the output will
+    inherit the method name but a custom name can be declared by
+    expressing the Parameter type using a keyword argument.
+
+    The simplest declaration simply declares the method returns an
+    object of without any Parameter type guarantees, e.g.:
```
```suggestion
    object without any type guarantees, e.g.:
```

param/parameterized.py
```diff
@@ -231,6 +238,70 @@ def _depends(*args,**kw):
...
+    To override the default name of the output the type may be declared
+    as a keyword argument, e.g.:
+
+      @output(custom_name=param.Number())
+
+    output also accepts Python object types which will be upgraded to
+    a ClassSelector, e.g.:
+
+      @output(int)
+    """
```
```suggestion
    The object returned by this method will be validated against the specified type, raising an exception for values not accepted by that type.
    """
```

jbednar: Ah, I see above that you don't plan to do this validation soon?

philippjfr: No, @jlstevens convinced me that it isn't immediately necessary and adding a validate method to all Parameters is a fair bit of work.

param/parameterized.py
philippjfr: Guess not.

param/parameterized.py
jbednar: Ok.

PR #306, opened 2018-12-06
Add support for instance parameters

philippjfr:
This PR implements the notion of an instance parameter which can be accessed using a new API. All parameters can in theory become instance parameters (as long as the new ``per_instance`` slot is set to True). The main balance to strike here is that a) we do not want to pay the cost of copying parameters for all instances and b) we do not want users to have to add any special boilerplate to get instance parameters.

The approach I've therefore taken is to only make a copy of a parameter when it is actually needed. This means that all existing param code will never pay the cost of making instance parameters, there's only two situations where the per instance copy is made:

1. When a instance parameter attribute is watched. This is because you do not want events triggered by changes on the class parameter to trigger events on the instance parameter and vice versa.
2. The second situation when a copy is made is when one of the new APIs is used to access the parameter on an instance, e.g. when you access ``instance.param.param_name`` or ``instance.param['param_name']``. This API will make it easy to change the attributes of instance parameters, but the old ``obj.params()``/``obj.param.params()`` method is unaffected by this.

Leaving the old API unaffected will ensure that old param code never pays the costs associated with instance parameters. 

- [x] Add new ``parameterized.param.__getitem__`` and ``parameterized.param.__getattr__`` accessors for parameters
- [x] Modify setters to ensure they use instance parameters for validation
- [x] Ensure that watching instance parameter attributes makes copies
- [x] Add tests

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -153,7 +153,7 @@ def param_union(*parameterizeds, **kwargs):
                 kwargs.popitem()[0]))
     d = dict()
     for o in parameterizeds:
-        for k, p in o.param.params().items():
+        for k in o.param:
             if k != 'name':
                 if k in d and warn:
                     warnings.warn("overwriting parameter {}".format(k))
@@ -366,7 +366,7 @@ class Time(Parameterized):
         if time_type and val is None:
             raise Exception("Please specify a value for the new time_type.")
         if time_type:
-            type_param = self.param.params('time_type')
+            type_param = self.param.objects('current').get('time_type')
             type_param.constant = False
             self.time_type = time_type
             type_param.constant = True
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -957,9 +957,6 @@ class Composite(Parameter):
     attributes.
     """
 
-    # Note: objtype is same as _owner, but objtype left for backwards
-    # compatibility (I think it's used in places to detect composite
-    # parameter)
     __slots__=['attribs','objtype']
 
     def __init__(self,attribs=None,**kw):
@@ -1532,12 +1529,12 @@ class Path(Parameter):
     def _validate(self, val):
         if val is None:
             if not self.allow_None:
-                Parameterized(name="%s.%s"%(self._owner.name,self._attrib_name)).warning('None is not allowed')
+                Parameterized(name="%s.%s"%(self.owner.name,self._attrib_name)).warning('None is not allowed')
         else:
             try:
                 self._resolve(val)
             except IOError as e:
-                Parameterized(name="%s.%s"%(self._owner.name,self._attrib_name)).warning('%s',e.args[0])
+                Parameterized(name="%s.%s"%(self.owner.name,self._attrib_name)).warning('%s',e.args[0])
 
     def __get__(self, obj, objtype):
         """
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -27,9 +27,10 @@ import collections
 from .parameterized import Parameterized, Parameter, String, \
      descendents, ParameterizedFunction, ParamOverrides
 
-from .parameterized import depends, output   # noqa: api import
-from .parameterized import logging_level     # noqa: api import
-from .parameterized import shared_parameters # noqa: api import
+from .parameterized import depends, output     # noqa: api import
+from .parameterized import logging_level       # noqa: api import
+from .parameterized import shared_parameters   # noqa: api import
+from .parameterized import instance_descriptor # noqa: api import
 
 from collections import OrderedDict
 
@@ -488,7 +489,7 @@ class Dynamic(Parameter):
         else:
             return self._produce_value(gen)
 
-
+    @instance_descriptor
     def __set__(self,obj,val):
         """
         Call the superclass's set and keep this parameter's
@@ -873,7 +874,6 @@ class Tuple(Parameter):
         if not len(val)==self.length:
             raise ValueError("%s: tuple is not of the correct length (%d instead of %d)." %
                              (self._attrib_name,len(val),self.length))
-
 
 
 
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -26,11 +26,8 @@ import collections
 
 from .parameterized import Parameterized, Parameter, String, \
      descendents, ParameterizedFunction, ParamOverrides
-
-from .parameterized import depends, output     # noqa: api import
-from .parameterized import logging_level       # noqa: api import
-from .parameterized import shared_parameters   # noqa: api import
-from .parameterized import instance_descriptor # noqa: api import
+from .parameterized import (depends, output, logging_level, # noqa: api import
+     shared_parameters, instance_descriptor)
 
 from collections import OrderedDict
 
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -489,6 +489,7 @@ class Dynamic(Parameter):
         else:
             return self._produce_value(gen)
 
+
     @instance_descriptor
     def __set__(self,obj,val):
         """
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -726,18 +726,18 @@ class Number(Dynamic):
             if vmax is not None:
                 if incmax is True:
                     if not val <= vmax:
-                        raise ValueError("Parameter '%s' must be at most %s"%(self._attrib_name,vmax))
+                        raise ValueError("Parameter '%s' must be at most %s"%(self.name,vmax))
                 else:
                     if not val < vmax:
-                        raise ValueError("Parameter '%s' must be less than %s"%(self._attrib_name,vmax))
+                        raise ValueError("Parameter '%s' must be less than %s"%(self.name,vmax))
 
             if vmin is not None:
                 if incmin is True:
                     if not val >= vmin:
-                        raise ValueError("Parameter '%s' must be at least %s"%(self._attrib_name,vmin))
+                        raise ValueError("Parameter '%s' must be at least %s"%(self.name,vmin))
                 else:
                     if not val > vmin:
-                        raise ValueError("Parameter '%s' must be greater than %s"%(self._attrib_name,vmin))
+                        raise ValueError("Parameter '%s' must be greater than %s"%(self.name,vmin))
 
 
 
@@ -753,7 +753,7 @@ class Number(Dynamic):
             return
 
         if not _is_number(val):
-            raise ValueError("Parameter '%s' only takes numeric values"%(self._attrib_name))
+            raise ValueError("Parameter '%s' only takes numeric values"%(self.name))
 
         self._checkBounds(val)
 
@@ -797,7 +797,7 @@ class Integer(Number):
             return
 
         if not isinstance(val,int):
-            raise ValueError("Parameter '%s' must be an integer."%self._attrib_name)
+            raise ValueError("Parameter '%s' must be an integer."%self.name)
 
         self._checkBounds(val)
 
@@ -825,16 +825,16 @@ class Boolean(Parameter):
         if self.allow_None:
             if not isinstance(val,bool) and val is not None:
                 raise ValueError("Boolean '%s' only takes a Boolean value or None."
-                                 %self._attrib_name)
+                                 %self.name)
 
             if val is not True and val is not False and val is not None:
-                raise ValueError("Boolean '%s' must be True, False, or None."%self._attrib_name)
+                raise ValueError("Boolean '%s' must be True, False, or None."%self.name)
         else:
             if not isinstance(val,bool):
-                raise ValueError("Boolean '%s' only takes a Boolean value."%self._attrib_name)
+                raise ValueError("Boolean '%s' only takes a Boolean value."%self.name)
 
             if val is not True and val is not False:
-                raise ValueError("Boolean '%s' must be True or False."%self._attrib_name)
+                raise ValueError("Boolean '%s' must be True or False."%self.name)
         super(Boolean, self)._validate(val)
 
 
@@ -856,7 +856,7 @@ class Tuple(Parameter):
             self.length = len(default)
         elif length is None and default is None:
             raise ValueError("%s: length must be specified if no default is supplied." %
-                             (self._attrib_name))
+                             (self.name))
         else:
             self.length = length
         self._validate(default)
@@ -867,11 +867,11 @@ class Tuple(Parameter):
             return
 
         if not isinstance(val,tuple):
-            raise ValueError("Tuple '%s' only takes a tuple value."%self._attrib_name)
+            raise ValueError("Tuple '%s' only takes a tuple value."%self.name)
 
         if not len(val)==self.length:
             raise ValueError("%s: tuple is not of the correct length (%d instead of %d)." %
-                             (self._attrib_name,len(val),self.length))
+                             (self.name,len(val),self.length))
 
 
 
@@ -884,7 +884,7 @@ class NumericTuple(Tuple):
             for n in val:
                 if not _is_number(n):
                     raise ValueError("%s: tuple element is not numeric: %s." %
-                                     (self._attrib_name,str(n)))
+                                     (self.name,str(n)))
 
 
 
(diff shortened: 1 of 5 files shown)


philippjfr:
Another thing to decide, a parameter already has an ``_owner`` slot which is usually the class but is now replaced when the instance when the parameter is copied. I feel like the owning instance will need to be public for a lot of the functionality we want to build so should we just remove the underscore or rename it completely?

jbednar:
> The approach I've therefore taken is to only make a copy of a parameter when it is actually needed.

That seems like a good approach, but it seems to me that user code could use params() on an instance to get a Parameter object from the class, then one of the cases you outlined above would result in a per-instance copy to be made of that Parameter, and then the saved Parameter object would now be stale.  I guess the expectation is that such passing around of Parameter objects outside the class is not to be encouraged or supported anyway?

I guess making the owner be public would let you detect if this has happened...

philippjfr:
Yes, that is a worry I share, I would indeed not encourage users to store a cache of parameters outside the instance and cannot envision any time I would have done this but then we can't prevent users from doing it either. This is also related to questions about the future API, do we expect to keep ``obj.param.params()``? If we intend to say users should always access their parameters via the getitem/getattr on the Parameters object then we should probably also have a way to disable instance parameters at the class level instead of forcing users to set ``per_instance`` in each parameter constructor.

jbednar:
That all sounds reasonable to me.

param/parameterized.py
```diff
@@ -913,6 +926,44 @@ def self_or_cls(self_):
...
+        parameters = self_.params() if inst is None else inst.param.params()
+        p = parameters[key]
+        if inst is not None and p.per_instance and key not in inst._instance__params:
+            p = copy.copy(p)
+            p._owner = inst
+            inst._instance__params[key] = p
+        return p
+
+
+    def __dir__(self_):
```
jlstevens: Last I checked IPython isn't respecting ``__dir__` properly if you want this for tab completion purposes...

philippjfr: Works for me.

param/parameterized.py
```diff
@@ -913,6 +926,44 @@ def self_or_cls(self_):
...
+        cls = self_.__dict__.get('cls')
+        if cls is None: # Class not initialized
+            raise AttributeError
+
+        try:
+            params = list(getattr(cls, '_%s__params' % cls.__name__))
+        except AttributeError:
+            params = [n for class_ in classlist(cls) for n, v in class_.__dict__.items()
+                      if isinstance(v, Parameter)]
+
```
jlstevens: Looks to me like it should be the other way around - check the methods first, then fall back to the parameters. I wouldn't want a method to be shadowed by a parameter of the same name. What would be best would be to check if a parameter has a matching name to an existing method and warn when returning the method, explaining that the requested parameter has been shadowed (and maybe suggest to use `__getitem__` directly instead.

philippjfr: ``__getattr__`` only gets called if the attribute wasn't otherwise found.

param/parameterized.py
jlstevens: Probably works for some IPython versions and not others.

param/__init__.py
```diff
@@ -489,6 +490,7 @@ def __get__(self,obj,objtype):
             return self._produce_value(gen)
 
 
+    @instance_descriptor
```
jlstevens: Looks like this `instance_descriptor` decorator is used precisely once. Where else would you expect it to be used?

philippjfr: It should actually also be used on ``Parameter.__set__``, think that's an oversight.

jlstevens:
> I guess the expectation is that such passing around of Parameter objects outside the class is not to be encouraged or supported anyway?

Does that means this isn't compatible with https://github.com/pyviz/param/pull/317? I think there are valid uses of Parameters not bound to parameterized classes (and in that PR I use `copy.copy` to copy parameter instances).

param/parameterized.py
```diff
@@ -553,7 +564,8 @@ class Foo(Bar):
 
     def __init__(self,default=None,doc=None,precedence=None,  # pylint: disable-msg=R0913
                  instantiate=False,constant=False,readonly=False,
-                 pickle_default_value=True, allow_None=False):
+                 pickle_default_value=True, allow_None=False,
+                 per_instance=True):
```
jlstevens: The `per_instance` flag should be described in the docstring below. What I'm wondering is when a user should set this flag. I can imagine certain parameter *types* shouldn't allow `per_instance` but in that case, why expose this through the constructor? If a user disabled `per_instance` when setting a parameter, couldn't this have confusing repercussions if the watch mechanism is used for example?

param/parameterized.py
```diff
@@ -1089,6 +1141,9 @@ def params(self_, parameter_name=None):
             setattr(cls,'_%s__params'%cls.__name__,paramdict)
             pdict= paramdict
 
+        if self_.self is not None:
+            pdict.update(self_.self._instance__params)
+
```
jlstevens: I'm not entirely sure, but it seems that if copies are made (to make a parameter into an instance parameter) in ``__getitem__``,  then the parameter must always exist as a normal parameter first and then as an instance parameter. Won't both those parameters have the same name and therefore won't you have a key clashing in this dictionary?

jlstevens: Unless the semantics is updated so `params` always prioritizes returns instance parameters. I'm not sure that would always be a good policy so maybe there should be a new flag to control this?

philippjfr: Yes, a flag makes definite sense.

jlstevens:
Anyway, having stared at this a bit longer, I think the overall approach is sound. The tricky bit is figuring out the possible consequences of this API but I think the way the instance parameters are created and stored is reasonable.

param/parameterized.py
jlstevens: Ok. In that case you can't get a warning if there is a name clash.

jbednar:
> Does that means this isn't compatible with #317

I don't think so; #317 is about building up Parameterized from Parameters; the restriction here is the other way around (i.e. that pulling a Parameter out from a Parameterized might have surprising behavior later).

jlstevens:
@jbednar  That's what I thought. I just wanted to double check!

param/__init__.py
philippjfr: It does seem to be used there, must have missed it in the diff.

param/parameterized.py
philippjfr: You might genuinely want to share parameter attributes across multiple instances of a class.

philippjfr:
Okay I just wanted to summarize the outcome of our discussions today. The main starting point of our discussion was that the ``.params`` method was highly problematic in this new scheme because we had three bad options 1) it would always return class parameters for backward compatibility which would be confusing when instance parameters exist 2) it would always return instance parameters which would mean all kinds of old code would end up paying the unnecessary costs or 3) it would return the instance parameters at a particular point in time which would be confusing if a user cached them. The other reason we wanted to get rid of .params is because it was weirdly overloaded either returning a whole dict of params or a single parameter and because ``.param.params`` looks very redundant.

Therefore we decided we should replace with new API that is less ambiguous. In addition to the newly added ``__getitem__`` and ``__getattr__`` methods we therefore decided to add the following API:

#### Make foo.param an iterator over parameter names

```python
list(parameterized.param)
```

This replaces code like this:

```python
list(parameterized.param.params().keys())
```

#### Add a ``__contains__`` implementation

```python
param_name in parameterized.param
```

Replacing code like:

```python
param_name in parameterized.param.params()
```

#### New ``.param.objects``

Naming suggestions welcome, this method is roughly equivalent to .params but makes the choice between instance and class parameters explicit, the default is to return instance parameters and this will force copies of those parameters to be made, to force the return of class parameters you can do this:

```python
foo.param.objects(instance=False)
```

One thing we hadn't discussed is that internal param (and external library) code wants to access the parameters as well and wants to avoid creating instance parameters if it's unnecessary. Therefore I've currently also allowed ``.param.objects(instance='current')`` which returns the parameter objects that are currently actually in use. Please suggest a better name for this option if you can think of one.

#### Warnings if using ``.params``

Eventually we want to deprecate ``.params``, however for the time being it should behave exactly as it did. To make sure this is safe and users don't accidentally use it to get a handle on class parameter instances when there are instance parameters we issue this warning:

```python
The Parameterized instance has instance parameters created using
new-style param APIs, which are incompatible with .params.  Either use
the new APIs on the .param accessor, i.e. either .param.objects to
query all class or instance parameters or use .param[name] to access a
specific parameter object by name.
```

#### Disable instance parameters globally

We wanted to make the per_instance behavior to be toggleable globally, what we came up with was finding a way to set the default for ``Parameter.per_instance`` to ``False``. I may be wavering on this as we'd have to use some tricks to get this to work with Parameter slots. The other suggestion was to add a special flag just like ``__abstract`` on the class, neither one of us was sure about this though.

philippjfr:
Also I apologize for some of the diffs, I can't work on param without getting annoyed by the inconsistencies in spacing, outdated comments etc. so I ended up updating things that I touched.

jbednar:
That all sounds like a good plan. I agree that the .params method should be deprecated and then removed in 2.0, but it seems like we can do so in a compatible way that doesn't currently raise any warnings for old code. Basically, I don't think any old code will cause instance parameters to be created, so can we make .params() return class parameters only, but warn loudly if there are any instance parameters defined?  That way old code keeps working fine (and quietly) with param 1.x, but when we move to 2.0 we remove .params entirely.

philippjfr:
>Basically, I don't think any old code will cause instance parameters to be created, so can we make .params() return class parameters only, but warn loudly if there are any instance parameters defined?

That's exactly what we decided and what's implemented here.

param/__init__.py
```diff
@@ -27,9 +27,10 @@
 from .parameterized import Parameterized, Parameter, String, \
      descendents, ParameterizedFunction, ParamOverrides
 
-from .parameterized import depends, output   # noqa: api import
-from .parameterized import logging_level     # noqa: api import
-from .parameterized import shared_parameters # noqa: api import
+from .parameterized import depends, output     # noqa: api import
+from .parameterized import logging_level       # noqa: api import
+from .parameterized import shared_parameters   # noqa: api import
+from .parameterized import instance_descriptor # noqa: api import
```
jbednar: Seems like a strange mix of formatting styles; I'd favor matching the style on on line 27 to avoid all the redundancy.

param/__init__.py
```diff
@@ -152,7 +153,7 @@ def param_union(*parameterizeds, **kwargs):
                 kwargs.popitem()[0]))
     d = dict()
     for o in parameterizeds:
-        for k, p in o.param.params().items():
+        for k in o.param:
```
jbednar: That works out to a nice simplification.

jbednar:
> That's exactly what we decided and what's implemented here.

Ok, great. I misread your description as warning if there are instance parameters allowed, but it sounds like it's only warning if there are actually any created on this object, which is what I was after.  I'm still sorting through the diffs...

param/parameterized.py
```diff
@@ -569,6 +581,11 @@ def __init__(self,default=None,doc=None,precedence=None,  # pylint: disable-msg=
         inheritance of Parameter slots (attributes) from the owning-class'
         class hierarchy (see ParameterizedMetaclass).
 
+        per_instance defaults to False and controls whether a new
+        Parameter instance is created for every Parameterized
+        instance. Enabling this will share the same parameter for all
+        instances including all validation attributes.
```
jbednar: This docstring seems all mixed up!  Shouldn't it be:

```
per_instance defaults to True and controls whether a new
Parameter instance can be created for every Parameterized
instance. If False, all instances of a Parameterized 
class will share the same parameter object, including
all validation attributes.
```

param/parameterized.py
```diff
@@ -1070,24 +1159,15 @@ def params(self_, parameter_name=None):
...
-            setattr(cls,'_%s__params'%cls.__name__,paramdict)
-            pdict= paramdict
+        pdict = self_.objects(instance=False)
+        if self_.self is not None and self_.self._instance__params:
+            self_.warning('The Parameterized instance has instance '
+                          'parameters created using new-style param '
+                          'APIs, which are incompatible with .params. '
+                          'Either use the new APIs on the .param accessor, '
+                          'i.e. either .param.objects to query all class '
+                          'or instance parameters or use .param[name] to '
```
```suggestion
                          'or instance parameters, or use .param[name] to '
```

philippjfr:
I ended up renaming the ``.options(instance='current')`` to ``.options(instance='existing')``. I also added a class decorator called ``disable_instance_params`` (better/shorter name welcome). This PR is now ready to merge.

param/parameterized.py
```diff
@@ -209,6 +209,25 @@ def _f(*args, **kwargs):
     return _f
 
 
+def disable_instance_params(cls):
+    """
+    Disables instance parameters on the class
+    """
+    cls._disable_instance__params = True
```
jbednar: What's with the __?

param/parameterized.py
```diff
@@ -209,6 +209,25 @@ def _f(*args, **kwargs):
     return _f
 
 
+def disable_instance_params(cls):
```
jbednar: Maybe `no_instance_params`?  Disabling sounds like there could be instance parameter objects they just would be non-functional, where `no` seems more like a declaration that there are not and should not be any instance params.

param/parameterized.py
philippjfr: Was following the weird pattern we use to cache parameters on classes, i.e.:

```
'_Parameterized__params'
```

philippjfr:
Oh, one more thing, we added ``_owner`` to Parameters which now holds the owning instance on instance parameters but really we want a public API for this. Should I just add a ``owner`` property to make ``_owner`` public?

jbednar:
> Should I just add a `owner` property to make `_owner` public?

Seems fine, though it seems simpler just to rename `_owner` to `owner`.

philippjfr:
True, it's private API so it should be safe.

philippjfr:
For convenience we should also make the `_attrib_name` slot public since that information is important to be able to watch the parameter.

philippjfr:
Any objection to simply calling that ``name``, which will be complemented by a ``label``?

jbednar:
Sounds good.

philippjfr:
This is ready to review and merge from my perspective. I'd like to merge and cut a dev release today.

philippjfr:
I've also explicitly disallowed reusing Parameter instances on multiple classes, this would already have caused various issues but wasn't explicitly disallowed.

param/parameterized.py
```diff
@@ -724,11 +748,20 @@ def _post_setter(self, obj, val):
...
-        self._attrib_name = attrib_name
+        raise TypeError("Cannot delete '%s': Parameters deletion not allowed." % self.name)
+
+
+    def _set_names(self, attrib_name):
+        if None not in (self.owner, self.name) and attrib_name != self.name:
+            raise AttributeError('The %s parameter %r has already been '
+                                 'assigned a name by the %s class, '
+                                 'could not assign new name %r. Parameters '
+                                 'may not be shared by multiple classes, '
```
```suggestion
                                 'may not be shared by multiple classes; '
```

param/parameterized.py
```diff
@@ -536,24 +555,21 @@ class Foo(Bar):
     # attributes.  Using __slots__ requires special support for
     # operations to copy and restore Parameters (e.g. for Python
     # persistent storage pickling); see __getstate__ and __setstate__.
-    __slots__ = ['_attrib_name','_internal_name','default','doc',
+    __slots__ = ['name','_internal_name','default','doc',
```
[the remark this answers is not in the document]
philippjfr: Yes, I think it probably would. I can test shortly.

jlstevens:
How about `disable_instance_params` to `without_instance_params` (or `no_instance_params`)? I don't think you are disabling instance parameters as such: instead you are stopping them from ever being created.

philippjfr:
>no_instance_params

It's already been renamed to that.

jbednar:
> I don't think you are disabling instance parameters as such: instead you are stopping them from ever being created.

That's precisely why I suggested `no_instance_params`; good to see we are on the same wavelength.

jlstevens:
Going to be very happy with the functionality offered by this PR. There is a potential pickling issue that may need to be addressed but once that is done and the tests pass, I am happy to see this merged.

Issue #267, opened 2018-09-19
Delete 'root' from Param logging messages?


jbednar:
From @sdrobert: 

> Parameterized(name="param_union").warning("overwriting parameter %s",k)

Is there any way to avoid the 'root' part of the message?

WARNING:root:param_union: overwriting parameter a
WARNING:root:param_union: overwriting parameter b

From @ceball:

I don't know what purpose 'root' serves in general.

param's logging system uses Python's logging module (see https://github.com/ioam/param/blob/master/param/parameterized.py#L32); I think everyone would be happy to see it simplified and/or clarified, but nobody's had time to work on it. This link indicates the name's from the logging module's "root" logger; setting param.logger to some other named logger would probably change "root" to that name (because param would be using a different logger).

From @jbednar:

Maybe we should just declare that the root logger has the name "param", to give people an idea where these messages come from?

sdrobert:
My two cents (if desired):

From the Python docs:

```
The name is potentially a period-separated hierarchical value, like foo.bar.baz (though it could also be
just plain foo, for example). Loggers that are further down in the hierarchical list are children of loggers
higher up in the list. For example, given a logger with a name of foo, loggers with names of foo.bar,
foo.bar.baz, and foo.bam are all descendants of foo. The logger name hierarchy is analogous to the
Python package hierarchy, and identical to it if you organise your loggers on a per-module basis using
the recommended construction logging.getLogger(__name__). That’s because in a module, __name__ is
the module’s name in the Python package namespace.
```

Staying faithful to the `logging` system seems most sensible to me. For functions like `param_union`, you'd use the logger ``"param"``, corresponding to the package's logger. Then if you want to submit a warning through a parameterized instance, you'd ask for the logger ``"param".<name of parameterized instance>``, which would inherit from the ``"param"`` logger and any handlers/formatters. Here's an example:
``` python
import logging
param_logger = logging.getLogger('param')
formatter = logging.Formatter(fmt='%(levelname)s:%(name)s: %(message)s')
handler = logging.StreamHandler()
handler.setFormatter(formatter)
param_logger.addHandler(handler)
param_logger.warning('foo')  # prints "WARNING:param: foo"
param_bar_logger = logging.getLogger('param.bar')
param_bar_logger.warning('foo')  # prints "WARNING:param.bar: foo"
```

[...]

PR #308, opened 2018-12-09
Updates to logging system

sdrobert:
This PR is (finally) in response to discussions in #267. I've made the changes outlined there. More specifically:

1. The root logger is no longer touched. Instead, a logger named "params" is spawned. Then, every parameterized instance, by default, uses the logger "params.\<name\>", where "\<name\>" is the name of the instance.
2. ``_db_print`` does minimal formatting now. Instead it's handled by the ``StreamHandler`` attached to the "params" logger.
3. ``param_union`` no longer uses ``warnings``.

Let me know if anything needs clarification or fixing.

Best,
Sean

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -21,11 +21,10 @@ import sys
 import glob
 import re
 import datetime as dt
-import warnings
 import collections
 
 from .parameterized import Parameterized, Parameter, String, \
-     descendents, ParameterizedFunction, ParamOverrides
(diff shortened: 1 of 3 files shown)


jbednar:
Cool! It should probably be "param", not "params"...

sdrobert:
Whoopsies! Sorry and thanks. Fixed.

philippjfr:
This was merged as part of #330. Thanks for your contribution @sdrobert, and sorry it took so long.

PR #312, opened 2019-01-27
Added support for multiple outputs

philippjfr:
This PR adds support for declaring multiple outputs as the return type of a method. The expectation is that if there are multiple outputs they are expected to be in a tuple. To declare multiple outputs two main formats are supported, either as keywords (only in py>=3.6) or for compatibility with older python versions it also supports tuples:

      @output(number=param.Number(), string=param.String())
      def some_method(self):
          ...

      @output(('number', param.Number()), ('string', param.String()))
      def some_method(self):
          ...

The ``parameterized.param.outputs()`` method returns a dictionary like this:

```python
{'number': (param.Number(), parameterized.some_method, 0),
 'string': (param.String(), parameterized.some_method, 1)} 
```

i.e. a mapping from the output's name to a tuple of the (param_type, method,  index), where the index is the index of the output in the return value of the method.

Additionally this PR adds an ``evaluate`` argument to ``Parameterized.param.outputs`` which evaluates the outputs for you and returns a dictionary of the values. Now that a method might return multiple outputs this avoids the user manually having to check the index of a returned value and unpacking it appropriately.

- [x] Implements https://github.com/pyviz/param/issues/310
- [x] Adds unit tests

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -249,7 +249,7 @@ def output(func, *output, **kw):
     Parameterized.param.outputs method. By default the output will
     inherit the method name but a custom name can be declared by
     expressing the Parameter type using a keyword argument. Declaring
-    multiple return types is only supported in Python >= 3.6.
+    multiple return types using keywords is only supported in Python >= 3.6.
 
     The simplest declaration simply declares the method returns an
     object without any type guarantees, e.g.:
@@ -296,6 +296,7 @@ def output(func, *output, **kw):
         if (py_major < 3 or (py_major == 3 and py_minor < 6)) and len(kw) > 1:
             raise ValueError('Multiple output declaration using keywords '
                              'only supported in Python >= 3.6.')
+          # (requires keywords to be kept ordered, which was not true in previous versions)
         outputs = [(name, otype, i if len(kw) > 1 else None)
                    for i, (name, otype) in enumerate(kw.items())]
     else:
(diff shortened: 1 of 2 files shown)


param/parameterized.py
```diff
@@ -1313,21 +1343,33 @@ def params_depended_on(self_,name):
         return _params_depended_on(MInfo(cls=self_.cls,inst=self_.self,name=name,method=getattr(self_.self_or_cls,name)))
 
 
-    def outputs(self_):
+    def outputs(self_, evaluate=False):
         """
         Returns a mapping between any declared outputs and a tuple
-        of the declared Parameter type and the output method.
+        of the declared Parameter type, the output method and the
```
```suggestion
        of the declared Parameter type, the output method, and the
```

philippjfr:
>Can you add something to the docs about how to use this?

Would be happy too, any suggestions on where? Also worth noting that ``param.depends`` and watch are also not documented.

jlstevens:
The API looks good to me and I hope we can forget about the python 2 version when all our projects drop support for it.

My only other thought is that maybe adding an `evaluate` argument overloads that method a bit too much so maybe there should be another method instead? With the namespace I think we should avoid overcomplicating existing methods and should feel free to define new methods when it makes sense.

Reviewing the code now but the functionality looks good!

jlstevens:
Code looks good though I am still a little unsure about the API around the `evaluate` argument. My understanding is that panel doesn't actually require it and my instinct is that this should be a new method that we could add later and don't need right away.

Once you've decided what you want to do about `evaluate` (possibly remove it for now?) I am happy to merge.

philippjfr:
I think adding `evaluate` was indeed premature, I was initially using it in panel but found I needed something slightly more complex. We can come up with a fully baked public API for that later. I'll strip it out now.

jlstevens:
All looks good and the tests are green. Merging.

PR #317, opened 2019-02-19
Added keywords_to_params promotion utility

jlstevens:
This PR adds a utility that has cropped up several times and has resulted in some level of code duplication. This has cropped up on some dashboarding projects and is used in the HoloViews [`Streams.define`](https://github.com/pyviz/holoviews/blob/master/holoviews/streams.py#L85) classmethod. It is also relevant to the sort of promotion used to generate widgets in the style of `interact` (e.g panel's `interact` function).

Currently WIP as there are several things to discuss:

* Tuples: These map to `param.Tuple` but if the elements are numeric types this could be mapped to `param.NumericTuple`. Note that numeric 2-tuples can also define the bounds on numeric types but you would need a default value: i.e I don't think you can map numeric 2-tuples to a number parameter including bounds. There is also `param.Range` which is related (maybe for 2-tuples only?)
* Dict and List: These map to `param.Dict` and `param.List` but maybe it is more useful to map these to `Selector` instead. How can you switch between these behaviors? Is a flag appropriate?
* HookList: Could check if a list only contains callable elements?
* Path and Filename: Currently only strings are supported though I note that Python 3 has [explicit path objects](https://docs.python.org/3/library/pathlib.html) which would make this unambiguous.
* Callable, Action: If the object is callable, could maybe make this a final fallback before resorting to `param.Parameter`? How to pick between `Callable` or `Action`?
* ClassSelector: Also a fallback if the value is a `type`?
* Foldername, ListSelector, MultiFileSelector, Magnitude, XYCoordinates: Not thought about how to handle any of these yet. 
* Color: Could try to detect hex strings of the right length but this might be risky?

TODO (i.e shouldn't require much discussion):

* Support the new pandas types `Series` and `DataFrame`.
* Support `Date` from datetimes.
* Support `DateRange` from a 2-tuple of datetimes.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -158,26 +158,26 @@ def keywords_to_params(selector, **kwargs):
     params = {}
     for k, v in kwargs.items():
         kws = dict(default=v, constant=True)
-        if isinstance(v, param.Parameter):
+        if isinstance(v, Parameter):
             params[k] = v
         elif isinstance(v, bool):
-            params[k] = param.Boolean(**kws)
+            params[k] = Boolean(**kws)
         elif isinstance(v, int):
-            params[k] = param.Integer(**kws)
+            params[k] = Integer(**kws)
         elif isinstance(v, float):
-            params[k] = param.Number(**kws)
+            params[k] = Number(**kws)
         elif isinstance(v, str):
-            params[k] = param.String(**kws)
+            params[k] = String(**kws)
         elif isinstance(v, dict):
-            params[k] = param.Dict(**kws)
+            params[k] = Dict(**kws)
         elif isinstance(v, tuple):
-            params[k] = param.Tuple(**kws)
+            params[k] = Tuple(**kws)
         elif isinstance(v, list):
-            params[k] = param.List(**kws)
+            params[k] = List(**kws)
         elif isinstance(v, np.ndarray):
-            params[k] = param.Array(**kws)
+            params[k] = Array(**kws)
         else:
-            params[k] = param.Parameter(**kws)
+            params[k] = Parameter(**kws)
 
     return params
 
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -182,6 +182,19 @@ def keywords_to_params(**kwargs):
     return params
 
 
+def create_parameterized_class(name, params, bases=None):
+    """
+    Dynamically create a parameterized class with the given name and the
+    supplied parameters using the specified bases (if specified).
+    """
+    if bases is None:
+        bases = (Parameterized,)
+    if isinstance(bases, list):
+        bases = tuple(bases)
+    elif not isinstance(bases, tuple):
+        bases = (bases,)
+    return type(name, bases, params)
+
 def _get_min_max_value(min, max, value=None, step=None):
     """Return min, max, value given input values with possible None."""
     # Either min and max need to be given, or value needs to be given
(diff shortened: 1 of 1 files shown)


jlstevens:
To elaborate further on why this is often used and how:

1. In `hv.Streams.define` you only need to define parameters on a class inheriting from `Stream` (where all the implementation lives). It is very convenient to use `hv.Streams.define('Count', count=4)` instead of writing out a parameterized class inheriting from `Stream`.
2. When using `interact` you have very similar behavior as you want to ask the user for as little information to get some viable widgets to get something immediately working, even if it means guessing things like appropriate bounds. Over the years, mapping parameterized classes to widgets has been a core concept in paramtk, paramnb, parambokeh and now panel.
3. When defining dashboards this code can help with rapid prototyping and experimentation by defining initially defining literals that can be easily changed (e.g `3` instead of `param.Number(3)`) and then supplying a more fully specified parameter object later. It can also be useful to use a dictionary to make the set of defined parameters dynamic.

There is another argument that isn't about these *heuristics* but about having bags of parameter be meaningful and useful without having to attach them to a class. As this utility currently returns a dictionary of parameters (`params`) , you would typically need to create a parameterized class dynamically using `type` e.g with `type(name, (Stream,), params)` or `type(name, (param.Parameterized,), params).`

jbednar:
Thanks. We should discuss what advantages and disadvantages there are to trying to centralize this behavior.

jlstevens:
@jbednar The PR now implements the discussed functionality. Two things that can be improved:

* Function names - looking for better suggestions!
* Any elaboration/tweaks to the docstrings.

Once the names are settled, I'll write the corresponding unit tests (I think it is fine to write the tests later as these functions are dead simple).

param/__init__.py
```diff
@@ -150,6 +151,109 @@ def param_union(*parameterizeds, **kwargs):
     return d
 
 
+def keywords_to_params(**kwargs):
```
jbednar: guess_param_types? (to make it parallel to guess_bounds)?

param/__init__.py
```diff
@@ -150,6 +151,109 @@ def param_union(*parameterizeds, **kwargs):
...
+    Dynamically create a parameterized class with the given name and the
+    supplied parameters using the specified bases (if specified).
+    """
+    if bases is None:
+        bases = (Parameterized,)
+    if isinstance(bases, list):
+        bases = tuple(bases)
+    elif not isinstance(bases, tuple):
+        bases = (bases,)
+    return type(name, bases, params)
```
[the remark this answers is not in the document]
jlstevens: Addressed in afac46d

param/__init__.py
```diff
@@ -150,6 +151,109 @@ def param_union(*parameterizeds, **kwargs):
...
+            params[k] = List(**kws)
+        elif isinstance(v, np.ndarray):
+            params[k] = Array(**kws)
+        else:
+            params[k] = Parameter(**kws)
+
+    return params
+
+
+def create_parameterized_class(name, params, bases=None):
```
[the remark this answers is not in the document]
jlstevens: Addressed in a850fe6

param/__init__.py
jlstevens: Addressed in d8dd921

param/__init__.py
```diff
@@ -150,6 +151,120 @@ def param_union(*parameterizeds, **kwargs):
...
+            else:
+                params[k] = Parameter(**kws)
+
+    return params
+
+
+def parameterized_class(name, params, bases=Parameterized):
+    """
+    Dynamically create a parameterized class with the given name and the
+    supplied parameters using the specified bases (if specified).
```
jbednar:
```suggestion
    supplied parameters, inheriting from the specified base(s).
```

PR #319, opened 2019-03-01
Add label slot to Parameters

philippjfr:


jlstevens:
Looks good. This should be very useful! I'll merge when the tests pass.

jbednar:
Looks good.  Can you put in a property or other mechanism that lets people retrieve the label if it is present, but falling back to the name?  That way display can always simply display the label, but usually it will just be the name (unless someone needed to override it).

philippjfr:
>Can you put in a property or other mechanism that lets people retrieve the label if it is present, but falling back to the name? 

Seems okay for users, but note this is not useful for our code in any way. A name will have formatters applied to make it look pretty, a label should be respected as is. If you still want to add it please suggest a name.

jbednar:
Hmm.  And in practice the formatters are defined at the Panel level?  It might make sense to put all of that machinery into Param, so that label becomes a property that is set to the final result after all this formatting (if not explicitly set), or the explicitly set value (if any).  Seems like providing a displayable name for something is a core feature of Param, to me, as it can appear in many different contexts (GUIs, reports, docs, ...

I'm not *certain* it ought to be in Param, but I do think that should be considered.

philippjfr:
Right, that's all true, my main concern moving this into param is where the name formatter would be defined.

jbednar:
I'm not sure either, and I'm not supposed to be near a computer right now. :-)

philippjfr:
Maybe we just start with a default formatter:

```python
pname = pname.replace('_',' ')
pname = pname[:1].upper() + pname[1:]
```

And think about configurability later? As long as we decide that we want to support formatting in param in principle that seems okay to me.

jbednar:
That seems ok in principle, but given that Panel is already offering configurability and people are already using it, moving that configurability to Param seems urgent if that's where it will eventually live (which is what I think).

param/parameterized.py
```diff
@@ -628,6 +628,17 @@ class hierarchy (see ParameterizedMetaclass).
...
 
 
+    @property
+    def label(self):
+        if self.name and self._label is None:
+            return label_formatter(self.name)
+        else:
+            return self._label
+
+    @label.setterg
```
philippjfr: Typo

jlstevens: Fixed in 7028e4f

param/parameterized.py
```diff
@@ -609,6 +628,17 @@ class hierarchy (see ParameterizedMetaclass).
         self.per_instance = per_instance
 
 
+    @property
+    def label(self):
+        if self.name and self._label is None:
```
philippjfr: If someone were to set name to ``''`` somehow this would still return ``self._label``, i.e. None. I don't think that matters much though since no one should ever be able to set name to ``''``.

jlstevens: Right, I think it is impossible to have an empty attribute name...

param/parameterized.py
```diff
@@ -54,6 +54,24 @@ def get_logger():
...
+            return overrides[pname]
+        if replace_underscores:
+            pname = pname.replace('_',' ')
+        if capitalize:
+            pname = pname[:1].upper() + pname[1:]
+        return pname
+    return label_formatter
+
+
+label_formatter = create_label_formatter()
```
jbednar: Just to check my understanding, it looks like this is creating a closure that stores the state of `replace_underscores` and `overrides`, and that the reason it's done this way as opposed to the usual approach of a ParameterizedFunction is to avoid circularity, in that this label_formatter is used in the definition of Parameter?  If so that's ok.  Not sure if it would be cleaner to define it as a ParameterizedFunction but only set the value on Parameter once Parameterized has been defined; the advantage would be that then one could do `LabelFormatter.replace_underscores=False`, but the disadvantage would be the difficult to track circularity.

jlstevens: > Just to check my understanding, it looks like this is creating a closure that stores the state of replace_underscores and overrides, and that the reason it's done this way as opposed to the usual approach of a ParameterizedFunction is to avoid circularity, in that this label_formatter is used in the definition of Parameter? 

That is exactly it. The original *was* a parameterized function that I tried to preserve but the circularity was a problem (also, it needs the parameters themselves which are in the other file, causing circular imports).

jlstevens: I switched back to `ParameterizedFunction` in efa8343 but only because I am using `param.Parameter` instead of the more appropriate parameter types...

param/parameterized.py
```diff
@@ -2698,6 +2713,32 @@ def pprint(self, imports=None, prefix="\n    ",unknown_value='<?>',
...
+        if pname in self.overrides:
+            return self.overrides[pname]
+        if self.replace_underscores:
+            pname = pname.replace('_',' ')
+        if self.capitalize:
+            pname = pname[:1].upper() + pname[1:]
+        return pname
+
+
+label_formatter = default_label_formatter.instance()
```
jbednar: What is the benefit of `.instance()` here?

param/parameterized.py
```diff
@@ -2698,6 +2713,32 @@ def pprint(self, imports=None, prefix="\n    ",unknown_value='<?>',
 
 
 
+class default_label_formatter(ParameterizedFunction):
```
jbednar: We usually use function-style names for ParameterizedFunctions, to make it more natural for people to call them like functions, but here it might be more natural to have a name like `LabelFormatter`, so that users can do `LabelFormatter.replace_underscores=False`.  `default_label_formatter.replace_underscores=False` is ok, but seems less obvious to me in what it does.  But if you disagree, feel fine to ignore this comment.

param/parameterized.py
jbednar: Looks good.

param/parameterized.py
jlstevens: Just that I would expect `.instance` to be used to set the `label_formatter` and did the same internally. e.g:

```
param.parameterized.label_formatter = default_label_formatter.instance(capitalized=True)
```

param/parameterized.py
jlstevens:
```
LabelFormatter.replace_underscores=False
```

This would mean mutating the formatter in use (the actual parameterized function, not instances of it). I am ok with this approach, but right now, because of this line:

```
label_formatter = default_label_formatter.instance()
```

it won't currently work. The question is whether it is better to have a single label formatter function you mutate for customization, or a way to set any label formatter function you want, e.g a lambda function (what I've done in this PR). What *would* currently work in the style you suggest is this:

```
label_formatter.replace_underscores=False
```

Which mutates the default `default_label_formatter` instance...

jbednar: I agree that the latter would work.  But what does `.instance()` achieve, compared to simply doing `label_formatter=default_label_formatter`?  Was it to do with pickling?  I can't recall..

jlstevens: There is no particularly strong reason to use `.instance` (I didn't think it would harm anything and would isolate the original parameterized function from mutations on `label_formatter` if it were used elsewhere...which is unlikely) so I removed it in f8032bb

jbednar:
Looks ready to merge, @philippjfr!

PR #320, opened 2019-03-02
Minor bug fixes

philippjfr:
* Fixes issue with batch watch flag not being restored on exceptions
* Fixes issues with Parameterized.params not returning instance parameters because it is a classmethod


philippjfr:
I'm going to merge this as soon as it passes so I can cut a dev release which is actually backward compatible.

PR #323, opened 2019-03-08
Ensure all watchers are executed in scheduling order

philippjfr:
Fixes https://github.com/pyviz/param/issues/322

- [x] Add tests

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -78,8 +78,11 @@ def logging_level(level):
 @contextmanager
 def batch_watch(parameterized, run=True):
     """
-    Context manager which batches the watchers on the parameterized.
-    If run=True it will dispatch the accumulated events on the watchers.
+    Context manager to batch watcher events on a parameterized object.
+    The context manager will queue any events triggered by setting a
+    parameter on the supplied parameterized object and dispatch them
+    all at once when the context manager exits. If run=False the
+    queued events are not dispatched and should be processed manually.
     """
     BATCH_WATCH = parameterized.param._BATCH_WATCH
     parameterized.param._BATCH_WATCH = True
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -75,6 +75,22 @@ def logging_level(level):
         param_logger.setLevel(logging_level)
 
 
+@contextmanager
+def batch_watch(parameterized, run=True):
+    """
+    Context manager which batches the watchers on the parameterized.
+    If run=True it will dispatch the accumulated events on the watchers.
+    """
+    BATCH_WATCH = parameterized.param._BATCH_WATCH
+    parameterized.param._BATCH_WATCH = True
+    try:
+        yield
+    finally:
+        parameterized.param._BATCH_WATCH = BATCH_WATCH
+        if run and not BATCH_WATCH:
+            obj.param._batch_call_watchers()
+
+
 def classlist(class_):
     """
     Return a list of the class hierarchy above (and including) the given class.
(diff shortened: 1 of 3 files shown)


philippjfr:
One way to make this PR cleaner would be to add a context manager for setting and unsetting the BATCH_WATCH flag.

philippjfr:
Ready to review/merge.

param/parameterized.py
```diff
@@ -75,6 +75,22 @@ def logging_level(level):
         param_logger.setLevel(logging_level)
 
 
+@contextmanager
+def batch_watch(parameterized, run=True):
+    """
+    Context manager which batches the watchers on the parameterized.
+    If run=True it will dispatch the accumulated events on the watchers.
```
jbednar: This docstring is a bit obscure; I can't tell what the purpose of this context manager is from it. Can you elaborate?

param/parameterized.py
```diff
@@ -1215,6 +1238,7 @@ def set_param(self_, *args,**kwargs):
         positional arguments, but the keyword interface is preferred
         because it is more compact and can set multiple values.
         """
+        BATCH_WATCH = self_.self_or_cls.param._BATCH_WATCH
```
jbednar: Why the scary capitals?

param/parameterized.py
philippjfr: Updated the docstring, hope it's clearer now.

param/parameterized.py
philippjfr: Just following the style of the original variable name, not sure it has to be capitalized but if we decide to change I would change it everywhere.

jbednar: As you wish.

param/parameterized.py
```diff
@@ -78,8 +78,11 @@ def logging_level(level):
 @contextmanager
 def batch_watch(parameterized, run=True):
     """
-    Context manager which batches the watchers on the parameterized.
-    If run=True it will dispatch the accumulated events on the watchers.
+    Context manager to batch watcher events on a parameterized object.
+    The context manager will queue any events triggered by setting a
+    parameter on the supplied parameterized object and dispatch them
+    all at once when the context manager exits. If run=False the
+    queued events are not dispatched and should be processed manually.
```
jbednar: Much clearer, thanks!

martindurant:
Quick question on this: does the watcher get the event itself in its arguments? I would say, that an event which is going to cause another like event of a different value to be spawned should explicitly set the event to be ignored downstream. That way, if you want a "less than 4" filter on an element as you described, other watchers *never* see the value 5, even temporarily. This leaves it up to the programmer to think about whether it makes sense for an event to propagate or not. 
(This is a separate concept from batching events as above)

philippjfr:
>Quick question on this: does the watcher get the event itself in its arguments?

Yes it does, and I agree with you. I think a watcher should be able to invalidate an event it receives so that downstream watchers never see it. That said maybe a better solution would be to offer a way to define preprocessors on parameters so the value is validated before it is set and triggers downstream events. I can't quite envision if there are other scenarios where you might want a watcher to invalidate an event it receives.

martindurant:
> to offer a way to define preprocessors on parameters so the value is validated before it is set

This way means having a different type of watcher, which may be confusing. I suppose it could be a very simple function valid(val) -> True/False.

I can imagine scenarios, such as the new value caused a procedure to fire which failed, so it is OK to set that value (it's not invalid, only failing at the current moment and set of other values) but follow-on procedures should not run. You could argue this case is moot, any follow-ons should respond the a change in a separate output area...

philippjfr:
This PR also happens to fix https://github.com/pyviz/panel/issues/143.

I think we should merge this PR and then file an issue to discuss any suggestions related to letting users invalidate events.

martindurant:
That's certainly fine with me.

jlstevens:
I think I'll go ahead and merge this PR so we can test it. I agree with Jim that the earlier discussion suggests that this is the right approach.

PR #325, opened 2019-03-13
Support batch watch on Parameter attributes

philippjfr:
Currently batch watching merges all events into a single dictionary by name and matches the watcher to dispatch it to by name. This ignores the fact that there may be multiple events for different attributes of the parameter (e.g. ``value``, ``object``, ``precedence`` etc.) and that each attribute might have different watchers. 

Previously this was not problematic but in recent PRs batching is now used to ensure that events are queued correctly internally which means parameter attribute change events can end up in the batch queue.

This PR ensures that the batch watching code correctly respect the ``what`` argument, which describes which parameter attribute is triggering the event.


jlstevens:
Looks good to me! Merging.

PR #326, opened 2019-03-15
Added a step slot to param.Parameter

jlstevens:
This adds a step slot to `Number` parameters. This will be useful for any library that maps parameters to widgets (e.g panel).

TODO:

- [x] Unit tests
- [x] Validation of the step attribute/slot

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -756,8 +756,8 @@ class Number(Dynamic):
         self.bounds = bounds
         self.inclusive_bounds = inclusive_bounds
         self._softbounds = softbounds
-        self._validate(default)
         self.step = step
+        self._validate(default)
 
 
     def __get__(self,obj,objtype):
@@ -872,6 +872,9 @@ class Number(Dynamic):
         if not _is_number(val):
             raise ValueError("Parameter '%s' only takes numeric values"%(self.name))
 
+        if self.step is not None and not _is_number(self.step):
+            raise ValueError("Step parameter can only be None or a numeric value")
+
         self._checkBounds(val)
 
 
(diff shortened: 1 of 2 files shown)


jbednar:
A step can also be useful outside of widget libraries, e.g. for making cross products of possible values (as in lancet), and/or to allow the parameter to be sampled or iterated over in a meaningful way.  So the immediate application is for widgets, but having a user indicate a suggested granularity for a continuous parameter space is meaningful with or without a GUI.

jlstevens:
From our discussion, we have decided that validating this attribute would be more hassle than it is worth. I've ticked that item off the TODO list as a result.

jbednar:
There's one validation that is safe to do and should avoid downstream libraries having to implement code for this: you could at least validate that the step is of the right type (integer for integer parameters, number for Number parameters, etc.).  Actually validating that the step falls inside the bounds is just as likely to be annoying as helpful, as people won't necessarily always update the step as they adjust the bounds.

Jean-Luc [3:58 PM]

philippjfr:
>Actually validating that the step falls inside the bounds is just as likely to be annoying as helpful, as people won't necessarily always update the step as they adjust the bounds.

Yeah I agree with this. The other point is that most widget libraries will gracefully handle a step that is too large (usually snapping to the first and last value), while they will error for a step of the wrong type.

jlstevens:
@jbednar @philippjfr Ready to review/merge when the tests are green.

param/__init__.py
```diff
@@ -870,6 +872,9 @@ def _validate(self, val):
         if not _is_number(val):
             raise ValueError("Parameter '%s' only takes numeric values"%(self.name))
 
+        if self.step is not None and not _is_number(self.step):
```
philippjfr: Does setting a Parameter attribute actually trigger this validation? I'm not actually sure we implement that. Something for another PR of course but that seems like a pretty big hole in the validation now that it's really easy to set parameter attributes dynamically.

jlstevens: That is a good point. It seems that having to validate in multiple places is pretty redundant from a code perspective - I'll now have a look to see where that is.

philippjfr:
Looks good thanks.

PR #327, opened 2019-03-15
Ensure that param watch decorator is resolved using MRO

### LINKED ISSUE #321: depends watching does not work when defined on superclass
When a method decorated with param.depends and watch=True is defined on a superclass, subclasses will not trigger the event:

```python
class A(param.Parameterized):

    count = param.Integer(default=0)

    add = param.Action(default=lambda x: x.param.trigger('add'))

    @param.depends('add', watch=True)
    def _add(self):
        self.count +=1

class B(A):
    pass
```

Then if we try the action we can see that the method is not triggered:

```python
> b = B()
> b.add(b)
> b.count
0
```

Moving the decorated method onto the subclass fixes the issue.

philippjfr:
Fixes #321 

- [x] Adds unit tests

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2246,14 +2246,17 @@ class Parameterized(object):
         object_count += 1
 
         # add watched dependencies
-        for n in self.__class__.param._depends['watch']:
-            # TODO: should improve this - will happen for every
-            # instantiation of Parameterized with watched deps. Will
-            # probably store expanded deps on class - see metaclass
-            # 'dependers'.
-            for p in self.param.params_depended_on(n):
-                # TODO: can't remember why not just pass m (rather than _m_caller) here
-                (p.inst or p.cls).param.watch(_m_caller(self,n),p.name,p.what)
+        for cls in classlist(self.__class__):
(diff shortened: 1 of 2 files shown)


param/parameterized.py
```diff
@@ -2246,14 +2246,17 @@ def __init__(self,**params):
...
-                # TODO: can't remember why not just pass m (rather than _m_caller) here
-                (p.inst or p.cls).param.watch(_m_caller(self,n),p.name,p.what)
+        for cls in classlist(self.__class__):
+            if not issubclass(cls, Parameterized):
+                continue
+            for n in cls.param._depends['watch']:
+                # TODO: should improve this - will happen for every
+                # instantiation of Parameterized with watched deps. Will
+                # probably store expanded deps on class - see metaclass
+                # 'dependers'.
```
philippjfr: Maybe I should look into these issues.

jlstevens:
Looks good to me! The issue mentioned in the comment doesn't have to be addressed as part of this PR - I'm happy to merge once the tests pass.

jlstevens:
Tests have passed. Merging.

PR #329, opened 2019-03-29
Resolve annotated outputs using MRO

### LINKED ISSUE #328: @param.output and inherited classes
I have two classes that both contain a `@param.output()` statement. I have a third class that inherits from both. What is the expected output behavior from the third class?   
  
I'm working with annotators from [Earthsim](https://github.com/pyviz-topics/EarthSim) : 
   
`PointAnnotator` has `@param.output(points=gv.Points)`    
`PolyAnnotator` has `@param.output(path=hv.Path)`    
PolyAndPointAnontator inherits from both: `PolyAndPointAnnotator(PolyAnnotator, PointAnnotator)`    
    
Is it expected that `PolyAndPointAnnotator` have two outputs?   
  
That was my assumption but I couldn't get it to pass to the next pipeline stage properly. I'm wondering if I need to continue debugging or is this not possible?

philippjfr:
Fixes #328 

- [x] Adds test

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1562,16 +1562,18 @@ class Parameters(object):
         index into the output if multiple outputs are returned.
         """
         outputs = {}
-        for name in dir(self_.self_or_cls):
-            method = getattr(self_.self_or_cls, name)
-            dinfo = getattr(method, '_dinfo', {})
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good and tests are green. Merging.

PR #330, opened 2019-03-29
No longer use root logger

philippjfr:
Supersedes https://github.com/pyviz/param/pull/308

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -40,6 +40,7 @@ def get_logger(name=None):
     if logger is None:
         root_logger = logging.getLogger('param')
         if not root_logger.handlers:
+            root_logger.setLevel(logging.INFO)
             formatter = logging.Formatter(
                 fmt='%(levelname)s:%(name)s: %(message)s')
             handler = logging.StreamHandler()
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -31,17 +31,27 @@ except:
 VERBOSE = INFO - 1
 logging.addLevelName(VERBOSE, "VERBOSE")
 
-# Logger instance to use for param; if "logger" is set to None, the root logger
-# will be used.
+# Get the appropriate logging.Logger instance. If `logger` is None, a
+# logger named `"params"` will be instantiated. If `name` is set, a descendant
+# logger with the name ``"params.<name>"`` is returned (or
+# ``logger.name + ".<name>"``)
 logger = None
-def get_logger():
+def get_logger(name=None):
     if logger is None:
-        # If it was not configured before, do default initialization
-        if not logging.getLogger().handlers:
-            logging.basicConfig(level=INFO)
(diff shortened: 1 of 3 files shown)


param/parameterized.py
```diff
@@ -31,17 +31,27 @@
...
+# logger named `"param"` will be instantiated. If `name` is set, a descendant
+# logger with the name ``"param.<name>"`` is returned (or
+# ``logger.name + ".<name>"``)
 logger = None
-def get_logger():
+def get_logger(name=None):
     if logger is None:
-        # If it was not configured before, do default initialization
-        if not logging.getLogger().handlers:
-            logging.basicConfig(level=INFO)
```
philippjfr: Should we keep this?

param/parameterized.py
```diff
@@ -1684,14 +1693,12 @@ def __db_print(self_,level,msg,*args,**kw):
...
-        if get_logger().isEnabledFor(level):
+        if get_logger(name=self_or_cls.name).isEnabledFor(level):
 
             if dbprint_prefix and callable(dbprint_prefix):
-                prefix=dbprint_prefix() # pylint: disable-msg=E1102
-            else:
-                prefix=""
+                msg = dbprint_prefix() + ": " + msg  # pylint: disable-msg=E1102
 
-            get_logger().log(level, '%s%s: '+msg, prefix, self_or_cls.name, *args, **kw)
```
jbednar: Is this going to change the formatting of log messages?

param/parameterized.py
jbednar: Now that we're making our own logger, I think we should explicitly set the default, yes.

param/parameterized.py
philippjfr: Yes, a warning now looks like this:

>WARNING:param.Curve01322: Some warning

It used to look like this:

>WARNING:root:Curve01315: Some warning

I'd prefer it if it was something like:

>WARNING:param:Curve01315: Some warning

but the ``.`` separator is the canonical way to define logger descendants.

jbednar: Ok.  Sounds fine, and I agree with your preference, but it's not too bad.

jbednar:
@jlstevens , I think this is mergeable, but if you have any objections, please make that clear asap!  

I think the changes to the warning messages mean that this release should be 1.9 and not 1.8.3, because even though it's not strictly part of the API people may still be relying on the warning message format using scripts that grep for warnings or errors, so it's important to acknowledge the change.

philippjfr:
>I think the changes to the warning messages mean that this release should be 1.9 and not 1.8.3

This was already the plan.

jlstevens:
I think this is a good idea and I'm happy to merge now. If this PR results in some unexpected behavior somewhere, I think that is more acceptable than modifying the root logger which can cause all sorts of problems. Merging.

PR #334, opened 2019-05-14
Allow param.depends to annotate functions

philippjfr:
This PR extends the `depends` decorator to allow it to decorate regular functions not just methods on Parameterized classes. Unlike the usage with a method the decorator must be given parameter instances associated with a Parameterized object. This makes it very easy to tie a simple function to the state of a Parameterized class, scheduling callbacks, and if used in conjunction with Panel makes a reactive pattern possible without requiring users to write entire classes.

The usage with Panel is the primary motivator here, so let us see some examples. Here is a simple example tying the value of a ``FloatSlider`` to the function: 

screencast 2019-05-14 12-20-04

Another simple example is this implementation of a plot selector using hvplot:

reactive_plot

The decorator allows defining both regular arguments and keyword arguments to be passed to the function and also supports the ``watch`` keyword to schedule a callback.

- [x] Add tests

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -262,9 +262,9 @@ def ismethod(obj):
     if sys.version_info.major == 2:
         return hasattr(obj, 'im_class')
     else:
-        if '.' in obj.__qualname__:
-            return True
-        return 'self' in inspect.signature(obj).parameters
+        owner = getattr(inspect.getmodule(meth),
+                        meth.__qualname__.split('.<locals>', 1)[0].rsplit('.', 1)[0])
+        return isinstance(owner, type)
 
 
 @accept_arguments
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -27,6 +27,7 @@ try:
 except:
     param_pager = None
 
+basestring = basestring if sys.version_info[0]==2 else str # noqa: it is defined
 
 VERBOSE = INFO - 1
 logging.addLevelName(VERBOSE, "VERBOSE")
@@ -258,15 +259,6 @@ def instance_descriptor(f):
     return _f
 
 
-def ismethod(obj):
-    if sys.version_info.major == 2:
-        return hasattr(obj, 'im_class')
-    else:
-        owner = getattr(inspect.getmodule(meth),
-                        meth.__qualname__.split('.<locals>', 1)[0].rsplit('.', 1)[0])
-        return isinstance(owner, type)
-
-
 @accept_arguments
 def depends(func, *dependencies, **kw):
     """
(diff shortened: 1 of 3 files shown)


jbednar:
Beautiful.  I think this will really be useful.  Is it necessary to have ".value" in "slider.param.value", or could we also accept "slider.param"?  It seems like it's nearly always ".value" that we'd want to depend on by default, and having to specify it feels odd.

Also, I think the docs should have a non-trivial example implemented using all of the now five different Panel development styles:

1. `interact`
2. `interact` but re-packed into a compact dashboard
3. `@depends` with bare functions
4. `@depends` with a Parameterized class
5. `.watch` with bare widgets

Having the same thing done each way will help people figure out the similarities and differences between these approaches, to choose the right one for their own projects.

philippjfr:
>I think this will really be useful. Is it necessary to have ".value" in "slider.param.value", or could we also accept "slider.param"?

Yes, I absolutely do not want special handling for specific classes like panel widgets, you are listening to the value parameter so you should specify it.

philippjfr:
>Also, I think the docs should have a non-trivial example implemented using all of the now five different Panel development styles

I strongly agree, not quite sure how far to push it on the non-trivial front yet though.

jbednar:
(I moved that panel-docs comment to the Panel issue, as I didn't notice that this was just on Param.)

param/parameterized.py
```diff
@@ -258,33 +258,70 @@ def _f(self, obj, val):
     return _f
 
 
+def ismethod(obj):
+    if sys.version_info.major == 2:
+        return hasattr(obj, 'im_class')
+    else:
+        owner = getattr(inspect.getmodule(meth),
+                        meth.__qualname__.split('.<locals>', 1)[0].rsplit('.', 1)[0])
+        return isinstance(owner, type)
```
philippjfr: Have to think about this, since py3 does not have the concept of an unbound method we need this hack to find out whether we are decorating a simple function or not. It works well but I'm not entirely convinced it's perfectly robust. If we give up on this then two things would need to happen:

1. We either do not allow keywords at all (only args) or we pointlessly also allow them for methods
2. We do not support ``watch=True`` for functions.

jbednar:
> Yes, it is necessary, I absolutely do not want special handling for specific classes like panel widgets, you are listening to the value parameter so you should specify it.

Good point; Param definitely should not know about Panel.  I guess a Panel widget could have a special property that returns the slider's value parameter instance, e.g. `slider.value`, but that's arbitrary and just seems like it would cause confusion.  Really it would be nice if somehow we could just pass `slider`, and slider would support some general interface that Param recognizes but is not specific to Panel.  I can't think of any clean way to achieve that; basically it would mean a Panel widget masquerading as a param instance.

philippjfr:
Since I cannot reliably distinguish between an unbound method and a regular function in Python 3, I've decided to unify the implementation with the following semantics:

* You can supply either string specs OR parameter instances, a mixture of both is **not** supported.
* When supplying string specs:
  - The method is called without arguments (as it is currently)
  - It is not allowed to provide string specs as keywords (as it is currently)
* When supplying parameter instances:
  - The method is called with the args **and kwargs** defined in the decorator
  - 

The main reason you can't mix string specs and parameter instances because it is ambiguous whether or not these would then be passed in as arguments or not. If we decide that mixing them should be supported the only reasonable resolution to that issue would be to say that if there are any parameter instances in the decorator then it will pass everything in as explicit arguments (which might be reasonable?).

The only other question I have is whether we inspect the function signature to ensure you don't define mismatching depends and function specs?

jbednar:
That all sounds reasonable to me. Using inspect on the function signature would be helpful for debugging, but doesn't seem absolutely required; I'll leave that to you.

jlstevens:
That also sounds reasonable to me. 

My only question is whether you would consider having two decorators, one for methods and one for functions to help split the semantics accordingly. We haven't been fully happy with the name `depends` (and I've gotten less happy with it the longer I've been in the US!). At one point I remember thinking of a name I wish we had used instead but unfortunately, I've currently forgotten what that was!

philippjfr:
@jbednar insisted pretty strongly that it should be one decorator and I think I agree, I think the semantics are clearly enough separated with this proposal that it's not a major issue. Apart from the unfortunate association with the US brand name I think I quite like it since it conveys the concept quite well, certainly prefer it over the ``callback`` decorator in plotly dash.

philippjfr:
If we're all happy with it then this is ready for review.

jlstevens:
Ok, I was just double checking! The suggested API seems to be a decent compromise and I'll review the code shortly...

param/parameterized.py
```diff
@@ -261,29 +262,64 @@ def _f(self, obj, val):
...
+                  (isinstance(dep.owner, ParameterizedMetaclass))):
+            owner = 'None' if dep.owner is None else '%s class' % type(dep.owner).__name__
+            raise ValueError('Reactive functions input parameters '
+                             'must be associated with a Parameterized '
+                             'instance, parameter owner is %s.' % owner)
+
+    if len({type(dep) for dep in deps}) > 1:
+        raise ValueError('Dependencies must either be defined as strings '
+                         'referencing parameters on the class defining '
+                         'the decorated method or as parameter instances, '
```
jlstevens: Full stop instead of a comma here and a new sentence for 'Mixing of string specs and parameter instances is not supported.'?

param/parameterized.py
```diff
@@ -261,29 +262,64 @@ def _f(self, obj, val):
...
+        elif not isinstance(dep, Parameter):
+            raise ValueError('Reactive functions must have '
+                             'only parameters as dependencies. '
+                             'Found %s type instead.' %
+                             type(dep).__name__)
+        elif not (isinstance(dep.owner, Parameterized) or
+                  (isinstance(dep.owner, ParameterizedMetaclass))):
+            owner = 'None' if dep.owner is None else '%s class' % type(dep.owner).__name__
+            raise ValueError('Reactive functions input parameters '
+                             'must be associated with a Parameterized '
```
jlstevens: Little hard for a user to understand I think. Is this equivalent to 'Supplied parameters must be bound to a class or object'?

param/parameterized.py
```diff
@@ -261,29 +262,64 @@ def _f(self, obj, val):
...
+            dep_kwargs = {n: getattr(dep.owner, dep.name) for n, dep in kw.items()}
+            return func(*args, **dep_kwargs)
+
+        for dep in deps:
+            dep.owner.param.watch(cb, dep.name)
+
+    _dinfo = getattr(func, '_dinfo', {})
+    _dinfo.update({'dependencies': dependencies,
+                   'kw': kw, 'watch': watch})
+
```
jlstevens: Good to see all the new logic appearing as one main block.

param/parameterized.py
philippjfr: Ah, yes it's also outdated, you can supply a Parameter associated with a class instead of an instance too now.

jlstevens:
Looks good to me! My only suggestions relate to making the exception messages a little easier for users to understand.

philippjfr:
@jlstevens I revised the exceptions. Let me know if you're happy with them.

jlstevens:
Great! I'll go ahead and merge this now.

PR #339, opened 2019-05-20
Replace assert with real error

philippjfr:


jlstevens:
Simple PR and looks good to me. Any reason not to immediately merge @jbednar ?

PR #342, opened 2019-05-31
Added context managers to manage events and edit constant parameters

philippjfr:


param/parameterized.py
```diff
@@ -106,6 +106,44 @@ def batch_watch(parameterized, run=True):
             parameterized.param._batch_call_watchers()
 
 
+@contextmanager
+def edit_constant(parameterized):
```
jbednar: That name should work, thanks.

param/parameterized.py
```diff
@@ -106,6 +106,44 @@ def batch_watch(parameterized, run=True):
...
+        raise
+    finally:
+        for (p, const) in zip(params, constants):
+            p.constant = const
+
+
+@contextmanager
+def discard_events(parameterized):
+    """
+    Context manager which discards any events within its scope
```
```suggestion
    Context manager that discards any events within its scope
```

Issue #346, opened 2019-06-18
param.DateRange  should accept date object - or at least raise a warning


Casyfill:
right now param.DateRange accepts date objects without any warnings - but does not work that way; I feel it is reasonable to expect it to work with dates, given the name :-)

jbednar:
See example of the problem at https://github.com/pyviz/panel/issues/486 .

maximlt:
Fixed by https://github.com/holoviz/param/pull/341

PR #348, opened 2019-07-05
Date and datetime

randomstuff:
Rename `Date` into `DateTime` and add `CalendarDate` (for `date`).

Same for `DateRange`.

For compatiblity `Date` is an alias for `DateTime`. The idea is to rename `CalendarDate` into `Date` later on.

see #346 and #168

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1812,7 +1812,7 @@ class MultiFileSelector(ListSelector):
 
 class Date(Number):
     """
-    Date parameter of datetime type.
+    Date parameter of datetime or date type.
     """
 
     def __init__(self, default=None, **kwargs):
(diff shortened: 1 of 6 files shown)


philippjfr:
This looks great to me. @jbednar @jlstevens Any objections?

randomstuff:
(Merged the first two commits)

param/__init__.py
```diff
@@ -1940,13 +1970,42 @@ def _validate(self, val):
...
+
+# Compatibility:
+DateRange = DateTimeRange
+
+
+class CalendarDateRange(Range):
+    """
+    A date range specified as (start_date, end_date).
+
+    Dates must be specified as datetime-like types (see
```
jbednar: Should this be "datetime" or "date"? The latter seems correct to me, but if the former is true, it's fine.

randomstuff:
If this change is merges I'm not sure it makes sense to allow `DateTimeRange` to hold a `date` value instead of a `datetime` one. I find it's preddy strange:

```python
import datetime

import param

    p = param.DateTime()

foo = Foo()
foo.p = datetime.date.today()
print(repr(foo.p))
# => datetime.date(2019, 7, 25)
```

Changing this would be a breaking change however.

randomstuff:
Fixed the documentation of the changes which has some confusing regarding datetime vs date, etc.

randomstuff:
Another solution would be to keep `Date` dand `DateRange` for now (with the idea or replacing them later on with `CalendarDate` and `CalendarDateTime`) which would support both for now and add new `DateTime` and `DateTimeRange` which would only accept `datetime`

randomstuff:
Updated to this alternative proposition:

* keep `Date` unmodified;
* `CalendarDate` for `date`-only.

I'd be tempted to add (in other PRs):

* [`NaiveDateTime`](https://www.w3.org/TR/html52/sec-forms.html#local-date-and-time-state-typedatetimelocal) for naive date times (no timezone information);
* `DateTime` (?) for non-naive datetime;
* [`Time`](https://www.w3.org/TR/html52/sec-forms.html#time-state-typetime) for (naive) `time`.

philippjfr:
Thanks @randomstuff! Sorry for the long review cycle. This is hugely appreciated.

randomstuff:
I was looking of the different `panel` widgets and how they would behave if using `CalendarDate` instead of `Date` this change and I discovered that `datetime` is actually a super class of `date`. (Which is not actually super useful but whatever …).

As a consequence, `CalendarDate` can currenly hold `datetime` values which is not what I intended. The only difference between `Date` and `CalendarDate` is that the latter cannot hold a `np.datetime64` value (because `np.datetime64` is not a super-class of `date`). This is somewhat nice because `np.datetime64` does not behave as a `date` at all (no `.month`, etc.).

Conclusion:

* As `datetime` is a superclass of `date`, it is kind-of expected that the `Date`/`CalendarDate` parameters would accept `datetime` values so the `Date` parameter type is actually OK in this regard. I wonder if it would not be useful to automatically convert `datetime` to `date` on assignment (either for `Date`/`CalendarDate`).

*  I'm not sure `CalendarDate` is actually any useful in its current state and I'd vote for reverting it for now (at least before release if no clear plan as been found).

* Allowing `Date` parameters to hold `np.datetime64` is not super useful because `np.datetime64` do not behave as `date` at all. It might be more useful to convert `np.datetime64` to `datetime`/`date` automatically when assigning.

Issue #322, opened 2019-03-08
Watcher may be called after newer event has been processed


philippjfr:
In looking into https://github.com/pyviz/panel/issues/287 I discovered a very tricky problem with order in which events are currently processed. In the referenced example there are two Parameterized objects (the widget and the users parameterized instance), which are bi-directionally synced. Additionally there is a callback on the user defined parameter which validates the parameter value and set it to 3 if the set value > 4. 

Now a user sets the parameter value to 5, which currently triggers this cascade of events:

1. The Parameterized has two watchers, it starts by invoking the first watcher which sees value > 4 and sets the parameter value to 3
2. This second change in the parameter value invokes the watchers again, but this time the validation is satisfied, so the second watch callback is invoked
3. The second watcher updates the widget, setting it to the correct value of 3
4. Now, here is the problematic part, the parameterized now finally gets around to processing the second watcher that was invoked in step 1., now updating the widget to a value of 5

Here is a terrible drawing of what happens, ``P`` is the Parameterized instance, ``W`` is the widget instance, arrows indicate events, lines inside a node indicate watchers, black numbering corresponds to the list above, orange number corresponds to the value being sent:

![20190308_022200](https://user-images.githubusercontent.com/1550771/54003238-5d956280-4149-11e9-89fa-615819893191.jpg)

The problem therefore is that an old event arrives **after** a newer event has already overridden the value and therefore leaves objects partially in the wrong state. Both solutions I can think of here are painful:

1. There is a global execution queue for watchers to ensure they are always invoked in the sequence they originally arrived in. When a watcher is triggered it gets put on the queue, and the code recursively processes the queue until it is empty, if a newer event for the same parameter change is encountered the old event is discarded from the queue. 
2. Events have timestamps and parameterized's keep track of the last event time for each parameter, if an event that's older than the last recorded event time arrives it is discarded.

philippjfr:
I just thought of a third, less bad option:

3. Keep an execution count on each Watcher, if the count increases between a watcher being scheduled and actually being triggered skip it

philippjfr:
I guess the most obvious solution I left out:

[...]

PR #351, opened 2019-07-31
Add eager option to watch calls

philippjfr:
In order to fix https://github.com/pyviz/param/issues/322 I implemented a scheme by which watch calls would execute in the order the user scheduled them. This meant that if a watch callback triggered another event, that event would not get called until all previous events had been processed. 

This fix was important for certain (relatively rare) scenarios and ensured correctness, however it also has the quite annoying side-effect that you cannot define a watch callback which temporarily sets and unsets a parameter that has some side-effect, e.g. when you click a button in panel you might want to disable that button while you perform some computation (e.g. see #350), in this scenario you want param to eagerly execute any callbacks as they are triggered. The distinction here is akin to depth-first vs. breadth-first execution of callbacks.

To control this I've added an eager keyword to watch calls. I'd be happy to rename that to something that reflects the depth- vs. breadth-first distinction. Additionally I'm now also not clear whether the current default of breadth-first execution is more desirable, and would consider changing the default. I think it's a very common usecase particularly in panel to trigger some temporary side-effect in a callback and teaching users about the execution model seems difficult.

Closes https://github.com/pyviz/param/issues/350

- [x] Decide on API and naming
- [x] Decide whether to change the default
- [x] Add tests

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -88,7 +88,7 @@ def logging_level(level):
 
 
 @contextmanager
-def batch_watch(parameterized, run=True):
+def batch_watch(parameterized, enable=True, run=True):
     """
     Context manager to batch watcher events on a parameterized object.
     The context manager will queue any events triggered by setting a
@@ -97,7 +97,7 @@ def batch_watch(parameterized, run=True):
     queued events are not dispatched and should be processed manually.
     """
     BATCH_WATCH = parameterized.param._BATCH_WATCH
-    parameterized.param._BATCH_WATCH = True
+    parameterized.param._BATCH_WATCH = enable or parameterized.param._BATCH_WATCH
     try:
         yield
     finally:
@@ -112,7 +112,7 @@ def edit_constant(parameterized):
     Temporarily set parameters on Parameterized object to constant=False
     to allow editing them.
     """
-    params = parameterized.objects('existing').values()
+    params = parameterized.param.objects('existing').values()
     constants = [p.constant for p in params]
     for p in params:
         p.constant = False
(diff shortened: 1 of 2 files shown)


philippjfr:
It also seems like depth-first execution is what a regular user would naively expect, because naively you wouldn't expect an event queue to handle the bread first case for you.

jbednar:
I don't have an opinions about the name. It's worth working out which scenarios work better in eager=True, and which work better in eager=False, and try to estimate which of those are more common scenarios.  If they are about equally common, I'd say to keep the current default behavior.  If one of the categories is way more common and typical than the other, favor that one.  I don't have a good idea myself which way that falls.

philippjfr:
I think depth-first is way more common, wanting some side-effect to take place while you run some callback is very common, e.g. change the style or color of a widget or disable it while it executes is a common usecase. Even in pure param usage you may want a callback to trigger some other side-effect on the same class. Comparing that to the breadth-first case I can't think of many cases where that would occur, you will note for example that I had to draw out a whole diagram for #322 to even visualize and understand the three way interaction where breadth-first execution is important.

jbednar:
What about a busy indicator or progress bar; sounds like easier in depth-first?

philippjfr:
>What about a busy indicator or progress bar; sounds like easier in depth-first?

If the busy indicator lives on the same class yes. One thing I didn't explain well enough that the depth vs breadth first execution model only matters for events happening on the same object, for everything else it's always depth-first. That inconsistency is, I believe another good reason to go for depth-first.

jbednar:
Ok, sounds good.

jlstevens:
Instead of `eager` (which makes me think of eager versus lazy evaluation) how about `queued`? By default (currently) it sounds like things go into a queue so `queued=False` would mean 'eager'.

PR #354, opened 2019-08-16
Import abstract collections from collections.abc where possible

sdrobert:
In Python 3.3, abstract collections were moved from the ``collections`` module to the submodule ``collections.abc``. In 3.7, importing abstract collections directly from ``collections`` results in a ``DeprecationWarning``. In python 3.8, ``collections`` will no longer contain abstract base classes.

This PR tries to import and use ``collections.abc``. If it fails (python 2.7), it falls back to ``collections``

I was tired of DeprecationWarnings :D

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -54,6 +54,12 @@ try:
 except:
     pass
 
+
+try:
+    import collections.abc as collections_abc
+except ImportError:
+    collections_abc = collections
+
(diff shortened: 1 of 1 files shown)


jbednar:
Looks good, thanks!  I"m not seeing those warnings on 3.7, but the changes look good regardless!

PR #358, opened 2019-10-15
added some tests and assertions that in testwatch

thomasdeneux:
I have learnt about param's watching mechanism by inspecting testwatch.py
Here are a few tests and assertions i added to further understand it, i thought they might be useful to others as well.

I also removed some blank lines to satisfy PEP8 recommandation (and remove IDE code inspection highlights)


philippjfr:
Thanks, looks good to me!

PR #360, opened 2019-11-07
Fix ClassSelector.get_range() for class_ tuple

philippjfr:
ClassSelector allowed using `class_=some_tuple` but ClassSelector.get_range() would error in that case.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1335,8 +1335,11 @@ class ClassSelector(SelectorBase):
         Only classes from modules that have been imported are added
         (see concrete_descendents()).
         """
-        classes = concrete_descendents(self.class_)
-        d=OrderedDict((name,class_) for name,class_ in classes.items())
+        classes = self.class_ if isinstance(self.class_, tuple) else (self.class_,)
+        all_classes = {}
+        for cls in classes:
+            all_classes.update(concrete_descendents(cls))
+        d=OrderedDict((name,class_) for name,class_ in all_classes.items())
         if self.allow_None:
             d['None']=None
         return d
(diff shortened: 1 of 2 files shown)


ceball:
Looks correct to me.

I've tried to propose a better first line of the docstring - maybe someone can do better?

Current:
```
    """
    Parameter whose value is a specified class or an instance of that class.
    By default, requires an instance, but if is_instance=False, accepts a class instead.
    Both class and instance values respect the instantiate slot, though it matters only
    for is_instance=True.
    """
```

Proposed:
```
    """
    Parameter whose value is - or is an instance of - class_, where class_ is a single class or tuple of classes.
    By default, requires an instance, but if is_instance=False, accepts a class instead.
    Both class and instance values respect the instantiate slot, though it matters only
    for is_instance=True.
    """
```

Starting again maybe the `class` slot would be called something like `classinfo` instead (like `isinstance()`'s `classinfo` parameter).

param/__init__.py
```diff
@@ -1335,8 +1335,11 @@ def get_range(self):
         Only classes from modules that have been imported are added
         (see concrete_descendents()).
         """
-        classes = concrete_descendents(self.class_)
-        d=OrderedDict((name,class_) for name,class_ in classes.items())
+        classes = self.class_ if isinstance(self.class_, tuple) else (self.class_,)
```
jbednar: Is this the same as `classes = tuple(self.class_)`?  If so this simpler version seems more readable and clear to me.

philippjfr: It's not the same no.

Issue #353, opened 2019-08-08
Subclassing Parameter


thomasdeneux:
Hello, 

i try to add my own information to `param.Parameter` instances by subclassing it

here is my attempt (the additional information that i want to store in parameters is whether they should remain enabled in the graphic interface when a run has started)

```
import param as pm

# Add run_enabled slot to pm.Parameter classes
class RunEnabledInfo:

    __slots__ = ['run_enabled']

    def __init__(self, *args, run_enabled=None, **kwargs):
        super(MyBoolean, self).__init__(*args, **kwargs)
        self.run_enabled = run_enabled

class MyBoolean(pm.Boolean, RunEnabledInfo):
    pass

class MyNumber(pm.Number, RunEnabledInfo):
    pass
```

But i get

```
Traceback (most recent call last):
  File "E:/Users/Thomas/Python/Naivia/AlphaI/test.py", line 18, in <module>
    class MyBoolean(pm.Boolean, RunEnabledInfo):
  File "C:\ProgramData\Anaconda3\lib\site-packages\param\parameterized.py", line 503, in __new__
    return type.__new__(mcs,classname,bases,classdict)
TypeError: multiple bases have instance lay-out conflict
```

The problem seems to be with `MyBoolean` inheriting from two classes. Indeed the following code works, but i wouldn't like to rewrite it for every possible parameter type! Any suggestion how to do?

```
class MyBoolean(pm.Boolean):

    __slots__ = ['run_enabled']

    def __init__(self, *args, run_enabled=None, **kwargs):
        super(MyBoolean, self).__init__(*args, **kwargs)
        self.run_enabled = run_enabled
```

jbednar:
Can't you use the parameter's `precedence` value to determine if they should be shown, without having to do any subclassing?

philippjfr:
>should remain enabled in the graphic interface when a run has started

In Panel setting ``parameter.constant=True`` does this.

thomasdeneux:
@jbednar Sure i could use an available slot such as `precedence`, but i might need in the future to add more than one to my parameters, so this is why i try to subclass

@philippjfr I am not using Panel for my GUI, and in any case the `run_enabled` attribute i want to add should really be set at the parameter level and not at the GUI level, because it is needed in other places (for example, when there is no active run and a parameter is changed, to infer whether some reinitialization is needed before the next run)

jbednar:
[...]

Issue #106, opened 2015-03-07
ObjectSelector's allow_None=None


ceball:

jlstevens:
I assume this refers to `allow_None` needing to be validated against the allowable objects in `ObjectSelector`.

ablythed:
Closing for lack of description.

PR #366, opened 2019-12-13
Bug fixes for ParamQt

thomasdeneux:
Relates to discussion #353: I propose Qt support for param. The pull request is only about the first 3 commits here, which are some bug fixes that i needed in param. Then it is for convenience that i put the ParamQt.py file in the 4th commit, but it is not meant actually to be added to param.


philippjfr:
Sorry we haven't really reviewed this. We have never really mixed up GUI code with param itself, so if this is something you are interested in continuing to work on it would be better to create a new ParamQt package, just like there is already a largely unmaintained ParamTk package.

MLspike:
Hello, the idea of this PR is not to add ParamQt.py to param, but to allow you have a look at what it does, give me some comments, manifest your degree of interest for this development. And from there for sure a new repository will be created.
Besides, i needed to make a few minor corrections to param itself and those should be merged into param for ParamQt to work properly.

MLspike:
See discussion #353 of course.

ceball:
I think @MLspike is right to have this PR here, for visibility. Otherwise, how would anyone looking at param know a Qt wrapper exists (which might result in collaboration etc)? 

But as @philippjfr points out, another option could be splitting this PR up, e.g.
  1. A PR against param's docs and/or readme to link to known GUI toolkits (ParamQt, [ParamTk](http://ioam.github.io/paramtk/)[1], [Panel](https://panel.holoviz.org/)). @MLspike would have to make a ParamQt project first, though.
  2. One or more PR(s) with param fixes. I bet these would be likely to get reviewed.

Just to be clear, I'm not personally interested in desktop apps at the moment, so I'm not requesting any action at all, and I can't review any related code. But I know desktop apps are interesting to many people (and are the only option in certain scenarios), so it seems like a really great thing to make ParamQt available to people. Plus it will no doubt lead to fixes/improvements in param itself.

[1] Note: ParamTk's docs say that while ParamTk was used successfully, it was a proof of concept. It's still worth noting its existence, though, since there isn't a replacement. Whereas the other two GUI wrappers I remember, ParamNb and ParamBokeh, can both thankfully be eliminated from memory thanks to Panel :)

thomasdeneux:
First sorry i mistakenly posted this morning under the @MLspike login.

I agree with @ceball suggestion and will make new pull requests: a first one for the needed param corrections, a second one, later, after a separate ParamQt repository will be created, to add information about it.

Yet i have one global question about nested parameters that i need to continue discussing with the `param` team, and the discussion #353 needs to be reactivated. I did it now, please have a look.

param/__init__.py
```diff
@@ -1164,8 +1164,9 @@ class ObjectSelector(SelectorBase):
 
     # ObjectSelector is usually used to allow selection from a list of
     # existing objects, therefore instantiate is False by default.
-    def __init__(self,default=None,objects=None,instantiate=False,
-                 compute_default_fn=None,check_on_set=None,allow_None=None,**params):
+    def __init__(self, default=None, objects=None, instantiate=False,
+                 compute_default_fn=None, check_on_set=None, allow_None=False,
```
ceball, line 1168: Team: looks like https://github.com/holoviz/param/issues/106, from 2015

thomasdeneux: Sure, but i see no comment in #106 that would let me know if there was a reason why default value for `allow_None` is None. I do not see any reason and consider it an error, and i wanted to change it to be able to compute the number of possible values as `param.allow_None + len(param.objects)`, instead of having to write `bool(param.allow_None) + len(param.objects)`.

Now i see that the default value needs also to be changed in `Selector.__init__`, i add a commit for this.

tonyfast:
closing in favor of the aforementioned pull requests.

PR #368, opened 2020-01-10
Adding Travis-Ci Support For Arm64

ossdev07:
Travis-CI has added support for ARM64. Added ARM64 job in Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,11 @@ jobs:
         - tox -e $TOX_ENV
 
     - <<: *default
+      arch: arm64
+      python: 3.7
+      env: TOX_ENV=py37
+
+    - <<: *default
       python: 3.8
       env: TOX_ENV=py38
 


ossdev07:
Only one Travis-CI job is failing and that is because it failed to download python3.7 from Travis archives. If we will re-run the railing test, it will pass.

ceball:
@ossdev07 I have the same requests of you here as in https://github.com/pyviz-dev/pyct/pull/78: please explain why this is valuable and worth the increased build times.
 
The ARM64 jobs for param are e.g. 2.5 mins or 7.5 mins vs just 30 seconds for AMD64 jobs. Granted, there isn't a lot of activity right now, so it's not the end of the world. But note that on some holoviz projects it might be prohibitive. (By the way, what about trying ARM64 for datashader - that might be more interesting :P )

ceball:
Also, param has a matrix of jobs; I'm sure not all jobs need an ARM64 version - maybe just pick one?

jbednar:
I've restarted that job. 

I agree; I think we'd only want ARM tests to be enabled for one Python version (3.7, presumably), and we could add more in the unlikely case that an actual user files a bug for ARM64 on some other version of Python.

ossdev07:
ping @ceball, this PR will increase the scope of testing by adding Travis-CI job for ARM64 architecture. As per your suggestion, I have modified the PR to enable ARM tests on only one Python version(3.7).

.travis.yml
```diff
@@ -25,6 +25,17 @@ jobs:
...
+      arch: amd64
+      stage: test
+      python: 3.7
+      env: TOX_ENV=py37
+      install:
+        - pip install tox
+      script:
+        - tox -e $TOX_ENV
+
+    - &default
```
ceball: Duplicate (see line 27)

.travis.yml
```diff
@@ -34,38 +45,48 @@ jobs:
         - tox -e $TOX_ENV
 
     - <<: *default
+      arch: amd64
```
ceball: redundant (here and in all the subsequent blocks)

.travis.yml
ossdev07: Addressed the comments.

.travis.yml
ossdev07: Addressed the comments.

ceball:
@ossdev07 Thanks for fixing - I think we can merge this now.

The new ARM64 job takes about 120 seconds, while the rest take about 30. If that becomes a problem in the future, we could change the ARM64 job to just be a daily cron job (param's unlikely to grow some new functionality that would need to be tested on ARM64 before being merged...).

jbednar:
Thanks!

PR #369, opened 2020-01-16
Update years and team name.

ceball:
Just happened to notice the license is out of date. Do we really have to keep updating it? (I mean the copyright years - not just because we keep changing the team name 😂.)

(I also tried to make the third paragraph independent of the team name based on the text at https://opensource.org/licenses/BSD-3-Clause. But could we drop the third paragraph entirely and move to https://opensource.org/licenses/BSD-2-Clause?)

Also: HoloViz team? HoloViz? HoloViz developers (https://github.com/orgs/holoviz/teams/holoviz-dev)? Param developers? Param contributors? Etc

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (c) 2005-2018, IOAM (ioam.github.com)
+Copyright (c) 2005-2020, HoloViz team.
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
@@ -13,9 +13,9 @@ met:
    documentation and/or other materials provided with the
    distribution.
 
- * Neither the name of IOAM nor the names of its contributors
-   may be used to endorse or promote products derived from this
-   software without specific prior written permission.
+ * Neither the name of the copyright holder nor the names of any
+   contributors may be used to endorse or promote products derived
+   from this software without specific prior written permission.
 
 THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
 "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT


jbednar:
Thanks! This is timely; Jean-Luc and I were just discussing this and it's currently on his plate to update this info across all our sites. 

We looked around for guidance on the years, and found people strongly advocating for every opinion possible:

1. Don't bother with a copyright notice; it's based on the lifetime of the author now, not any explicit copyright claim, so what's the point?
2. Put the date the material was first created, to show that your copy (should someone else duplicate it) is the original.
3. Put the date the material was last updated, or else your readers will think your site is out of date.
4. Don't try to balance 2 and 3 by putting a range; a range doesn't mean anything in copyright law!

He and I decided to go for a range from the first time the material was created up to the last time it was updated, reasoning that item 1 indicates that nothing we do matters for copyright law (thus invalidating objection 4), and we want to establish both 2 and 3 (i.e. that the material is actually solid and has been around a while, while also not being stale, and is in fact currently updated).  You could argue us into another position, but otherwise, 2005-2020 is correct for Param.

Thanks for changing the license text as indicated; that's an improvement. I don't see any reason to change BSD-3 to BSD-2 now; I think it's always been BSD-3 since the move to Github and seems more likely to cause confusion and concern to change it than it is to achieve anything useful, even though strictly BSD-2 is more permissive than BSD-3. 

As for the team name, I think it should be the same across all HoloViz sites just so we don't have to think about it.  I don't want to claim there is a Param team, but there is definitely a HoloViz team, i.e. the members of the HoloViz organization on Github.  

Between HoloViz developers, HoloViz team, and HoloViz contributors, I think I favor HoloViz team, because developers and contributors both indicate that the sum total of all contributors controls the copyright, which I don't think is appropriate; really it's controlled by the team that is in charge, which is fluid but can't immediately be hijacked by simply submitting one PR. Any individual contributor relinquishes the copyright to the team, and the team is then the one to decide e.g. a change in licensing, without requiring unanimous consent of all contributors (which would be infeasible!).

ceball:
That's all fine with me. Thanks for thinking about it! And for explaining it.

I think the change matches what you say, but I'll leave this for Jean-Luc/you to fix up/merge when you are doing the others.

(Maybe with another 15 years, we'll be able to finish this thing? 🤣 )

jbednar:
Thanks. I'll let @jlstevens merge it, to remind him to make similar updates to the licensing text on other sites.

philippjfr:
I'm going to merge since I want to tag. Hopefully @jlstevens sees this anyway.

PR #378, opened 2020-02-18
Fixes #182

### LINKED ISSUE #182: Check spellings
I noticed 'suprising' just now which is one of my standard typos. One day we should just run the whole project through a spell checker...

scottwedge:


ceball:
Thanks!

(However, I just looked, and I can immediately see other spelling errors too, e.g. 'instatiate". I think it's a lost cause...)

PR #383, opened 2020-04-07
Removing the Support section from the Param home page

kebowen730:
Small PR removing the support section

### DIFF
--- a/examples/index.ipynb
+++ b/examples/index.ipynb
@@ -281,11 +281,6 @@
     "Recent release notes are available on [GitHub](https://github.com/ioam/param/releases).\n",
     "\n",
     "For older releases, see our [historical release notes](historical_release_notes.html).\n",
-    "\n",
-    "\n",
-    "# Support\n",
-    "\n",
-    "Questions and comments are welcome at https://github.com/ioam/param/issues."
    ]
   }
  ],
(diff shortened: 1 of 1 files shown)


jlstevens:
Looks good to me! Happy to merge when the tests complete.

examples/index.ipynb
```diff
@@ -281,11 +281,6 @@
     "Recent release notes are available on [GitHub](https://github.com/ioam/param/releases).\n",
     "\n",
     "For older releases, see our [historical release notes](historical_release_notes.html).\n",
```
jbednar, line 283:
```suggestion
    "For older releases, see our [historical release notes](historical_release_notes.html).\n"
```

(I think JSON won't allow the trailing comma. In general it's important to load a .ipynb file back into Jupyter any time it is edited directly like this; it's too easy to break the syntax otherwise!)

jlstevens: Yes, I am not sure that trailing comma is valid. Safest to remove it.

jlstevens:
@kebowen730 Can you confirm the notebook still opens ok in classic notebook/jupyterlab? It should be ok but it is worth checking.

If it is working then I'll go ahead and hit merge.

kebowen730:
@jbednar I ran it locally and everything still seems to be working correctly

PR #384, opened 2020-04-08
Convert param accessor to weakref

philippjfr:
Avoids creating circular reference between Parameterized and Parameters objects.


philippjfr:
I think using a property is probably safer and easier tbh, will push that up tomorrow.

philippjfr:
weakrefs won't pickle so a property definitely seems preferable.

PR #386, opened 2020-04-09
Turn .param into a property

philippjfr:
Supersedes https://github.com/holoviz/param/pull/384

The reasoning behind this is that having accessors which hold a reference back to the Parameterized object create a circular reference which stops CPython's garbage collection from immediately collecting a Parameterized instance when it is no longer referenced by any external object instead having to wait for the next `gc.collect()` cycle.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1109,6 +1109,16 @@ class Parameters(object):
     def self_or_cls(self_):
         return self_.cls if self_.self is None else self_.self
 
+    def __setstate__(self, state):
+        # Set old parameters state on Parameterized._parameters_state
+        self_or_cls = state.get('self', state.get('cls'))
+        for k in self_or_cls._parameters_state:
+            key = '_'+k
+            if key in state:
+                self_or_cls._parameters_state[k] = state.pop(key)
+        for k, v in state.items():
+            setattr(self, k, v)
+            
     def __getitem__(self_, key):
         """
         Returns the class or instance parameter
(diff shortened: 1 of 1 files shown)


param/parameterized.py
```diff
@@ -1072,16 +1072,43 @@ def __init__(self_, cls, self=None):
...
         self_.cls = cls
         self_.self = self
-        self_._BATCH_WATCH = False  # If true, Event and watcher objects are queued.
-        self_._TRIGGER = False
-        self_._events = []         # Queue of batched eventd
-        self_._watchers = []         # Queue of batched watchers
+
+    @property
+    def _BATCH_WATCH(self_):
+        return self_.self_or_cls._parameters_state['BATCH_WATCH']
```
philippjfr, line 1078: To allow `.param` to be a property any state that was previously stored on `Parameters` now needs to be stored on the Parameterized itself. I didn't want to rock the boat too much so for now I've simply proxied the state that is now on the Parameterized class or instance using these properties. If we want to refactor more then all references to `parameterized.param._BATCH_WATCH`, `parameterized.param._TRIGGER`, `parameterized.param._events` and `parameterized.param._watchers` could be updated to refer to the `parameterized._parameters_state` instead.

jlstevens:
Looks good. If pickling is working we'll want to merge ASAP to test!

philippjfr:
It is indeed, @jbednar should we merge to start testing?

philippjfr:
I'll merge and tag soon so we can start testing. Can always revert if we find some horrendous regression.

PR #390, opened 2020-04-10
update the html theme to use sphinx_holoviz_theme

kebowen730:
@philippjfr @jlstevens I updated the conf.py file to use sphinx_holoviz_theme

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,7 +128,7 @@ jobs:
       before_install:
         - pip install graphviz
       install:
-        - pip install nbsite sphinx_ioam_theme "tornado<6"
+        - pip install nbsite sphinx_holoviz_theme "tornado<6"
         - pip install -e .
       script:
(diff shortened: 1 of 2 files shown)


ceball:
I'm totally no longer competent to review this stuff, but what about https://github.com/holoviz/param/blob/master/.travis.yml#L131 - does it need to be updated too?

(Such dependencies should be in setup.py where people might have a chance of finding them...)

ceball:
@kebowen730 Like I said, I'm not sure how this all works any more, but I can't immediately think of what else you'd need to change, so I'll merge. Could you then try pushing the appropriate tag to get a dev website re-build to happen (i.e. to update pyviz-dev.github.io/param) and check the build works and the site comes out ok?

I don't know if this general pattern (i.e. how to rebuild our sites) is documented somewhere. If not, it would be great to capture it somewhere.

Thanks!

PR #401, opened 2020-04-16
Batch param.depends

philippjfr:
The `param.depends` decorator was previously registering a separate watch callback for each parameter it was watching. This PR ensures that `param.depends` registers a single callback for each parameterized object being depended on.


jbednar:
Makes sense to me.

PR #402, opened 2020-04-22
Fix broken link in ipython %params message

maximlt:
I tried the ipython magics %params and saw that broken link: http://ioam.github.io/param/


philippjfr:
Great, thanks!

Issue #381, opened 2020-02-29
xarray Dataset and other data science oriented parameters


poplarShift:
I've been enjoying doing my data science pipelines with panel+param lately which makes workflows tidy and parameter sweeps a breeze. There already is a `param.DataFrame`, how about e.g. `Dataset` which would hold an `xarray.Dataset`?

jbednar:
Sounds great!  And presumably xarray.DataArray?  The main work is just to figure out how to validate what's given (i.e. to express constraints on the dimensions/coordinates/types/etc), starting by copying param.DataFrame and adapting as needed. (If not for validation, of course, you can already use param.Parameter.)  If you're interested in starting a PR, we would love to have it!  And then maybe eventually we could adapt xrviz to display the Dataset or DataArray as a Panel.

ceball:
In case it's useful (haven't investigated yet): traitlets related to numpy, pandas, and xarray: https://github.com/jupyter-widgets/traittypes/

poplarShift:
traittypes looks interesting but doesn't seem to be checking a lot of the properties (dimensions, coordinates, ...) so I have not tried to use that for inspiration.

> And then maybe eventually we could adapt xrviz to display the Dataset or DataArray as a Panel.

The new xarray html repr would be a great starting point for linking it to a Panel widget.

PR #404, opened 2020-04-29
Draft xarray parameters

poplarShift:
Implements #381

I'm just putting this here for discussion. The PR would need some tests I imagine.

Questions:

1. Should we validate the provided default? Currently, param is e.g. ok with `param.Number(default=None, bounds=(1, 100))` but will throw an error at `param.DataFrame(default=None, rows=(1, 10))`.

2. I've tried to carry over (from DataFrames) the distinction between passing sets or lists into the arguments. This might get confusing though... Opinions?

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1534,6 +1534,120 @@ class Series(ClassSelector):
             self._length_bounds_check(self.rows, len(val), 'Row')
 
 
+class _XrBase:
+    """
+    Provides mixin methods for DataArray and Dataset. Not useful on its own.
+    """
+    __slots__ = []
+
+    def _validate_dims_coords(self, val):
+        if self.dims is not None:
+            self._validate_property(self.dims, val.dims, 'dimensions')
+
+        if self.coords is not None:
+            self._validate_property(self.coords, list(val.coords),
+                                           'coordinates')
+            if isinstance(self.coords, dict):
+                self._coord_check(self.coords, val.coords)
+
+
+    def _validate_property(self, expected, property, name):
+        # if isinstance(expected, (list, set)):
+        difference = set(expected) - set([str(el) for el in property])
+        if difference:
+            msg = ('Provided DataArray {name} {found} do not contain '
+                   'required {name} {expected}')
+            raise ValueError(msg.format(
+                found=list(property),
+                expected=sorted(expected),
+                name=name,
+            ))
+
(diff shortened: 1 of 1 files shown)


param/__init__.py
```diff
@@ -1534,6 +1534,125 @@ def _validate(self, val):
...
+            self._validate_property(self.coords, list(val.coords),
+                                           'coordinates')
+            if isinstance(self.coords, dict):
+                self._coord_check(self.coords, val.coords)
+
+
+    def _validate_property(self, expected, property, name):
+        difference = set(expected) - set([str(el) for el in property])
+        if difference:
+            msg = ('Provided DataArray {name} {found} do not contain '
```
jbednar, line 1557:
```suggestion
            msg = ('Provided DataArray {name} {found} does not contain '
```

jbednar:
>1. Should we validate the provided default? Currently, param is e.g. ok with param.Number(default=None, bounds=(1, 100)) but will throw an error at param.DataFrame(default=None, rows=(1, 10)).

Param usually sets `allow_None=True` if the default is set to None, under the assumption that the Parameterized class author clearly thus intends None to be valid for this Parameter.  If param.DataFrame does not do that, I would be happy for such code to be added to param.DataFrame (and to these new classes) to do that; it seems like a reasonable thing to do for the convenience of authors.

> 2. I've tried to carry over (from DataFrames) the distinction between passing sets or lists into the arguments. This might get confusing though... Opinions?

I agree that's confusing, and completely did not remember that param.DataFrame acted that way. It's important to be able to make that distinction, and sets vs. list/dict does concisely convey that, but it's mysterious.  It might be nice to have a more explicit way of conveying the same information, but I can't easily think of any non-clunky way to do it.

param/__init__.py
poplarShift: The argument I passed for `name` are all plural ("columns", "data variables", ...)

poplarShift:
PS: Note that a `list` columns argument to DataFrame will also check for column _order_. I don't think that makes much sense for a DataArray, so I would drop that distinction here.

jbednar:
> Then how about `set` is for minimal requirements, `list` is for an exact match, and `tuple` is not allowed?

Sounds reasonable to me.

maximlt:
Just noting that this will need some tests and documentation before being merged.

Issue #410, opened 2020-05-14
Subobject dependencies should be dynamic


philippjfr:
The `param.depends` decorator supports expressing dependencies on parameters on subobjects. However these dependencies are evaluated only on initialization meaning that if the object is not yet defined then it will error and if it is redefined later it will not set up that dependency. The following example should work but currently errors:

```python
import param

class A(param.Parameterized):
    
    value = param.Parameter()
    
class B(param.Parameterized):
    
    a = param.ObjectSelector()
    
    @param.depends('a.value', watch=True)
    def test_depends(self):
        print(self.a.value)

b = B()
b.a = A()
b.a.value = 1
```

jbednar:
I guess whether this is a bug or a requested feature depends on your outlook on life. :-)  Seems like you're coming down  on the side of "possible feature" given that it wasn't part of the original design, but I agree that it _should_  behave properly in that case, making it also kinda be a bug...

philippjfr:
Agreed, I clicked on feature request but then almost decided to change the label.

ceball:
On Thu, May 14, 2020 at 8:54 PM James A. Bednar <notifications@github.com>
wrote:

> I guess whether this is a bug or a requested feature depends on your
> outlook on life. :-)
>
If it's code you wrote, someone's suggesting a feature enhancement. If it's
code someone else wrote, it's a bug report! :)

PR #411, opened 2020-05-14
[WIP] Dynamically reevaluate dependencies on subobjects

philippjfr:
Implements https://github.com/holoviz/param/issues/410


param/parameterized.py
```diff
@@ -2407,9 +2417,15 @@ def __init__(self, **params):
                 # 'dependers'.
                 for p in self.param.params_depended_on(n):
                     # TODO: can't remember why not just pass m (rather than _m_caller) here
-                    (p.inst or p.cls).param.watch(_m_caller(self, n), p.name, p.what, queued=queued)
-
-        self.initialized = True
+                    key = (n, p.name, p.what)
```
philippjfr, line 2420: Need to make sure this key is sufficient to uniquely identify a callback.

param/parameterized.py
```diff
@@ -864,6 +864,8 @@ def __set__(self,obj,val):
                 obj.__dict__[self._internal_name] = val
 
         self._post_setter(obj, val)
+        if hasattr(val, '_setup_dependencies'):
```
philippjfr, line 867: Ideally would have a better condition here, but can't do `isinstance(val, Parameterized)` since this is evaluated before Parameterized exists.

jbednar: Makes sense.

tonyfast:
@philippjfr can you merge or close this pr.

Issue #382, opened 2020-04-06
Literal serialization and deserialization


jlstevens:
The effort to serialize/deserialize parameterized objects is being restarted (see #159 and #153)  with the goal of controlling param state via REST APIs. As can be seen above, a lot of thought went into how much can be expressed in YAML in particular to support as much of param as possible.

The current focus on REST means that the immediate goal is to get JSON support working as that is the format most typically used for data exchange over the web. As JSON is less rich that YAML, I am now thinking about what can be achieved without using any YAML specific features. In particular, I am thinking about serializing the parameter values *only* (and not things like bounds, objects or other `Parameter` state).

Lastly note that this focus on JSON doesn't mean YAML can't be supported as long as only the subject of YAML is used that is supported by JSON.
I believe that a serialization approach based on what JSON can handle (i.e lists, mappings, numbers, strings, bools...) can still be very useful even if things like callables aren't supported. However, things like the `NumericTuple` parameter is very common in param code but tuples are not supported by JSON. Accepting this limitation would be very annoying so I propose that we serialize into what JSON supports (e.g a tuple would serialize into a JSON list) and deserialize by noting that a list can be easily cast into the tuple that the parameter requires. Here is how I think we can handle most parameters with JSON:

**Dynamic** Dynamic parameters have a current state that can often be serialized but also a callable that generates that state that can't be. Only the current state will be serialized.

**Number** Not a problem to accept integers and floats. Note that NaN is normally encoded as null in JSON.

**Integer** As above.

**Magnitude** As above where the value is expected between zero and one.

**Boolean** Not a problem

**Tuple** JSON lists are cast to tuples.

**NumericTuple** As above.

**XYCoordinates** As above

**Callable** Not supported

**Action** Not supported

**Composite** Not supported but also rarely used.

**ObjectSelector** Supports any other supported type.

**Selector** Supports any other supported type.

**ClassSelector** First tricky case that isn't callable. Classes could be encoded as appropriately qualified strings (i.e module/class name).

**List** Supported directly in JSON.

**HookList** Not supported.

**Dict** Supported as JSON mapping.

**Array** Could be supported e.g using list of list format in numpy array constructor.

**DataFrame** Can be supported using appropriate literal constructor (e.g dict of lists).

**Series** As above

**Path** String with validation.

**Filename** String with validation.

**Foldername** String with validation.

**FileSelector** String with validation.

**ListSelector** Value is a list (i.e supported)

**MultiFileSelector** As above.

[...]

PR #414, opened 2020-06-10
JSON serialization and schema generation

jlstevens:
This PR outlines another attempt at serializing/deserializing to JSON (for reference see
#153 for earlier attempts at tackling this problem). The design used in this PR aims to try to keep things simpler but also adds the ability to generate [JSON Schemas](https://json-schema.org/) which should help with testing as well as specification of allowable parameter values (e.g this could be used to specify suitable widgets in panel).

This PR is based off the ideas suggested in #382.

### TODO:

- [ ] The main open problem is about finding the appropriate API and JSON (and schema) representation for entire parameterized objects (i.e including the class identity as well as all the parameters).

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1456,7 +1456,7 @@ class DataFrame(ClassSelector):
         self.rows = rows
         self.columns = columns
         self.ordered = ordered
-        super(DataFrame,self).__init__(pdDFrame, default=default, allow_None=True, **params)
+        super(DataFrame,self).__init__(pdDFrame, default=default, **params)
         self._validate(self.default)
 
 
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1014,7 +1014,7 @@ class Tuple(Parameter):
                              (self.name,len(val),self.length))
 
 
-    def _deserialize(self, value):
+    def deserialize(self, value):
         return tuple(value) # As JSON has no tuple representation
 
 
@@ -1841,7 +1841,7 @@ class Date(Number):
 
         self._checkBounds(val)
 
-    def _deserialize(self, value):
+    def deserialize(self, value):
         return dt.datetime.fromisoformat(value)
 
 class CalendarDate(Number):
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1853,7 +1853,7 @@ class Date(Number):
 
     @classmethod
     def serialize(cls, value):
-        return value.replace(microsecond=0).isoformat()
+        return value.isoformat()
     @classmethod
     def deserialize(cls, value):
         return dt.datetime.fromisoformat(value)
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1884,6 +1884,14 @@ class CalendarDate(Number):
 
         self._checkBounds(val)
 
+    @classmethod
+    def serialize(cls, value):
+        return value.strftime("%Y-%m-%d")
+
+    @classmethod
+    def deserialize(cls, value):
+        return dt.datetime.strptime(value, "%Y-%m-%d").date()
+
 
 class Color(Parameter):
     """
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1853,6 +1853,8 @@ class Date(Number):
 
     @classmethod
     def serialize(cls, value):
+        if not isinstance(value, (dt.datetime, dt.date)): # i.e np.datetime64
+            value = value.astype(dt.datetime)
         return value.strftime("%Y-%m-%dT%H:%M:%S.%f")
 
     @classmethod
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1430,6 +1430,15 @@ class Array(ClassSelector):
         from numpy import ndarray
         super(Array,self).__init__(ndarray, allow_None=True, default=default, **params)
 
+    @classmethod
+    def serialize(cls, value):
+        return value.tolist()
+
+    @classmethod
+    def deserialize(cls, value):
+        from numpy import array
+        return array(value)
+
 
 class DataFrame(ClassSelector):
     """
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1853,7 +1853,7 @@ class Date(Number):
 
     @classmethod
     def serialize(cls, value):
-        return value.isoformat()
+        return value.strftime("%Y-%m-%dT%H:%M:%S.%f")
     @classmethod
     def deserialize(cls, value):
         return dt.datetime.fromisoformat(value)
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1014,7 +1014,8 @@ class Tuple(Parameter):
                              (self.name,len(val),self.length))
 
 
-    def deserialize(self, value):
+    @classmethod
+    def deserialize(cls, value):
         return tuple(value) # As JSON has no tuple representation
 
 
@@ -1841,7 +1842,8 @@ class Date(Number):
 
         self._checkBounds(val)
 
-    def deserialize(self, value):
+    @classmethod
+    def deserialize(cls, value):
         return dt.datetime.fromisoformat(value)
 
 class CalendarDate(Number):
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1854,9 +1854,10 @@ class Date(Number):
     @classmethod
     def serialize(cls, value):
         return value.strftime("%Y-%m-%dT%H:%M:%S.%f")
+
     @classmethod
     def deserialize(cls, value):
-        return dt.datetime.fromisoformat(value)
+        return datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f")
 
 class CalendarDate(Number):
     """
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1014,6 +1014,9 @@ class Tuple(Parameter):
                              (self.name,len(val),self.length))
 
 
+    def _deserialize(self, value):
+        return tuple(value) # As JSON has no tuple representation
+
 
 class NumericTuple(Tuple):
     """A numeric tuple Parameter (e.g. (4.5,7.6,3)) with a fixed tuple length."""
@@ -1838,6 +1841,8 @@ class Date(Number):
 
         self._checkBounds(val)
 
+    def _deserialize(self, value):
+        return dt.datetime.fromisoformat(value)
 
 class CalendarDate(Number):
     """
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1015,7 +1015,7 @@ class Tuple(Parameter):
 
 
     @classmethod
-    def deserialize(cls, value):
+    def ir_deserialize(cls, value):
         return tuple(value) # As JSON has no tuple representation
 
 
@@ -1431,11 +1431,11 @@ class Array(ClassSelector):
         super(Array,self).__init__(ndarray, allow_None=True, default=default, **params)
 
     @classmethod
-    def serialize(cls, value):
+    def ir_serialize(cls, value):
         return value.tolist()
 
     @classmethod
-    def deserialize(cls, value):
+    def ir_deserialize(cls, value):
         from numpy import array
         return array(value)
 
@@ -1511,11 +1511,11 @@ class DataFrame(ClassSelector):
             self._length_bounds_check(self.rows, len(val), 'Row')
 
     @classmethod
-    def serialize(cls, value):
+    def ir_serialize(cls, value):
         return value.to_dict('records')
 
     @classmethod
-    def deserialize(cls, value):
+    def ir_deserialize(cls, value):
         from pandas import DataFrame as pdDFrame
         return pdDFrame(value)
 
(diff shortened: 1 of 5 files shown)


param/serializer.py
```diff
@@ -0,0 +1,178 @@
...
+                {"type": "number"}, {"type": "number"}],
+            "additionalItems": False
+        }
+
+    @classmethod
+    def range_schema(cls, p, safe=False):
+        # Extend for allow None
+        return  {
+            "type": "array",
+            "items": [{"type": "number"},{"type": "number"}],
```
philippjfr, line 165: This should handle the bounds on the range.

param/serializer.py
```diff
@@ -0,0 +1,178 @@
...
+        return JSONNullable(schema) if p.allow_None else schema
+
+    @classmethod
+    def integer_schema(cls, p, safe=False):
+        return cls.number_schema(p)
+
+    @classmethod
+    def numerictuple_schema(cls, p, safe=False):
+        return {"type": "array",
+                "additionalItems": { "type": "number" }}
```
philippjfr, line 149: This should encode the length if provided.

param/serializer.py
```diff
@@ -0,0 +1,178 @@
...
+
+    # Custom Schemas
+
+    @classmethod
+    def date_schema(cls, p, safe=False):
+        return { "type": "string", "format": "date-time"}
+
+    @classmethod
+    def tuple_schema(cls, p, safe=False):
+        return { "type": "array"}
```
philippjfr, line 126: This should encode the length if provided, i.e. probably using:

```
  "minItems": 2,
  "maxItems": 3
```

param/serializer.py
```diff
@@ -0,0 +1,178 @@
...
+class JSONSerialization(Serialization):
+    """
+    Class responsible for specifying JSON serialization, deserialization
+    and JSON schemas for Parameters and Parameterized classes and
+    objects.
+    """
+
+    unserializable_parameter_types = ['Callable']
+
+    json_schema_literal_types = {int:'integer', float:'number', str:'string'}
```
philippjfr, line 45: param.String schema should include the optional regex attribute: https://json-schema.org/understanding-json-schema/reference/regular_expressions.html

philippjfr:
Can you also make sure you handle ObjectSelector/Selector with enums in your initial PR?

philippjfr:
Also encode the `label` attribute as the 'title' property in the schema.

philippjfr:
Oh also ListSelector should be `{'type': ['array'], 'items': {'enum': [...]}}`

jlstevens:
All good suggestions, thanks! I won't implement the regexp attribute for strings though as there is no easy/meaningful way to try to convert Python style regular expressions to equivalent javascript.

param/serializer.py
jlstevens: Addressed in c353f7b

param/serializer.py
jlstevens: Addressed in c353f7b

param/serializer.py
jlstevens: Addressed in c353f7b

param/serializer.py
jlstevens: Would be nice but JavaScript and Python regular expressions are not compatible enough for this to be worth attempting.

jlstevens:
Still more work to be done, but the suggested refactor and all of Philipp's suggestions are now implemented.

param/__init__.py
```diff
@@ -1014,6 +1014,9 @@ def _validate(self, val):
                              (self.name,len(val),self.length))
 
 
+    def _deserialize(self, value):
```
philippjfr, line 1017: Can we make serialize/deserialize public methods? I will be using them heavily. Also can they be classmethods?

jlstevens: Applied this suggestion in 56d5089 and edadd08 though it is worth nothing the names aren't really right as neither serialize nor deserialize actually work with strings (i.e the serialized form).

`serialize_python_value` and `deserialize_python_value` would be more accurate names but also ugly.

philippjfr: Bokeh calls it `transform` and `from_json` (even though it's already a Python object by that point). The former would work but the latter is clearly not appropriate, they also aren't symmetric like `serialize` and `deserialize` so I'm okay keeping the names.

param/__init__.py
```diff
@@ -1838,6 +1842,9 @@ def _validate(self, val):
 
         self._checkBounds(val)
 
+    @classmethod
+    def deserialize(cls, value):
+        return dt.datetime.fromisoformat(value)
```
philippjfr, line 1847: Confused by this, either there should be a symmetric serialize method that outputs the iso formatted string or this should happen in the JSON decoder, which seems like the better solution.

philippjfr: Serialize and deserialize should always support a round-trip:

```python
p = parameterized.param.value
parameterized.value = p.deserialize(p.serialize(parameterized.value))
```

jlstevens: Round tripping will be tested but unfortunately Python's JSON `Encoder` and `Decoder` are *not* symmetric if you check the docs (see the example for encoding/decoding a complex number).

What I can do is move the logic out the encoder though which will make it symmetric for datetimes at least.

param/__init__.py
jbednar: ir_serialize and ir_deserialize?

param/__init__.py
jbednar: Seems like we should attempt to support round tripping as best we can, to preserve our sanity and to support lots of potential workflows that depend on it.  So the fact that the classes aren't symmetric is annoying, but our syntax should be symmetric if at all possible, at least when using a schema during decoding. (Without a schema there's no hope, as e.g. we don't know whether "0.9" is meant to be a Magnitude or a Number...)

jbednar:
>  I won't implement the regexp attribute for strings though as there is no easy/meaningful way to try to convert Python style regular expressions to equivalent javascript.

I'm not sure I agree here; I don't think we need JS to process the regexp attribute for Param.String. That attribute expresses a Python regexp by which we can validate the parameter value in Python, and it seems fine to encode that into a schema whether or not JS can actually use it.  Can't we encode it into the schema, send it over as JSON, unpack it in Python, and apply the test in Python (or construct an appropriate Parameter that will then apply it) whether or not JS can do anything with the regex? JSON is just a transport mechanism for this metadata, and JS doesn't have to process it necessarily.

That said, there are Python libraries for working with JS-flavor regexes (https://pypi.org/project/js-regex/), and there may be some equivalent JS library for working with Python-flavor regexes, for anyone who did need to do the validation in JS...

param/parameterized.py
```diff
@@ -10,6 +10,7 @@
 import random
 import numbers
 import operator
+from .serializer import JSONSerialization
```
jbednar, line 13: This one import breaks parameterized.py's longstanding status as a file that can be used standalone, with no other files (not even `__init__.py`).  The original idea was that people might want to make their own set of specific Parameter classes to use with Parameterized, e.g. something much smaller than the full list that we provide, and we wanted people to be able to do that and to work with Parameterized in general without necessarily bringing in the specific Parameter classes beyond the one actually defined in Parameterized (param.String).  We aren't using that capability ourselves, so losing that won't hurt us, but it's been a feature of parameterized.py for so many years that I'm reluctant to lose it now.

```suggestion
# Allow this file to be used standalone if desired, albeit without JSON serialization
try: 
   from .serializer import JSONSerialization
except ImportError:
   JSONSerialization = None
```

param/__init__.py
philippjfr: Just to be very clear here, my requirement here is that I expect to use serialize/deserialize for other purposes, e.g. for URL parameter encoding and decoding. URL parameters are going to be much more limited than JSON so it will only work for a subset of types but I want to ensure that I can round-trip the few types that will be supported in a straightforward way and my hope that way is simply:

- serialize: `html.escape(Parameter.serialize(value))`
- deserialize `Parameter.deserialize(html.unescape(value))`

jlstevens:
> That attribute expresses a Python regexp by which we can validate the parameter value in Python, and it seems fine to encode that into a schema whether or not JS can actually use it. Can't we encode it into the schema, send it over as JSON, unpack it in Python, and apply the test in Python (or construct an appropriate Parameter that will then apply it) whether or not JS can do anything with the regex?

We *can* but not without extending JSON schema with our own semantics which we ought to avoid as it makes things non-standard. The point here is that JSONSchema *looks* like it has the same concept as param, but we can't make use of it as the regular expressions will be JS and not Python.

We can definitely come up with our own param specific schema later but for now the goal is to match the semantics of param and JSON schema *where they already overlap* and without having to define our own custom schemas.

param/parameterized.py
jlstevens: Sure, happy to make this serialization file optional.

jlstevens: Implemented in c1c3978

param/__init__.py
jlstevens: I think with the IR approach, the `serialize` and `deserialize` methods should always be symmetric as you say. Only when we need a custom `JSONEncoder` does it get tricky (and so far it has been possible to avoid).

param/__init__.py
```diff
@@ -1497,6 +1501,15 @@ def _validate(self, val):
...
             self._length_bounds_check(self.rows, len(val), 'Row')
 
+    @classmethod
+    def serialize(cls, value):
+        return value.to_dict('records')
+
+    @classmethod
+    def deserialize(cls, value):
+        from pandas import DataFrame as pdDFrame
+        return pdDFrame(value)
```
philippjfr, line 1511: This may not be sufficient to fully reconstruct things since:

a) The column order isn't encoded
b) The index columns can't be reconstructed this way

I wonder if it would be helpful if `deserialize` could optionally accept the schema to help it reconstruct the original data better (it may not help in this particular case though).

jlstevens: Not sure what to do about the index but I do know that the JSON schema for the row-oriented data can't be relied on to supply the column order as the columns are declared in JSON objects/Python dictionaries. Python dictionaries *could* encode order information assuming the order of JSON parsing is preserved but I'm not sure this is behavior we should rely on.

philippjfr: The JSON spec says not to rely on property ordering https://github.com/json-schema-org/json-schema-spec/issues/571

As that issue suggests we may look at extensions to json-schema such as json-schema-form.

jlstevens: There seems to have been discussion about a `propertyOrder` specification but it doesn't seem to have gone anywhere https://github.com/json-schema/json-schema/issues/119

jbednar:
> We can but not without extending JSON schema with our own semantics which we ought to avoid as it makes things non-standard. 

Agreed in general.  But if there is a way to express that information in a way that will safely be ignored by a JSON schema validator, then it's better to express it than simply lose it; being able to apply tighter checking in at least some cases seems better than not being able to at all, particularly when in practice we'll be using the Python-based validation in nearly all cases. I don't actually know if there _is_ a handy place one can stuff that information safely, so the point may be moot.

tests/API1/testjsonserialization.py
```diff
@@ -0,0 +1,125 @@
...
+        self.assertEqual({'a':
+                          {'type': 'integer', 'minimum': 2, 'exclusiveMaximum': 30,
+                           'description': 'Example doc', 'title': 'A'}},
+                         param_schema)
+
+        exception = "1 is not of type 'object'"
+        with self.assertRaisesRegexp(ValidationError, exception):
+            validate(instance=1, schema=schema)
+
+
```
philippjfr, line 77: Nitpicky (and I know this isn't quite ready) but can we standardize on one newline between methods like the rest of param?

philippjfr:
Coould you also tick off the types that are supported in my list above?

tests/API1/testjsonserialization.py
jlstevens: Addressed in one of the following commits.

param/__init__.py
jlstevens: Used this suggestion in a7a4449

param/__init__.py
```diff
@@ -1014,6 +1014,10 @@ def _validate(self, val):
                              (self.name,len(val),self.length))
 
 
+    @classmethod
+    def ir_deserialize(cls, value):
```
philippjfr, line 1018: My preference would be to keep these as `serialize` and `deserialize` to keep the API simple.

jlstevens: I have no real preference here and I am happy to revert this change. @jbednar ?

jlstevens:
@jbednar @philippjfr Coverage is very slightly down (-0.005%) but otherwise tests are passing. Other than the `ir_serialize` versus `serialize` renaming, I think this might be a good place to merge. The subsequent PR I am planning would address serialization of entire parameterized objects but I think this first step of serializing parameters (with schemas) is already valuable.

param/__init__.py
jbednar: No preference.

jlstevens: Ok, in that case Philipp wins the vote and I'll revert!

jlstevens: Done in 191d431

param/__init__.py
```diff
@@ -1426,6 +1430,15 @@ def __init__(self, default=None, **params):
...
         super(Array,self).__init__(ndarray, allow_None=True, default=default, **params)
 
+    @classmethod
+    def serialize(cls, value):
+        return value.tolist()
+
+    @classmethod
+    def deserialize(cls, value):
+        from numpy import array
+        return array(value)
```
philippjfr, line 1440: Not super important but it would be nice if you could pass in an array without making a copy.

```suggestion
        from numpy import asarray
        return asarray(value)
```

jlstevens: Good suggestion, thanks.

jlstevens:
@philippjfr @jbednar I think this PR is now in a useful state to merge. I'll go ahead and do that shortly unless you have any other suggestions or objections..

philippjfr:
Looks good. Merge away.

PR #421, opened 2020-07-27
Invoke ansi formatting when a IPython shell exists

### LINKED ISSUE #115: IPython extension causes confusing help text?
If I'm using ipython and I run ``%load_ext param.ipython`` everything seems good. But if I'm not using ipython, or if I use ipython without loading the ipython extension, I think the help text for a parameterized instance is confusing.

Normal python:

```python
>>> import param
>>> class X(param.Parameterized):
...     y = param.Parameter(default=1)
... 
>>> x = X(y=10)
>>> help(x)
```

Gives me:
```
Help on X in module __main__ object:

class X(param.parameterized.Parameterized)
 |  params(y=Parameter, name=String)
 |  ESC[1;32mParameters of 'X'
 |  =================
 |  ESC[0m
 |  ESC[1;31mParameters changed from their default values are marked in red.ESC[0m
 |  ESC[1;36mSoft bound values are marked in cyan.ESC[0m
 |  C/V= Constant/Variable, RO/RW = ReadOnly/ReadWrite, AN=Allow None
 |  
 |  ESC[1;34mNameValue    Type    Mode ESC[0m
 |  
 |  y   1  Parameter  V RW 
 |  
 |  ESC[1;32mParameter docstrings:
 |  =====================ESC[0m
 |  
 |  ESC[1;34my: < No docstring available >ESC[0m
 |  
...
```

Maybe my terminal's not very clever, so it can't display colors, but you can see that there was no attempt to color the ``y`` line, and the value shown is the default rather than the actual value
ipython
.

Additionally, the attached screenshot shows the result from within ipython (without having run ``%load_ext param.ipython``).

tonyfast:
fixes #115

### DIFF
--- a/param/ipython.py
+++ b/param/ipython.py
@@ -58,7 +58,7 @@ class ParamPager(object):
     def get_param_info(self, obj, include_super=True):
         """
         Get the parameter dictionary, the list of modifed parameters
-        and the dictionary or parameter values. If include_s'https://tools.ietf.org/html/rfc7946'uper is
+        and the dictionary or parameter values. If include_super is
         True, parameters are also collected from the super classes.
         """
 
--- a/param/ipython.py
+++ b/param/ipython.py
@@ -22,17 +22,20 @@ __author__ = "Jean-Luc Stevens"
 import re
 import textwrap
 import param
+import sys
 
 
 # Whether to generate warnings when misformatted docstrings are found
 WARN_MISFORMATTED_DOCSTRINGS = False
 
-# ANSI color codes for the IPython pager
-red   = '\x1b[1;31m%s\x1b[0m'
-blue  = '\x1b[1;34m%s\x1b[0m'
-green = '\x1b[1;32m%s\x1b[0m'
-cyan = '\x1b[1;36m%s\x1b[0m'
-
+if "IPython" in sys.modules and sys.modules["IPython"].get_ipython():
+    # ANSI color codes for the IPython pager
+    red   = '\x1b[1;31m%s\x1b[0m'
+    blue  = '\x1b[1;34m%s\x1b[0m'
+    green = '\x1b[1;32m%s\x1b[0m'
+    cyan = '\x1b[1;36m%s\x1b[0m'
+else:
+    red = blue = green = cyan = '%s'
 
 
 class ParamPager(object):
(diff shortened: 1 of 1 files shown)


jlstevens:
Looks great though something weird happened in the docstring.

param/ipython.py
```diff
@@ -55,7 +58,7 @@ def __init__(self, metaclass=False):
     def get_param_info(self, obj, include_super=True):
         """
         Get the parameter dictionary, the list of modifed parameters
-        and the dictionary or parameter values. If include_super is
+        and the dictionary or parameter values. If include_s'https://tools.ietf.org/html/rfc7946'uper is
```
jbednar, line 61:
```suggestion
        and the dictionary or parameter values. If include_super is
```

tonyfast:
thanks for cleaning that up @jbednar bednar. sorry i missed it.

jbednar:
This looks good, but I can't tell whether it's the right thing to do. This file is called `ipython.py`, indicating to me that the entire file is about customizations specifically for ipython, but it sounds like it's used even when IPython is not in use (or else such changes wouldn't be needed?). I find that very confusing. `parameterized.py` loads it using a try/except, and seems to have code already for working around the absence of that file, but I think that's meant for when Param is vendored into another codebase.  If there is code in ipython.py that's meant to be usable with or without IPython, it seems to me it should move out of this file, rather than have this file have workarounds for when IPython is not in use. @jlstevens ?

jlstevens:
> If there is code in ipython.py that's meant to be usable with or without IPython, it seems to me it should move out of this file, rather than have this file have workarounds for when IPython is not in use.

A good bit of the code here is about generating param docs with ANSI color codes. In that sense, it is not ipython specific but ipython will be the most common environment where such colors can be nicely displayed. HoloViews uses this mechanism in `hv.help` to display information about plot options but I *believe* that will fall back to the normal `help` if it needs to (i.e if this file is unavailable). The issue is that you want to use the colors but only when they are supported, or you see the ANSI codes.

I would rather not add more files to param and I don't think it is a huge deal to have this functionality in a file called `ipython.py`. If this file was purely ipython specific, it would be tiny and there would be a chunk of code that would have to live somewhere else.

jbednar:
I'm just not seeing what this code has to do with IPython.  Is this using IPython somehow, even though I'm invoking it as just `python`?

image

It doesn't seem like checking for IPython is the right way to control the ANSI codes.

jbednar:
https://stackoverflow.com/questions/7445658/how-to-detect-if-the-console-does-support-ansi-escape-codes-in-python ?

https://bixense.com/clicolors/?

jlstevens:
I would vote for option 2 and docs to explain the presence of the flag when ANSI is not supported.

tonyfast:
closing because this not the solution.

PR #422, opened 2020-07-28
Refactor JSONSerializer

philippjfr:
Refactors the JSONSerializer into something that makes a bit more sense to me. There are now methods to serialize/deserialize a set of parameters into a single serialized object, and another set of methods to serialize/deserialize individual parameter values.

- [x] Add tests for `serialize_parameter_value` and `deserialize_parameter_value`

### DIFF
--- a/param/serializer.py
+++ b/param/serializer.py
@@ -28,8 +28,36 @@ class Serialization(object):
 
     @classmethod
     def serialize_parameters(cls, pobj, subset=None):
+        """
+        Should serialize the parameters on a Parameterized object
+        into a single serialized object, e.g. a JSON string.
+        """
         raise NotImplementedError        # noqa: unimplemented method
 
+    @classmethod
+    def deserialize_parameters(cls, pobj, serialized, subset=None):
+        """
+        Should deserialize a serialized object representing one or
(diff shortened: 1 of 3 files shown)


param/serializer.py
```diff
@@ -28,8 +28,36 @@ def schema(cls, pobj, subset=None):
 
     @classmethod
     def serialize_parameters(cls, pobj, subset=None):
+        """
+        Should serialize the parameters on a Parameterized object
```
jbednar, line 32: "Should"?

jbednar: Oh, you mean that subclasses should implement that?  I'd prefer to make that explicit:

```suggestion
        Serialize the parameters on a Parameterized object. Must be implemented by subclasses.
```

jlstevens:
Looks fine to me. The original intent of `serialize_parameters` and `deserialize_parameters` was closer to the new `serialize_parameter_value` and `deserialize_parameter_value` methods: I agree this way is clearer though.

param/serializer.py
philippjfr: Okay sure, will change that consistently.

PR #423, opened 2020-08-03
Remove dependency on nose (no explicit dependency on pytest)

mcepl:
Also, clear up a lot of warning and removing deprecated constructs.

Still remaining:

* Failing test methods test_abstract_class in
  tests/API1/testparameterizedobject.py and
  tests/API0/testparameterizedobject.py with:
```
=================================== FAILURES ===================================
____________________ TestParameterized.test_abstract_class _____________________

self = <tests.API0.testparameterizedobject.TestParameterized testMethod=test_abstract_class>

    def test_abstract_class(self):
        """Check that a class declared abstract actually shows up as abstract."""
>       self.assertEqual(_TestAbstractPO.abstract,True)
E       AssertionError: False != True

tests/API0/testparameterizedobject.py:117: AssertionError
____________________ TestParameterized.test_abstract_class _____________________

self = <tests.API1.testparameterizedobject.TestParameterized testMethod=test_abstract_class>

    def test_abstract_class(self):
        """Check that a class declared abstract actually shows up as abstract."""
>       self.assertEqual(_TestAbstractPO.abstract, True)
E       AssertionError: False != True

tests/API1/testparameterizedobject.py:134: AssertionError
```
* still unresolved use of deprecated method inspect.getargspec() on the
  line 2446 of param/parameterized.py.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -974,7 +974,7 @@ class String(Parameter):
       class IPAddress(String):
         '''IPv4 address as a string (dotted decimal notation)'''
        def __init__(self, default="0.0.0.0", allow_None=False, **kwargs):
-           ip_regex = '^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$'
+           ip_regex = '^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$'
            super(IPAddress, self).__init__(default=default, regex=ip_regex, **kwargs)
 
 
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,6 @@ extras_require = {
     # pip doesn't support tests_require
     # (https://github.com/pypa/pip/issues/1197)
     'tests': [
-        'nose',
         'flake8',
         'jsonschema',
     ]
--- a/tests/API0/testclassselector.py
+++ b/tests/API0/testclassselector.py
@@ -26,7 +26,7 @@ class TestClassSelectorParameters(unittest.TestCase):
 
     def test_single_class_instance_error(self):
         exception = "Parameter 'e' value must be an instance of int, not 'a'"
-        with self.assertRaisesRegexp(ValueError, exception):
+        with self.assertRaisesRegex(ValueError, exception):
             self.P(e='a')
 
     def test_single_class_type_constructor(self):
@@ -35,7 +35,7 @@ class TestClassSelectorParameters(unittest.TestCase):
 
     def test_single_class_type_error(self):
         exception = "Parameter 'str' must be a subclass of Number, not 'type'"
-        with self.assertRaisesRegexp(ValueError, exception):
+        with self.assertRaisesRegex(ValueError, exception):
             self.P(f=str)
 
     def test_multiple_class_instance_constructor1(self):
@@ -47,8 +47,8 @@ class TestClassSelectorParameters(unittest.TestCase):
         self.assertEqual(p.g, 'A')
 
     def test_multiple_class_instance_error(self):
-        exception = "Parameter 'g' value must be an instance of \(int, str\), not '3.0'"
-        with self.assertRaisesRegexp(ValueError, exception):
+        exception = r"Parameter 'g' value must be an instance of \(int, str\), not '3.0'"
+        with self.assertRaisesRegex(ValueError, exception):
             self.P(g=3.0)
 
     def test_multiple_class_type_constructor1(self):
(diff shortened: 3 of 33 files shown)


param/parameterized.py
```diff
@@ -974,7 +974,7 @@ class String(Parameter):
       class IPAddress(String):
         '''IPv4 address as a string (dotted decimal notation)'''
        def __init__(self, default="0.0.0.0", allow_None=False, **kwargs):
-           ip_regex = '^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$'
+           ip_regex = '^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$'
```
jbednar, line 977:
```suggestion
           ip_regex = r'^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$'
```
?

philippjfr:
>What's the motivation to avoid nose?

The motivation is that it's unmaintained and no one should use it at this point. I wish all our projects had switched over to pytest years ago so I'm very happy to see this.

mcepl:
> > What's the motivation to avoid nose?
> 
> The motivation is that it's unmaintained and no one should use it at this point. I wish all our projects had switched over to pytest years ago so I'm very happy to see this.

Not only that, but there are rumours that nose1 will stop working with Python 3.9/3.10, and given there is no support for it, it seems easier just to run away from it.

param/parameterized.py
jbednar: @mcepl, is the above change ok? I did it to match the rest of your changes, where you used raw strings rather than escaping, but I haven't tested that it's equivalent.

ceball: Bit confusing, but if you expand the diff up a bit...

Screenshot from 2020-09-17 22-53-19

I.e. we are well inside a string by now :)

There are two things:
1. suggesting the right way for someone to write this code themselves
2. not getting a deprecation warning (?) from param itself

The suggested edit would be ok for 1, but I guess it does not help for 2.

The original is presumably ok for 2 (that is what you were doing I guess?), and I _think_ it's ok for 1 also. But I can't remember exactly.

ceball:
> Failing test methods test_abstract_class in tests/API1/testparameterizedobject.py and tests/API0/testparameterizedobject.py with:
>```
>    def test_abstract_class(self):
>        """Check that a class declared abstract actually shows up as abstract."""
>>       self.assertEqual(_TestAbstractPO.abstract,True)
>E       AssertionError: False != True
>```

@mcepl I think you'll either need to rename these "test classes" so that they don't have a leading underscore, or fix the (longstanding) assumption in param's abstract class mechanism that abstract classes don't start with a leading underscore.

From e.g. https://docs.python.org/2/tutorial/classes.html:

> Any identifier of the form __spam (at least two leading underscores, at most one trailing underscore) is textually replaced with _classname__spam, where classname is the current class name with leading underscore(s) stripped. This mangling is done without regard to the syntactic position of the identifier, as long as it occurs within the definition of a class.

https://github.com/holoviz/param/blob/301b9617e84f3578364393d5ae30c452d1c19cd0/param/parameterized.py#L2074-L2094

(i.e. param does not do the same thing as python)

If you go the renaming route, the test class names are not very good anyway - going from e.g. `_TestClass` to `SomeTestClass` or whatever you like wouldn't lose any meaning...

mcepl:
Like this?

maximlt:
#509 removed nose as a dependency and test runner in favor of pytest. Contrary to this PR, #509 focused on that transition, added an explicit dependency on pytest (using e.g. `pytest.mark.skipif(...)`) and set up the CI to run.

What is in this PR that was not done in #509:

* Suppressing deprecation warnings caused by `self.assertRaisesRegexp` by using `self.assertRaisesRegex` instead: this would fail for Python 2.7 so I guess we should wait until we drop it
* Turned strings into raw strings to fix them when they contain an invalid escape sequence: flake8 would have spotted those if the warning W605 wasn't skipped (and if flake8 would be allowed to run on the tests folder). I'll use that work in a new PR.

What was done differently:

* Renamed classes called `TestSomething` that were not meant to be test classes (previously decorated with `nottest` provided by the nose framework). In #509 these classes were not renamed, instead a `__test__` class variable was added and set to False. this is a feature of pytest.

Thanks @mcepl for the impetus on this work!

mcepl:
>     * Renamed classes called `TestSomething` that were not meant to be test classes (previously decorated with `nottest` provided by the nose framework). In [Replace nose with pytest #509](https://github.com/holoviz/param/pull/509) these classes were not renamed, instead a `__test__` class variable was added and set to False. this is a feature of pytest.

It seems to me silly to exchange one hard coded dependency on a non-standard library by the hard coded dependency on another non-standard library, but as you wish. I won't fix these problems when pytest goes the way of dodo in few years.

PR #424, opened 2020-08-08
Micro optimizations

philippjfr:
Relatively silly micro-optimizations which do add up when you're creating thousands of objects. In my particular usecase 5000 parameterized objects were created and these changes shaved 50-100 ms off from the 500 ms spent on creating parameterized objects.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -19,9 +19,9 @@ except ImportError:
 
 
 from collections import defaultdict, namedtuple, OrderedDict
+from functools import partial, wraps, reduce, lru_cache
 from operator import itemgetter,attrgetter
 from types import FunctionType
-from functools import partial, wraps, reduce
 
 import logging
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


param/parameterized.py
```diff
@@ -151,6 +151,7 @@ def discard_events(parameterized):
         parameterized.param._events = events
 
 
+@lru_cache(maxsize=128)
```
jbednar, line 154: lru_cache wasn't [introduced until Python 3.2](https://docs.python.org/3/library/functools.html#functools.lru_cache). and I don't think Param should abandon Python 2.7 until Param version 2.0. Maybe define it as a no-op for 2.7?

philippjfr:
I'll look into it and maybe start introducing our own memorization function since that was always on the roadmap for param depends.

param/parameterized.py
philippjfr: Removed the LRU cache for now but may end up readding it for something else later.

Issue #313, opened 2019-01-27
Param documentation


jbednar:
Param's website param.pyviz.org is appalling.  It has a brief intro, but no real documentation, with no clue about any of the more powerful features of Param (certainly not any of the more recently added ones like depends, watch, and output) and outdated pointers to ParamNB and ParamTk and no mention of Panel.  Some previous issues mention bits of docs that do exist that need to be moved into the website (https://github.com/pyviz/param/issues/157, https://github.com/pyviz/param/issues/94), but really it needs some dedicated effort to provide actual, up to date content, because it's such a big part of how to use Panel.

jbednar:
Just to start a grab bag of items that need documentation:

- Controlling order (definition order on py3.6+, sorted alphanumerically py2.7-3.5, within each precedence value; see https://github.com/pyviz/param/issues/185).

cocoaaa:
Hi, having these items documented would be really helpful. Do you think the documentation will be updated in the near future? Or, are there places I can read more about them -- esp. item 3,4,5,6?

Thanks!

ceball:
Extends #47 and #157 (have closed those).

ceball:
@cocoaaa Bit late for your question, but I would look at panel's docs and examples for now (e.g. see panel.pyviz.org).

kgullikson88:
@jbednar is there any documentation anywhere for some of those things? I am trying to figure out what `watch` does by turning it off and on, and can't tell. The controlling order thing is also something I've been trying to find and it sounds like it is possible.

jbednar:
The docs at panel.pyviz.org are your best bet for now; search those for "param" and "watch" and you should see examples of watch in heavy usage. Watch doesn't need Panel, of course, but that's where it's currently documented. I apologize for that; my only defense is that no one has ever paid us to document Param, and we're paid to work on a whole lot of other things that don't leave any time to work on Param docs.  I can't even guess how long "document Param" has been `#`7 on my priority list, and only the first few ever get addressed. :-/

kgullikson88:
No problem, I totally understand and appreciate the work you do!

jbednar:
FYI: We now have paid-for time to develop proper Param documentation, so it will happen sometime in 2020!  Sooner rather than later, I hope, but we have a large set of tasks in the current plan of work, and this one isn't the first in the list.

maximlt:
Param has now a proper documentation website, closing :)

Issue #349, opened 2019-07-25
pipeline @param.output usage


kcpevey:
This is a little strangeness. I'm not sure if this is anticipated behavior. If so, then it should be documented.  
  
this works (passes the anticipated owi_obj):  
```
@param.output(owi_obj=winds.OWI)
def output(self):  
        return self.execute()
```
This doesn't work (passes owi_obj as None):  
```
@param.output(owi_obj=winds.OWI)
def owi_obj(self):  
        return self.execute()
```
And this doesn't work (passes owi_obj as None):  
```
@param.output()
def owi_obj(self):  
        return self.execute()
```

jbednar:
> This doesn't work (passes owi_obj as None):
```
@param.output(owi_obj=winds.OWI)
def owi_obj(self):  
        return self.execute()
```
Shouldn't this be the same as the following, and if so, does that work?
```
@param.output(winds.OWI)
def owi_obj(self):  
        return self.execute()
```

ppwadhwa:
@kcpevey Can you summarize what changes we should include in the docs?  Thank you!

PR #428, opened 2020-08-20
New website and documentation

jbednar:
A long-overdue website for Param.  Essentially complete apart from the User Guide, which has only one page in it so far. Everything that's there is ready for review and comment, and feel free to volunteer to fill in one of the stubs in the User Guide or to suggest new sections for the User Guide.

image

Philipp suggested that we might consider moving to a different docs building tool before going further, which is fine by me but not something I'm going to set up personally.  Now's the time, if so!  The existing content is not heavily linked with each other, nor is the site particularly customized, so it wouldn't be much work to migrate at this point. That said, it works already (at least locally), so I'm not sure whether it's worth messing with at the moment.


jbednar:
Closes https://github.com/holoviz/param/issues/313 when merged.

jbednar:
Partial to-do list, beyond filling out the User Guide sections that are already listed:

- [ ] Harvest documentation from parameterized.py (and delete it from there)
- [x] Content from http://ioam.github.io/topographica/User_Manual/parameters.html
- [ ] Controlling order (definition order on py3.6+, sorted alphanumerically py2.7-3.5, within each precedence value; see https://github.com/pyviz/param/issues/185).
- [ ] `@param.depends`
- [ ] watch
- [ ] how and why to define a new Parameter class
- [ ] how to separate content from presentation using Parameters
- [ ] controlling parameters via command-line arguments, environment variables, .yml files, .json files, etc.
- [ ] param.output (#349 )
- [ ] Harvest some of the docs from Panel and refer to those docs where appropriate

philippjfr:
I've moved all text into markdown files and switched over to the pydata_sphinx_theme. Given how long this has been sitting now  I'm just going to merge.

PR #429, opened 2020-08-24
Ensure the color class applies allow_none

### LINKED ISSUE #303: Color does not respect allow_None
Color throws away the keyword ``allow_None``. See https://github.com/ioam/param/blob/c17db9bc02f35d55dc85f58d09b7ea5ab03c4171/param/__init__.py#L1714

tonyfast:
closes #303 by removing the allow_none parameter from the function signature.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1910,7 +1910,7 @@ class Color(Parameter):
     prefix.
     """
 
-    def __init__(self, default=None, allow_None=False, **kwargs):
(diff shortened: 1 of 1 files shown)


jbednar:
Thanks. I'm surprised this error wasn't caught by a linter; seems like it should be checking that we are doing _something_ with each argument we are given to a method (unless marked explicitly `noqa`).

PR #430, opened 2020-09-07
bug-fix in discard_events

thomasdeneux:
`watchers` and `events` need to be copied before being memorized, otherwise they are mere references to `parameterized.param._watchers` and `parameterized.param._events` and the `finally` block has no effect (i.e. does not ignore new events that have accumulated during the scope of the context manager)


philippjfr:
Sorry I didn't review this, the tests are failing because one of the tests is actually wrong. Have pushed a new PR in #442 fixing this.

PR #431, opened 2020-09-13
Fix indentation

ceball:
3 spaces?! It took me a while to realize why my editor was constantly trying to use 3 spaces. It's like you're trying to illustrate why we should just auto-format the code and be done with this kind of thing ;)


jlstevens:
I've noticed that emacs can get confused about spacing in other parts of param which suggests there are other bits of inconsistent indentation floating about somewhere...

PR #432, opened 2020-09-13
Test suite fixups

### LINKED ISSUE #246: Switch to unittest2 on older pythons
In tests/__init__.py, we have some redundant code to use unittest2 on python < 2.7:

```
if sys.version_info[0]==2 and sys.version_info[1]<7:
    del sys.modules['unittest']
    sys.modules['unittest'] = __import__('unittest2')
```

(redundant since we dropped 2.6 and earlier)

I plan to change that (and add conditional dependency to list of test dependencies) so we use unittest2 for whichever versions of python we support that don't use the current python 3 unittest. 

(One small example: assertRegexp type names are deprecated in py3 unittest - they're assertRegex now.)

### LINKED ISSUE #433: test suite failing on travis because jsonschema now unavailable on py2
```
Collecting pyrsistent>=0.14.0 (from jsonschema->param==1.10.0a5.post5+gf3fb24f)
  Downloading https://files.pythonhosted.org/packages/7d/ae/90ddcf28fb8eee5d4990920586d2856342e42faa95f39223f0b9762ef264/pyrsistent-0.17.2.tar.gz (106kB)
pyrsistent requires Python '>=3.5' but the running Python is 2.7.15
```

ceball:
* Remove redundant handling for python<2.7 in test suite (we don't support python<2.7).
* Don't use python 2.7 as a primary test environment on travis, use python 3.7. (python 2.7 is still tested.)
* Follow existing pattern for testing with optional components: jsonschema
* Re-enable testing with gmpy present
* Add python 3.9 on travis (3.9 comes out soon).
* All tests are now being run on travis.
* Measure coverage with all optional tests running.
* Add numbergen to modules being checked for coverage.
* Merged duplicate/inconsistent flake8 sections, with a set of rules that lead to no complaints about trivial stuff.
* Replaced a named lambda with a function (because flake8 complained :) I don't have much opinion).
* Fix various indentation issues and add indentation to flake checks (we can at least agree on 4 spaces only, not mixing with 3, I hope...)

Fixes #433 
Fixes #246

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,25 +1,18 @@
 [tox]
 envlist =
-    py38,py37,py36,py35,py27,pypy,
-    {py27,py37}-flakes,
-    {py27,py37}-with_numpy,
-    {py27,py37}-with_ipython
-    {py27,py37}-with_pandas
+    py39,py38,py37,py36,py35,py27,pypy,
+    py37-flakes,
+    py37-coverage,
+    py37-with_numpy,
+    py37-with_ipython,
+    py37-with_pandas,
+    py37-with_jsonsschema,
+    py37-with_gmpy,
+    py37-with_all
 
 [testenv]
 deps = .[tests]
 commands = nosetests
-
-[testenv:coverage]
-# remove develop install if https://github.com/ioam/param/issues/219
-# implemented
-setdevelop = True
-passenv = TRAVIS TRAVIS_*
-deps = {[testenv]deps}
-       coveralls
-commands = nosetests --with-coverage --cover-package=param
-           coveralls
-# TODO missing numbergen
 
 [testenv:with_numpy]
 deps = {[testenv]deps}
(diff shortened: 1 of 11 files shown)


ceball:
@jbednar a post-merge review would be welcome in case I've overlooked something/done something bad, if you have time

tox.ini
```diff
@@ -1,26 +1,19 @@
...
-    {py27,py37}-with_pandas
+    py39,py38,py37,py36,py35,py27,pypy,
+    py37-flakes,
+    py37-coverage,
+    py37-with_numpy,
+    py37-with_ipython,
+    py37-with_pandas,
+    py37-with_jsonsschema,
+    py37-with_gmpy,
+    py37-with_all
```
jbednar, line 11: At some point we should presumably migrate py37-x to py38-x or py39-x, whenever the number of 3.8 or 3.9 users dominates the number of 3.7 users. I.e., those tests only need to run on one Python version, and it may as well be whatever is most popular...

jbednar (commented):
Looks great! Thanks so much.

tox.ini
ceball: Yes, that's right. I don't know a way to express "current default python" (which would be set to py37) in the tox file. I'll add a comment.

PR #434, opened 2020-09-13
Group depends for watched deps within a class

### LINKED ISSUE #415: set_param not working as expected
Param: 1.9.3
Panel 0.9.5

Posted here on request of @philippjfr based on https://discourse.holoviz.org/t/how-to-trigger-param-depends-only-once-when-updating-multiple-parameters/848

#### Description of expected behavior and the observed behavior

I would expect to be able to use `set_param` to set multiple parameters but only run depending functions once. At least thats how @philippjfr says it should be working.

That is not how it is working for me.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import param
import panel as pn
pn.extension()

class MultiParamApp(param.Parameterized):
    count = param.Integer(0)
    param1 = param.Integer(0)
    param2 = param.Integer(0)
    update_once = param.Action()
    update_once_hack = param.Action()
    reset = param.Action()

    view = param.Parameter()

    def __init__(self, **params):
        super().__init__(**params)

        self._updating = False

        self.update_once = self._update_once
        self.update_once_hack = self._update_once_hack
        self.reset = self._reset

        self.view = pn.Param(
            self,
            widgets={
                "update_once": {"button_type": "danger"},
                "update_once_hack": {"button_type": "success"},
            },
        )

    @param.depends("param1", "param2", watch=True)
    def _inc(self):
        if not self._updating:
            self.count += 1

    def _update_once(self, _=None):
        self.param.set_param(param1=self.param1 + 1, param2=self.param2 + 1)

    def _update_once_hack(self, _=None):
        self._updating = True
        self._update_once()
        self._updating = False
        self._inc()

    def _reset(self, _=None):
        self._updating = True
        self.param1 = 0
        self.param2 = 0
        self.count = 0
        self._updating = False


MultiParamApp().view.servable()

```

#### Screenshots or screencasts of the bug in action

If I click the `Update Once` button I run the `_update_once` function but the `_inc` function is triggered twice. I would have wanted/ expected it to be triggered once.

image

If I then click the `Update Once hack` button I get the expected behaviour.

image

ceball:
Supposed to fix #415...but let's see if the existing test suite passes first...

I probably need a review from @philippjfr at this point before I go further, as 1. I'm a bit out of practice ;) and 2. seemingly if I ever try to inspect anything param-ish using pdb, I get `Fatal Python error: Cannot recover from stack overflow` and a core dump :( (no doubt related to the ever recurring repr problems).

To do:

* [ ] add tests

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2481,9 +2481,9 @@ class Parameterized(object):
                 # instantiation of Parameterized with watched deps. Will
                 # probably store expanded deps on class - see metaclass
                 # 'dependers'.
-                grouped = defaultdict(list)                
+                grouped = defaultdict(list)
                 for dep in self.param.params_depended_on(n):
-                    grouped[(id(dep.inst),id(dep.cls),dep.what)].append(dep)                    
+                    grouped[(id(dep.inst),id(dep.cls),dep.what)].append(dep)
                 for group in grouped.values():
                     # TODO: can't remember why not just pass m (rather than _m_caller) here
                     (dep.inst or dep.cls).param.watch(_m_caller(self, n), [dep.name for dep in group], dep.what, queued=queued)
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2486,7 +2486,8 @@ class Parameterized(object):
                     grouped[(id(dep.inst),id(dep.cls),dep.what)].append(dep)
                 for group in grouped.values():
                     # TODO: can't remember why not just pass m (rather than _m_caller) here
-                    (dep.inst or dep.cls).param.watch(_m_caller(self, n), [dep.name for dep in group], dep.what, queued=queued)
+                    gdep = group[0] # Need to grab representative dep from this group
+                    (gdep.inst or gdep.cls).param.watch(_m_caller(self, n), [d.name for d in group], gdep.what, queued=queued)
 
         self.initialized = True
 
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -494,7 +494,7 @@ def _params_depended_on(minfo):
 
 
 def _m_caller(self, n):
-    def caller(event):
+    def caller(*events):
         return getattr(self,n)()
     caller._watcher_name = n
     return caller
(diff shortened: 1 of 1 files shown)


jlstevens:
I bumped into this issue while working on the new Event parameter:

image

The tests have passed and I would like to see this merged! At a glance the code looks good but I haven't looked closely enough yet..

param/parameterized.py
```diff
@@ -2481,9 +2481,12 @@ def __init__(self, **params):
...
                 # probably store expanded deps on class - see metaclass
                 # 'dependers'.
-                for p in self.param.params_depended_on(n):
+                grouped = defaultdict(list)
+                for dep in self.param.params_depended_on(n):
+                    grouped[(id(dep.inst),id(dep.cls),dep.what)].append(dep)
+                for group in grouped.values():
                     # TODO: can't remember why not just pass m (rather than _m_caller) here
-                    (p.inst or p.cls).param.watch(_m_caller(self, n), p.name, p.what, queued=queued)
+                    (dep.inst or dep.cls).param.watch(_m_caller(self, n), [dep.name for dep in group], dep.what, queued=queued)
```
jlstevens, line 2489: I'm confused by this line: `dep` (in `dep.inst, dep.cls` and `dep.what`) looks like they refer to the leaky loop variable left over from the previous `for dep in self.param.params_depended_on(n)` loop.

jlstevens:
```suggestion
                    gdep = group[0] # Need to grab representative dep from this group
                    (gdep.inst or gdep.cls).param.watch(_m_caller(self, n), [d.name for d in group], gdep.what, queued=queued)
```

ceball: 🤦 ha ha

My excuse is all I ever saw was core dumps :)

ceball:
@jlstevens I guess you will add a test and finish this off? (I'm very happy with that if so - thanks :) )

param/parameterized.py
```diff
@@ -494,7 +494,7 @@ def _params_depended_on(minfo):
 
 
 def _m_caller(self, n):
-    def caller(event):
+    def caller(*events):
```
jlstevens, line 497: Did this need to change? `event` isn't even used so why does it matter if it is `event` or `*events`? (other than the semantically better pluralization)

Can anyone supply a docstring for this function?

jlstevens:
@philippjfr @ceball Jim and I agree to merge now but please do a post review merge and view the comments (I'm now going to test this on another PR).

param/parameterized.py
ceball: > Did this need to change? 

Because now you can have more than one event? (untested, as was comical code you spotted earlier)

ceball:
As an aside, note similarities to code elsewhere that I was using as a basis for this change:

https://github.com/holoviz/param/blob/301b9617e84f3578364393d5ae30c452d1c19cd0/param/parameterized.py#L368-L377

(i.e. string specs vs not - the separation makes it difficult to remember - just saying/not that I am going to implement a solution to that)

param/parameterized.py
jlstevens: That is what we suspected... :-)

jlstevens:
Looks like it is working now:

image

The only thing I am wondering is whether you might expect the callbacks to execute in definition order (at least in Python 3 where things are generally ordered) i.e 'A' then 'B' then 'C'. I don't think we ever promised such semantics but it might be nice if it is easy to provide...

Issue #435, opened 2020-09-13
parameters good, parameterized bad / "gradual" param


ceball:
Param looks like an interesting library, but unfortunately I can never use it because by the time I ever get to a codebase, it's too late! People already wrote their classes and stuff, and it's not possible to make them inherit from parameterized.

I've proposed variants of "gradual parameterization" in the past. One day I'll try to write a decent overall proposal here in this issue. However, looking at param's code today I saw a comment:

https://github.com/holoviz/param/blob/b785c42e8745644ad74aa0ace758925f311443e1/param/parameterized.py#L2190-L2191

Usually my reminder comments fade into obscurity 30 mins after being written, but that one is actually useful. Following it (see the associated PR) , I can immediately get:

```python
import param

class A:
    a0 = param.Number(default=0,doc="a0 doc")
```

```
>>> a = A()
>>> a.a0 = "number"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/sefkw/proj/param-gradual/param/param/parameterized.py", line 303, in _f
    return f(self, obj, val)
  File "/home/sefkw/proj/param-gradual/param/param/__init__.py", line 623, in __set__
    super(Dynamic,self).__set__(obj,val)
  File "/home/sefkw/proj/param-gradual/param/param/parameterized.py", line 303, in _f
    return f(self, obj, val)
  File "/home/sefkw/proj/param-gradual/param/param/parameterized.py", line 873, in __set__
    self._validate(val)
  File "/home/sefkw/proj/param-gradual/param/param/__init__.py", line 882, in _validate
    raise ValueError("Parameter '%s' only takes numeric values"%(self.name))
ValueError: Parameter 'a0' only takes numeric values
```

Of course, Parameterized and ParameterizedMeta class do a bunch of other stuff. However, I feel like it should be possible to factor that stuff out of those and also make it available another way. E.g. via a decorator people can use on their class. Or a number of decorators, each of which adds some particular functionality, if it makes sense on its own.

jbednar:
Very interesting!  You still have to edit the source code of `A`, but here `A` could be inheriting from some class that you _can't_ modify.  I'm in favor of making the power of Param available in cases where the full approach can't be adopted but we can still provide some benefit.

ceball:
Right, the fundamental thing is getting benefits of parts of param without having to go all in.
[...]

PR #252, opened 2018-08-24
Demonstrate some support for CLIs

ceball:
"write parameterizedfn(s) (with parameters), turn into commands (with options)"

http://nbviewer.jupyter.org/github/ceball/snippets/blob/master/cmd.ipynb

PR #436, opened 2020-09-13
Alternative to inheriting from parameterized

ceball:
See #435 

This isn't ready for any kind of review, it's just for visibility. I debated whether or not to make a PR at this stage; I'd say it's in the same kind of category as #252 in terms of speculative nature (though there is a bit more chance I'll work on this one), so we could close and leave just the corresponding issue.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,8 +3,6 @@ environment:
     - TOXENV: py38
     - TOXENV: py37
     - TOXENV: py36
-    - TOXENV: py35
-    - TOXENV: py27
 
 matrix:
   fast_finish: true
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,6 +9,11 @@ matrix:
 
 build: false
 
-install: pip install tox
+install:
+  # use python 3.7 to set things up rather than appveyor's default of 2.7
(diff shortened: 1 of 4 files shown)


jbednar:
Looks cool! Is there any downside other than not supporting Python2?

ceball:
> Is there any downside other than not supporting Python2?

Got to test it, got to document it, etc 🤣

I'll answer seriously at some point. But "got to document it" is partially serious already; the more ways to use param:
1. the more documentation to write
2. the more confusing it can be to newcomers
3. the more confusing it can be when coming to "parameterized" code (currently, you definitely know what to expect when you encounter code using param)

I feel like 1 and 2 will be ok, because I can see a story about ways to start using param that makes sense.

3 I'm not sure about. If this PR ends up offering a different way to get to the same behavior as current param, there's no issue. If it ends up allowing people to get bits of param, it might result in more confusing code. Not sure.

jlstevens:
Definitely an interesting idea! 

Being able to use parameters without inheritance is not something I would recommend when starting from scratch but as a way to help migrate to param it seems very useful.

droumis:
Closing for now because of incompletion

PR #438, opened 2020-09-16
Add new Event parameter type

jlstevens:
Very WIP PR to illustrate an idea we have been discussing. The API will probably change but here is the current behavior:

image

The idea is to add a new `Trigger` parameter type which would replace many uses of `Action`. What you want to know about a  `Trigger` parameter is whether it has been triggered (e.g a button clicked). The value is True if the parameter has just been triggered, otherwise False (which means a trigger call has happened since not involving that parameter).

One possibility which is more complicated is to avoid `b.param.trigger('a')` by setting `b.a = True` (setting to True triggers, whether or not the value is True already). This involves working inside `__set__` which can get quite tricky.

At this point, this PR aims to be a place to discuss this suggestion more than something to merge. The idea is that this parameter would mean https://github.com/holoviz/holoviews/pull/4611 is not necessary (though panel would need to be updated to display this parameter type as appropriate buttons).

### TODO

- [x] Make this work with class parameters (doesn't work with `Bar` at the moment in the example above)
- [ ] Improve API, possibly via `__set__`


jbednar:
Thanks. Can you add a more complete example using `depends` or `watch` to show how the end-to-end behavior would look?  I'm having trouble seeing how the full circle of triggering an action that then completes would look.

jbednar:
I tried adding such an example myself, but can't make sense of the results:

image

I had expected that `fn` would be executed when `a` gets triggered, and no other time. Instead it gets executed _twice_ for `a` being triggered, and also gets executed when `c` is triggered?  I'm confused!

jbednar:
Ok, I guess the two triggers are because 'a' is being set to True then False:

image

Not sure about the third one, though, and having two in the first case doesn't seem desirable for something that's explicitly for triggering, and is thus not about the value.

jlstevens:
Now works with classes and no double triggering:

image

jlstevens:
@jbednar The value should now always be False outside of the triggered function.

param/__init__.py
```diff
@@ -2040,3 +2040,6 @@ def _validate(self, val):
         # values are numeric, which is redundant, so just call
         # _checkBounds().
         self._checkBounds(val)
+
+
+class Trigger(Boolean): pass
```
jbednar, line 2045:
```suggestion
class Trigger(Boolean): pass

```

jbednar:
As far as I can see, this is a reasonable set of behavior now:

image

Thanks for updating it and addressing those quirks!

So let's see if we can keep this behavior as a reference and consider the implementation, the API, and the naming.

Taking a step back, the user story here is nominally about a Trigger parameter, but actually the Trigger Parameter itself is a no-op as currently implemented. Every Parameter can already have `.trigger` called on it, and the only thing special about the current param.Trigger happens for special casing for param.Trigger within `.trigger`.  That's spooky-action-at-a-distance stuff rather than clean object-oriented design; e.g. people can't inspect the source code of Trigger to find out anything about how it works.  So my gut reaction is not to like having a parameter called Trigger that doesn't actually implement anything about triggering. If we do want triggering to be implemented in this way, I would steer more towards some flag declared by a Parameter that enables this behavior rather than a specially named dummy Parameter type.

Would the `set` approach be able to be localized into the Parameter's definition, if it's feasible? If so that would be a big advantage of doing it that way.  Plus using `set` would map more cleanly to the semantics of Trigger, because the value does change during triggering, so it's nice to have that be explicit (even if the changing back to False is implicit). I'd be happy to have a set-based Trigger be called param.Trigger if the implementation is inside the Trigger class, whereas the current Trigger struggles to have the name Trigger be appropriate for it and seems like it shouldn't exist in that form.

jlstevens:
To be clear, the `set` approach means running the trigger with something like:

```
foo.button = True
```

Instead of `foo.param.trigger('button')`.

That *might* be possible by overriding the `__set__` method on this new parameter (I can't be sure yet) but I am sure that the special logic in the `trigger` method (in the `Parameters` namespace object) would still be required.

jbednar:
I'll see once there's an implementation.  In any case, ideally we'd have something other than the name Trigger to govern that special behavior if it's still needed.

jlstevens:
I sort of have it working:

image

It now relies on the change in value to run the callback (apparently most things don't even use trigger). Unfortunately `discard_events` doesn't seem to work inside ` __set__` (events were still being generated when setting to False) so I copied the core parts of `__set__` without all the watcher stuff to reset to False.

jlstevens:
At this point in the history, the implementation is simple as I've removed the changes from the trigger method (which still needs to be handled somehow).

jlstevens:
Explicit triggering now handled:

image

This does what I would expect now so I'll leave this PR alone till I get some feedback.

jbednar:
I've tested the current commits, and it doesn't seem like it's working reasonably, because the callback keeps getting fired for things that have nothing to do with `a`:

image

I'd only have expected it to be triggered for the single call to `.trigger('a')`, but it's also being called when other things are triggered? I suspect the changes to `.trigger` are forcing triggering for all Trigger parameters even if they weren't the ones being called here.

> One thing that isn't clear to me, should a become True when running b.param.trigger('a')? That would need special handling...

I don't _think_ a should become True when merely triggering `a`; I _think_ the value should change only when `a` is set (and then only transiently), leaving `.trigger` to simply trigger without a change in value.  I think that's more consistent with other parameters, leaving the only weird aspect of this one to be the fact that it auto-resets itself whenever you try to set it.  

If I remove the changes to `.trigger`, it seems to behave reasonably to me:

image

Without those changes, it's now fully self contained, fully local in its implementation and effects, and easy to describe: a Trigger parameter is one that can be set, but where setting is always transient, used only to trigger other actions, with the value immediately returning to the baseline by the end of the setting operation.  

So I'm happy with it at the param level, if you go ahead and remove the `.trigger` code changes altogether.  If you agree, then it's up to you (@jlstevens) and @philippjfr to say whether this is sufficient for HoloViews stream needs -- does this transient setting achieve the correct set of values and triggering events for you to use as part of the overall stream handling? If it does, then we're done and ended up with only a few, fully localized lines of code!

jlstevens:
Ok, I am happy to remove the latest changes to the trigger method again as the behavior on explicit triggering wasn't obvious to me (and not introducing special handling might well be the most sensible option).

As soon as there is some panel code that can display Trigger parameters as buttons, I'll be able to test to see if this properly supersedes the HoloViews PR.

jlstevens:
@jbednar Note, I spotted the bug in the code I just reverted: I simply forgot to filter the trigger parameters against `param_names`. So I think I *could* make `foo.param.trigger('a')` have `a` be True transiently (without executing unrelated triggers). Would you like to try it out or are you sure this behavior isn't a good idea?

param/__init__.py
```diff
@@ -2040,3 +2040,16 @@ def _validate(self, val):
         # values are numeric, which is redundant, so just call
         # _checkBounds().
         self._checkBounds(val)
+
+
+class Trigger(Boolean):
+
+    @instance_descriptor
+    def __set__(self,obj,val):
```
philippjfr, line 2048: Nitpicky but I'd like param to eventually have a consistent style again.

```suggestion
    def __set__(self, obj, val):
```

philippjfr:
>So I think I could make foo.param.trigger('a') have a be True transiently (without executing unrelated triggers). Would you like to try it out or are you sure this behavior isn't a good idea?

I think that should work because otherwise someone could trigger it and get `False` as a value which seems incorrect.

jlstevens:
Ok, both versions can work! With the latest commit, unrelated triggers (e.g `b.param.trigger('c')`) aren't being executed:

image

I think I prefer this than leaving the value False. As either version works, I am happy to pick either as long as we have a consensus on the correct behavior.

param/__init__.py
jbednar: Resolved, but not applied?

jbednar:
I see arguments above saying that it could be confusing to invoke `.trigger` and not see the value change, but I don't personally think that's confusing, because `.trigger` never changes the value for any other Parameter, so I don't see why people would expect to see it do so here. I think it's up to @philippjfr ; if the changes to .trigger here aren't needed for holoviews or panel, I'd rather keep the code and behavior localized to the Trigger class definition, but if it makes things work more consistently then it's ok to keep the changes to .trigger.

jlstevens:
> ... so I don't see why people would expect to see it do so here

In general I would agree, but for a parameter named `Trigger` it doesn't seem unreasonable to have semantics specially related to `.trigger`. I have a slight preference for consistency in terms of having a value of `True` while triggering but I am also happy to let @philippjfr  decide.

jbednar:
> for a parameter named Trigger it doesn't seem unreasonable to have semantics specially related to .trigger

Right; I was thinking about that too. Maybe it needs a better name?  param.Transient?  param.Pulse?  In electronics it's called a momentary switch, but I don't think that's useful here. 

It's hard to think of something that will make sense at the app level _and_ at the Param level.

jlstevens:
`param.Event` (as in a 'button press event')? Though to me events are typically more structured than this...

jbednar:
That's a contender. Right, events _typically_ have a payload and this doesn't, but it's true it's an event...

jbednar:
`param.Pulse`?

jlstevens:
I prefer `Event` to `Pulse`.

philippjfr:
I've also been considering `param.Event`, my only reservation was that `param.parameterized.Event` exists for the watch Events.

jlstevens:
I think it isn't a problem to have an `Event` parameter as well as an actual `Event` object...

param/__init__.py
```diff
@@ -2040,3 +2040,16 @@ def _validate(self, val):
         # values are numeric, which is redundant, so just call
         # _checkBounds().
         self._checkBounds(val)
+
+
+class Trigger(Boolean):
```
jbednar, line 2045: Proposed docstring, assuming both sets of changes are kept (Trigger and .trigger):

```suggestion
class Trigger(Boolean):
    """
    A Trigger Parameter is one whose value is intimately linked to the triggering of
    events for watchers to consume. Trigger has a Boolean value, which when set to
    True triggers events sent to watchers (as any Parameter does) and then is
    automatically set back to False. Conversely, if events are triggered directly via
    `.trigger`, the value is transiently set to True (so that it's clear which of many
    parameters being watched may have changed), then restored to False when the
    triggering completes. Trigger is thus like a momentary switch or pushbutton with
    a transient True value that serves only to launch some other action (e.g. via a
    param.depends decorator), rather than encapsulating the action itself as
    param.Action does.
"""
```

param/parameterized.py
```diff
@@ -1490,19 +1490,24 @@ def trigger(self_, *param_names):
         will be triggered whether or not the parameter values have
         actually changed.
```
jbednar, line 1491:
```suggestion
        actually changed. As a special case, the value will actually be changed for a Parameter
        of type Trigger, setting it to true so that it is clear which Trigger parameter has been
        triggered.
```

param/__init__.py
jlstevens: Used this as a template for the docstring pushed in af0a5d8

param/parameterized.py
jlstevens: Used this as the template for the docstring pushed in af0a5d8

jlstevens:
@philippjfr @jbednar I prefer the name `Event` a fair bit and think the chance of it being confused with the namedtuple is low.

If we are all agreed with this, I can now push some unit tests and hopefully get this PR merged!

param/parameterized.py
```diff
@@ -1488,21 +1488,28 @@ def trigger(self_, *param_names):
...
         will be triggered whether or not the parameter values have
-        actually changed.
+        actually changed. As a special case, the value will actually be
+        changed for a Parameter of type Event, setting it to True so
+        that it is clear which Event parameter has been triggered.
         """
+        trigger_params = [p for p in self_.self_or_cls.param
+                          if (self_.self_or_cls.param[p].__class__.__name__ == 'Event')]
+        for pname in [p for p in trigger_params if p in param_names]:
+            setattr(self_.self_or_cls, pname, True)
```
philippjfr, line 1498: If you're only triggering a trigger param shouldn't it return after this section? I guess an empty `set_param` is a no-op, still seems safer though. Also shouldn't this use `set_param` rather than `setattr` so the events are batched?

param/parameterized.py
```diff
@@ -1488,21 +1488,28 @@ def trigger(self_, *param_names):
         """
         Trigger watchers for the given set of parameter names. Watchers
         will be triggered whether or not the parameter values have
-        actually changed.
+        actually changed. As a special case, the value will actually be
+        changed for a Parameter of type Event, setting it to True so
+        that it is clear which Event parameter has been triggered.
         """
+        trigger_params = [p for p in self_.self_or_cls.param
+                          if (self_.self_or_cls.param[p].__class__.__name__ == 'Event')]
```
philippjfr, line 1496: Guess imports prevent you from doing `isinstance` but the name comparison doesn't seem entirely safe.

param/parameterized.py
jlstevens: Good point about using `set_param`, should be an easy fix. As for returning, I don't think that works as not all parameters being triggered are necessarily `Events`...unless you mean returning if `set(trigger_params) == set(param_names)` leaving an empty list for the rest of the method to process (which is fine).

philippjfr: Yes that's what I mean.

param/parameterized.py
jlstevens: Right, can't have cyclic imports. What do you think would be safer? Happy to take any suggestion..

That said, I don't really see how this could be problematic...the objects being processed have to be parameters and this is checking for parameters called `Event`. I suppose it could fail if you defined a different `Event` parameter but that seems like a bad idea regardless...

philippjfr:
I too prefer `Event` as a name. I get the point about events having a payload but on balance I still think it's clearer than Trigger.

param/parameterized.py
philippjfr: Can't control what parameters users define.

jlstevens: A user defining a parameter called `Event` when there is already `param.Event` is only going to cause confusion. Is there any reasonable use case for this?

param/parameterized.py
jlstevens: Addressed in 125c222

param/parameterized.py
jbednar: A user could very easily define any of our classes in their own code, if they wanted to change any of the behavior.  But the current behavior could either be desired nor not. E.g. if someone copies Event because they want it to act the same but have some additional functionality, naming it mymodule.Event would make this event-related processing still happen, which might be desired. But I still argue that duck typing is safer, i.e. to declare something about Event that says that it should have this special behavior, then respect that declaration in .trigger, rather than having it be based on the class name.  E.g. add one more slot and populate it only on Event with the high and low values to use during .trigger, or some other way that we can document for how a Parameter class selects this special behavior.

jlstevens: Sure, seems like a lot of effort to address a very rare use case but that would work!

jlstevens: I'm thinking of adding something like `_autotrigger_val` to the slots with value `True` in this case...

jbednar: It's less about addressing the use case than it is about making code in Parameterized that makes sense on its own, given that Event isn't defined there.

jlstevens: Implemented in 27f4e34

jlstevens:
Using `set_param` isn't working right and batching these properly looks like it is going to be a pain unfortunately.

jlstevens:
Ignoring all the other issues, even on master if you set up something like `@param.depends('c', watch=True)` and then doing `foo.c = 42` (or similar), both `_call_watcher` and `_batch_call_watchers` are invoked which is quite confusing (though only the unbatched is doing anything).

jlstevens:
Discussing this with Jim, we might want the current (messy) approach that implements the semantics we discussed (immediately switching back to False) but there is a possible alternative:

The `Event` parameter could have a slot called `reset_value`: only when some callback is actually triggered would this value be used to reset the parameter value to that slot value. Generally, it might be nice to have a  `__raw__set__` method which is like `__set_` but skips all the watcher stuff. Essentially only after calling a callback is `__raw_set__` used with `reset_value`.

This would have different semantics from what we discussed earlier. In particular, an `Event` parameter could have a permanent value of `True` if no watchers are hooked up to it. Only when something is actually triggered would it go back to `False`. This might be a way to notice that something is not hooked up if you type `foo.a = True` and then look at `foo.a` and it is still True, nothing was triggered.

The one issue is that setting to True might not do anything if the following things happen 1) an `Event` parameter is declared 2) it isn't hooked up to anything 3) it is set to True 4) it is dynamically hooked up to a callback. Now setting to `True` doesn't do anything as there is no parameter change.

@jbednar  @philippjfr Thoughts?

jbednar:
> Thoughts?

Personally, I like the simplicity of knowing that the only weird thing about Event is that when the event is processed the value is reset. It seems a lot easier to reason about than the approach currently in this PR, where I have to just close my eyes and hope that it's all ok.

>Edit: Would you change the value when calling .trigger? I think it has to in order to be useful. How would you know what value to use if there is only a reset value? I'm not sure the new semantics works...

You and Philipp agreed that the value needed to change when calling .trigger, so I guess you're right, you have to store both the reset and non-reset values still. :-(

jlstevens:
Even though we haven't decided whether to take a new approach or clean up the current one, I can report that the PR in its current state works with https://github.com/holoviz/panel/pull/1600 in a way I like (I can now easily tell which button was clicked!):

image

Personally, I don't see what the problem is with having a special boolean parameter that has some special handling to integrate with the watcher system. I'm not arguing that the code can't be cleaned up but I do think the approach isn't unreasonable:

1. Toggle the reset behavior off for event parameters at the start of `set_param`.
2. Set the parameter values (this step was always there, before this PR)
3. Reset the Event parameters after the watcher callbacks are invoked.
4. Restore the instant reset behavior of the event parameters.

Maybe we can find a nicer way to achieve the previously discussed behavior but unfortunately, I don't think a `reset_value` slot is sufficient (because it needs to signal the state in `.trigger`). As only two states are needed, I don't really see why you would need anything other than a boolean and why this mechanism ever needs to be generalized to any other parameters.

philippjfr:
There's a few things I'm struggling with the discussion here.

>This would have different semantics from what we discussed earlier. In particular, an Event parameter could have a permanent value of True if no watchers are hooked up to it. Only when something is actually triggered would it go back to False. This might be a way to notice that something is not hooked up if you type foo.a = True and then look at foo.a and it is still True, nothing was triggered.

This seems totally bizarre to me, to me the point of the `True` value is to say it is true that there is currently an event, so this totally breaks the semantics I've always maintained. Also changing the behavior depending on whether there is a watcher or not is also very strange to me. I'm totally missing any argument that would justify this behavior.

>I like the simplicity of knowing that the only weird thing about Event is that when the event is processed the value is reset. It seems a lot easier to reason about than the approach currently in this PR, where I have to just close my eyes and hope that it's all ok.

Totally disagree here, I think that is really weird behavior because it breaks the concept of this parameter signaling whether there is an event or not, but we've had sufficient back and forth here to establish our disagreement here 😄 

>You and Philipp agreed that the value needed to change when calling .trigger, so I guess you're right, you have to store both the reset and non-reset values still. :-(

This is the part I'm still working through so let's go through a chain of events. The main problem I see here is to do with batching of these events, which is a problem whether or not we implement the `.trigger` behavior, so I struggling to see how anything will be simpler without that. I also don't quite gather the need for storing a reset and non-reset values from this discussion. What I absolutely don't think will work is somehow making the resetting dependent on whether there are watchers, because if you add a watcher the value will be True and setting it will not trigger any event at all, which is simply wrong behavior, whichever semantics you ascribe to `param.Event`. It **always** has to reset to False.

That just leaves the resetting behavior in the batched which I'd have to work through with detailed examples but that doesn't at all depend on implementing the `.trigger` semantics.

jlstevens:
> What I absolutely don't think will work is somehow making the resetting dependent on whether there are watchers, because if you add a watcher dynamically the value will be True and setting it will not trigger any event at all, which is simply wrong behavior, whichever semantics you ascribe to param.Event. It always has to reset to False.

I agree. The semantics of the parameter just look like they need a more complicated implementation than we may like in order to make sense.

jbednar:
I think it's probably most efficient if @philippjfr and @jlstevens together agree on an implementation that you like.  I can give feedback at that point, but it won't be binding, just an opinion.  So please work it out so you're happy, and we can go from there!

param/parameterized.py
```diff
@@ -1428,6 +1428,13 @@ def set_param(self_, *args,**kwargs):
                 raise ValueError("Invalid positional arguments for %s.set_param" %
                                  (self_or_cls.name))
 
+        all_trigger_params = [p for p in self_.self_or_cls.param
+                              if hasattr(self_.self_or_cls.param[p],
+                                         '_autotrigger_value')]
+        trigger_params = [p for p in all_trigger_params if p in kwargs.keys()]
```
philippjfr, line 1434: Guess I'm missing why this isn't just:

```python
trigger_params = [k for k in kwargs if k in self_.self_or_cls.param and
                                hasattr(self_.self_or_cls.param[k], '_autotrigger_value')]
```

param/parameterized.py
```diff
@@ -1488,21 +1500,32 @@ def trigger(self_, *param_names):
...
         will be triggered whether or not the parameter values have
-        actually changed.
+        actually changed. As a special case, the value will actually be
+        changed for a Parameter of type Event, setting it to True so
+        that it is clear which Event parameter has been triggered.
         """
+        trigger_params = [p for p in self_.self_or_cls.param
+                          if hasattr(self_.self_or_cls.param[p], '_autotrigger_value')]
+        triggers = {p:self_.self_or_cls.param[p]._autotrigger_value
+                    for p in trigger_params if p in param_names}
```
philippjfr, line 1510: Same here:

```python
triggers = {p: self_.self_or_cls.param[p]._autotrigger_value
                    for p in self_.self_or_cls.param if p in param_names and
                    hasattr(self_.self_or_cls.param[p], '_autotrigger_value')}
```

Although this one is more debatable since that dictionary comprehension is getting a bit long.

philippjfr: Also can you remind me of the issue if you don't do this and do `set_param` for all of them at once? I'm a little worried about bypassing all this:

```python
        self_.self_or_cls.param._events  = []
        self_.self_or_cls.param._watchers = []
        self_.self_or_cls.param._TRIGGER = True
```

jlstevens: I need to remind myself exactly, but I believe I had to split out the trigger parameters from the other ones in other to avoid infinite recursion issues. I'll check whether it works without all these changes just to be sure.

param/parameterized.py
jlstevens: Good point. Implemented in ef290c0.

param/parameterized.py
jlstevens: Thanks for the suggestion, it looks like the original trigger method now works fine (the issues I was trying to work around must have been due to something else I was trying that is no longer needed). Reverted the trigger method implementation in cfb3518 as I don't see any issues with the original version.

As for the suggested dictionary comprehension, I do think it is a little too long to be worth changing.

jlstevens: Looking at the tests, it does seem my changes were necessary after all: the `assert self.e is True` fails in `test_watch_event_trigger_method` without them. Now I just need to trace through the code to remind myself why this is happening!

jlstevens:
@philippjfr Tests are green and I've addressed your suggestions. Ready to merge and cut a new release?

philippjfr:
Looks good.

PR #439, opened 2020-09-18
Fix .abstract for classes named with leading underscore(s)

ceball:
Fix the (longstanding) assumption in param's abstract class mechanism that abstract classes don't start with a leading underscore.

> Any identifier of the form __spam (at least two leading underscores, at most one trailing underscore) is textually replaced with _classname__spam, where classname is the current class name with leading underscore(s) stripped. This mangling is done without regard to the syntactic position of the identifier, as long as it occurs within the definition of a class.

(https://docs.python.org/2/tutorial/classes.html)

https://github.com/holoviz/param/blob/301b9617e84f3578364393d5ae30c452d1c19cd0/param/parameterized.py#L2074-L2094

(i.e. param does not do the same thing as python)

(originally from https://github.com/holoviz/param/pull/423#issuecomment-694526493)

Also added note about API0 and API1 tests (although not going so far as to say when you should be adding to one or the other or both, just noting the current duplication...)

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2087,9 +2087,9 @@ class ParameterizedMetaclass(type):
         # _ParameterizedMetaclass__abstract before running, but
         # the actual class object will have an attribute
         # _ClassName__abstract.  So, we have to mangle it ourselves at
-        # runtime.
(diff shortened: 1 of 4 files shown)


philippjfr:
Looks good to me!

jbednar (commented):
Looks good, thanks!

Issue #398, opened 2020-04-16
ObjectSelect objects key, value changes to value, key after update


MarcSkovMadsen:
#### My Pain

I'm trying to dynamically update the objects from which a user can select. But the behavior of `param.ObjectSelector.objects` is different before and after update.

```bash
$ python 'issue_object_selector.py'
Traceback (most recent call last):
  File "issue_object_selector.py", line 12, in <module>
    app.value=3
  File "C:\Users\masma\source\repos\orsteddcsmarketsanalytics\api\.venv\lib\site-packages\param\parameterized.py", line 294, in _f
    instance_param.__set__(obj, val)
  File "C:\Users\masma\source\repos\orsteddcsmarketsanalytics\api\.venv\lib\site-packages\param\parameterized.py", line 296, in _f
    return f(self, obj, val)
  File "C:\Users\masma\source\repos\orsteddcsmarketsanalytics\api\.venv\lib\site-packages\param\parameterized.py", line 823, in __set__
    self._validate(val)
  File "C:\Users\masma\source\repos\orsteddcsmarketsanalytics\api\.venv\lib\site-packages\param\__init__.py", line 1242, in _validate
    "valid options include %s"%(val,attrib_name, items))
ValueError: 3 not in Parameter value's list of possible objects, valid options include [Third, Fourth]
```

```python
import param
OBJECTS_ALL = {"First":1 , "Second": 2}


class App(param.Parameterized):
    value = param.ObjectSelector(default=1, objects=OBJECTS_ALL)

app=App()
app.value=2

app.param.value.objects = {"Third": 3, "Fourth": 4}
app.value=3
```

MarcSkovMadsen:
This is also strange. WHY DOES THE DICT BECOME A LIST OF CONSTRUCTION?

```bash
$ python 'issue_object_selector.py'
[1, 2]
2
{'Third': 3, 'Fourth': 4}
Third
(.venv)
```

```python
import param
import panel as pn
OBJECTS_ALL = {"First":1, "Second": 2}


class App(param.Parameterized):
    value = param.ObjectSelector(default=1, objects=OBJECTS_ALL)

    @param.depends("value", watch=True)
    def update(self):
        print(self.param.value.objects)
        print(self.value)

app=App()
app.value=2

app.param.value.objects = {"Third": 3, "Fourth": 4}
app.value="Third"
```

MarcSkovMadsen:
[...]

Issue #331, opened 2019-04-15
Dynamically adding to param.ObjectSelector after instantiation


kcpevey:
I have an object selector that I'd like to populate dynamically after instantiating the root class. I'll call it `poof` below, the other params are helpful for testing.   
  
```
class Simulation(param.Parameterized):
    bar = param.ObjectSelector(default=2, objects={'a':2, 'b':9})
    foo = param.ObjectSelector(default='f', objects=['f','g','h'])
    poof = param.ObjectSelector(objects=dict())

sim = Simulation()
pn.panel(sim)
```   
  
In some other bit of code, I add a value to `poof` :  
```
sim.param.poof.names['c'] = 99
sim.param.poof.objects.append(99)
```
  
In order for this to show up in my UI, I have to trigger a redraw with `pn.panel(sim)`. Which then shows that the `poof` widget now contains the `c` option on the UI. That's great.   
  
However, when I try to access the value of `poof` from the python side, it still shows up as empty, ie. `type(sim.poof) = NoneType`  
  
Is this a bug or is there a better way of going about this?

sdc50:
@kcpevey you don't mention it here, but assuming that you are reassigning the ``poof`` attribute on the ``sim`` object after you add the option to the ``poof`` param object:

```
sim.poof = 99
```

Then it looks like you get this error:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-5-ef65c8de2fa7> in <module>
      1 sim.param.poof.names['c'] = 99
      2 sim.param.poof.objects.append(99)# = list(sim.param.poof.names.values())
----> 3 sim.poof = 99

~/miniconda/envs/es-workflows/lib/python3.6/site-packages/param/parameterized.py in _f(self, obj, val)
    253         instance_param = getattr(obj, '_instance__params', {}).get(self.name)
    254         if instance_param is not None and self is not instance_param:
--> 255             instance_param.__set__(obj, val)
    256             return
    257         return f(self, obj, val)

~/miniconda/envs/es-workflows/lib/python3.6/site-packages/param/parameterized.py in _f(self, obj, val)
    255             instance_param.__set__(obj, val)
    256             return
--> 257         return f(self, obj, val)
    258     return _f
    259 

~/miniconda/envs/es-workflows/lib/python3.6/site-packages/param/parameterized.py in __set__(self, obj, val)
    784 
    785         for watcher in watchers:
--> 786             obj.param._call_watcher(watcher, event)
[...]

PR #440, opened 2020-09-29
Added support for setting ObjectSelector.objects

jbednar:
Addresses https://github.com/holoviz/param/issues/398 and possibly https://github.com/holoviz/param/issues/331 .

ObjectSelector was originally designed only to store a list of objects, which works fine if the objects are strings or have a .name attribute, but otherwise a GUI needs to work with names for the objects, and has to get them somewhere.  To support this case, ObjectSelector was previously extended to accept a dict of name,obj pairs as the `objects` argument to the constructor, but then extracted only the objects and stored it in `self.objects`, while storing the mapping in `self.names` (a lookup table that needs to be mappable both ways for GUI purposes, to give the label from the current value or the new value from a new label). This worked, but caused problems for anyone who wanted to dynamically adjust the objects after the Parameter was constructed.

With this PR, the behavior is the same if a user sets .objects after construction as it is during construction; in both cases the object values are extracted (now storing in ._objects) and the names mapping is stored in `self.names`.  This way a user can update the objects list whenever they like by passing a new dictionary.

It also supports cases where people have done a manual workaround like
```
        self.param.x.names = new_opts
        self.param.x.objects = list(new_opts.values())
```
by keeping the old names if a list is provided and the old names are compatible (have the same length).  This special case could be removed now, but it would affect people who had used the previously required workaround.

Note that this PR does _not_ address people trying to do `self.param.x.objects.append(...)`.  Anyone have ideas about whether that could be made to work?

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1208,8 +1208,9 @@ class ObjectSelector(SelectorBase):
             self._objects = list(value.values())
         else:
             self._objects = value
-            # Special case to allow names and objects to be set separately in any order
-            if self.names is not None and len(self.names) != len(value):
+            # Delete stale names mapping, if it doesn't cover the new object list
+            if len(value)==0 or (self.names is not None and 
+                                 not all([o in value for o in self.names.values()])):
                 self.names = None
 
     # CBNOTE: if the list of objects is changed, the current value for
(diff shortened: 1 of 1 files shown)


jbednar:
When evaluating this PR, I think a good counterproposal would be one that simply accepts that .objects will either be a dict or a list, and consistently treats it as such. That way a user could do .append safely as long as they never switched between dict or list, but it would still be confusing, and it's hard to see how to make it backwards compatible.

PR #441, opened 2020-10-05
Extend ObjectSelector to store a single obj dict

jbednar:
Alternative to https://github.com/holoviz/param/pull/440.  This version stores the objects for ObjectSelector and its subclasses as a string->obj dictionary in all cases, which makes the logic cleaner and easier but may not be able to be made backwards compatible.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -121,29 +121,61 @@ def hashable(x):
     else:
         return x
 
+def named_obj(obj, objtoname=None):
+    """
+    Return a name,obj pair for the given obj
+
+    Looks up the name in the provided object->name mapping, if any,
+    and otherwise introspects a string name.
+    """
+    if objtoname is not None and hashable(obj) in objtoname:
+        k = objtoname[hashable(obj)]
+    elif hasattr(obj, "name"):
+        k = obj.name
+    elif hasattr(obj, '__name__'):
+        k = obj.__name__
+    else:
+        k = as_unicode(obj)
+    return k,obj
+            
+
 def named_objs(objlist, namesdict=None):
     """
-    Given a list of objects, returns a dictionary mapping from
-    string name for the object to the object itself. Accepts
-    an optional name,obj dictionary, which will override any other
-    name if that item is present in the dictionary.
+    Return a dictionary name->object for the given objlist.
+
+    If objlist is already a dictionary, it is simply returned, but
+    otherwise a name->object dictionary is constructed, using a name
+    from namesdict (if provided) or otherwise an introspected name.
     """
-    objs = OrderedDict()
+    if objlist is None:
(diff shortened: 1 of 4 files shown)


jbednar:
Currently ignores self.name, but could potentially make it a property to update the objects dictionary to support some existing use cases.

jbednar:
Updated to provide backwards compatibility. With this PR, people should be able to use .objects freely as a dictionary, adding items to it or entirely replacing it with a string->object dictionary of the same type accepted in the constructor.  This approach should be backwards compatible with the previously required workarounds mentioned in #331 and #398, in that if self.objects is set to a list and self.names is set to a dictionary, the names set in self.names will be used for the objects. I.e., I think this approach gives good behavior for the future, while also supporting previous usage.  Examples:

image

jbednar:
Note that the type of self.objects is now always an OrderedDict, whereas before it was a list.  It still works to _set_ it to a list (via handling in the property), but I think it's important to return it as a dict so that people can use it naturally.  So there's a limitation to the backwards compatibility this PR can support -- anything that was accessing `p.objects` and treating it as a list will now need to access `list(p.objects.values)` instead.

@philippjfr, can you please think about the backwards compatibility issues involved, particularly for Panel users, and see if they are manageable?

param/__init__.py
```diff
@@ -1243,22 +1273,39 @@ def _validate(self, val):
...
+        self._names=value
+        self._objects=named_objs(self._objects, self._names)
+
+    @property
+    def objects(self):
+        return self._objects
+
+    @objects.setter
+    def objects(self, value):
+        self._objects=named_objs(value, self._names)
```
jbednar, line 1308: When `objects` is updated, it would be nice to update the current value of this parameter automatically to be something from the object list, using code like `param_value = named_obj_default(self.objects, self.default)`. This would make GUI code simpler; e.g. if one parameter selects a `continent` that then determines the objects list of a `country` parameter, then it would be nice to update the actual value of the country as well. I don't know of any way to do that, though, given that the current value is typically in the dictionary of a Parameterized object, not on this Parameter object, but maybe this object can trace some pointers back to the parameterized in question?

philippjfr:
This will be a major backwards compatibility breakage that I expect to affect a lot of people, not just Panel internally but we have plenty of examples iterating over `.objects` and users will have example updating `.objects`. I do think it's the right thing to do but I don't quite know on what the migration plan might be.

jbednar:
Is there anything we can do to smooth over any of the backwards compatibility issues?

philippjfr:
Really don't see how, I really only think we can safely do this in 2.0.

philippjfr:
Superseded by https://github.com/holoviz/param/pull/598

Issue #443, opened 2020-10-24
docstrings ordered by definition in class


ahuang11:
"Good question! Param was written for Python2, where the parameter order wasn’t preserved usefully, so we sorted them in various places. For Panel we now respect the definition order, and I think the Param docstring should do that now as well. I’m not sure whether this should be configurable or simply stop sorting altogether. My guess is that we don’t need a configuration option to control it, as we’ve never defined what the docstring behavior is, but maybe someone will complain if we don’t allow the old behavior to be restored. In any case I think it’s as simple as removing "sorted()` in ParamPager.param_docstrings, though there may be some other sorting elsewhere. Indeed, probably most “.sort()” and “sorted()” calls should be removed everywhere in the Param source code, because most are probably destroying user-created ordering. I’d appreciate seeing an issue on Param proposing removing such sorting, and/or a PR removing it specifically for the docstrings."

From jbednar
https://discourse.holoviz.org/t/is-there-a-way-to-make-docstring-follow-the-order-that-params-are-specified/1336/4

Or maybe sort based on setting the precedence value as slamer59 mentioned on the discourse.

jbednar:
I'd vote for both, which I think is what panel does; use definition order for the same precedence, but globally sort by precedence as well.

jbednar:
Fixed in #445

PR #445, opened 2020-11-14
Sort docstring by precedence

ahuang11:
https://github.com/holoviz/param/issues/443

### DIFF
--- a/param/ipython.py
+++ b/param/ipython.py
@@ -134,7 +134,14 @@ class ParamPager(object):
         (params, val_dict, changed) = info
         col_widths = dict((k,0) for k in order)
 
-        for name, p in params.items():
+        ordering = sorted(
+            sorted(params), # alphanumeric tie-breaker
+            key=lambda k: (- float('inf')  # No precedence is lowest possible precendence
+                           if params[k].precedence is None else
+                           params[k].precedence))
+
+        for name in ordering:
+            p = params[name]
             if only_changed and not (name in changed):
                 continue
 
@@ -172,13 +179,6 @@ class ParamPager(object):
             for col in info_dict[name]:
                 max_width = max([col_widths[col], len(info_dict[name][col])])
                 col_widths[col] = max_width
-
-        ordering = sorted(
-            sorted(info_dict), # alphanumeric tie-breaker
-            key=lambda k: (- float('inf')  # No precedence is lowest possible precendence
-                           if params[k].precedence is None else
-                           params[k].precedence))
-        info_dict = {k: info_dict[k] for k in ordering}
 
         return self._tabulate(info_dict, col_widths, changed, order, bounds_dict)
 
--- a/param/ipython.py
+++ b/param/ipython.py
@@ -127,9 +127,11 @@ class ParamPager(object):
 
 
     def sort_by_precedence(self, parameters):
-        """Sort the provided dictionary of parameters by their precedence value.
+        """
+        Sort the provided dictionary of parameters by their precedence value.
         In Python 3, preserves the original ordering for parameters with the
-        same precedence; for Python 2 sorts them lexicographically by name."""
+        same precedence; for Python 2 sorts them lexicographically by name.
+        """
         params = [(p, pobj) for p, pobj in parameters.items()]
         key_fn = lambda x: x[1].precedence if x[1].precedence is not None else 1e-8
         sorted_precedence = sorted(params, key=key_fn)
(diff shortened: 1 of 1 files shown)


param/ipython.py
```diff
@@ -134,7 +134,14 @@ def _build_table(self, info, order, max_col_len=40, only_changed=False):
...
 
-        for name, p in params.items():
+        ordering = sorted(
+            sorted(params), # alphanumeric tie-breaker
+            key=lambda k: (- float('inf')  # No precedence is lowest possible precendence
+                           if params[k].precedence is None else
+                           params[k].precedence))
+
+        for name in ordering:
+            p = params[name]
```
jbednar, line 144: Seems like this would overwrite the definition order?  I think it should respect the definition order unless explicitly overridden with prececence. Probably the best bet is to move the sorting code from Panel (https://github.com/holoviz/panel/blob/master/panel/param.py#L506) into a function that can be called here and from Panel. First step is to copy it here, then make it general, then Panel can be updated later.

ahuang11:
Okay thanks for the suggestion. It now follows: https://panel.holoviz.org/gallery/param/precedence.html
image

param/ipython.py
```diff
@@ -124,6 +126,21 @@ def param_docstrings(self, info, max_col_len=100, only_changed=False):
         return "\n".join(contents)
 
 
+    def _ordered_params(self, parameters):
```
jbednar, line 129: Maybe call it `sort_by_precedence`, because (at least for py3) compared to the dictionary that's passed in, the dict stays the same except for a global ordering by precedence. I don't think it should be private; other libraries (e.g. Panel) should be able to use this, so it shouldn't start with an underscore.  A docstring could be "Sort the provided dictionary of parameters by their precedence value. In Python 3, preserves the original ordering for parameters with the same precedence; for Python 2 sorts them lexicographically by name.".

jbednar: Otherwise, looks good!

ahuang11:
Any other changes needed?

jbednar:
Nope; looks great! I think it's ready to merge once the tests are passing. Not sure what's up with those, and probably won't have a chance to look at that today.

ahuang11:
I wonder if tests are failing because the order changed?
```
==================================
Parameters changed from their default values are marked in red.
Soft bound values are marked in cyan.
C/V= Constant/Variable, RO/RW = ReadOnly/ReadWrite, AN=Allow None
NameValue   Type     Bounds      Mode  
v    4    Number                C RW  
w    4    Number                C RO  
u    4    Number                V RW  
z    4    Number  (-1, 100)     V RW  
x   None  String              V RW AN 
y    4    Number  (-1, None)    V RW  
Parameter docstrings:
=====================
u: < No docstring available >
v: < No docstring available >
w: < No docstring available >
x: < No docstring available >
y: < No docstring available >
z: < No docstring available >
RAW STRING:
"\x1b[1;32mParameters of 'TestClass' instance\n==================================\n\x1b[0m\n\x1b[1;31mParameters changed from their default values are marked in red.\x1b[0m\n\x1b[1;36mSoft bound values are marked in cyan.\x1b[0m\nC/V= Constant/Variable, RO/RW = ReadOnly/ReadWrite, AN=Allow None\n\n\x1b[1;34mNameValue   Type     Bounds      Mode  \x1b[0m\n\nv    4    Number                C RW  \nw    4    Number                C RO  \nu    4    Number                V RW  \nz    4    Number  (-1, 100)     V RW  \nx   None  String              V RW AN \ny    4    Number  (-1, None)    V RW  \n\n\x1b[1;32mParameter docstrings:\n=====================\x1b[0m\n\n\x1b[1;34mu: < No docstring available >\x1b[0m\n\x1b[1;31mv: < No docstring available >\x1b[0m\n\x1b[1;34mw: < No docstring available >\x1b[0m\n\x1b[1;31mx: < No docstring available >\x1b[0m\n\x1b[1;34my: < No docstring available >\x1b[0m\n\x1b[1;31mz: < No docstring available >\x1b[0m"
FAIL
test_bad_compute_default (tests.API0.testlistselector.TestListSelectorParameters) ... WARNING:param.ParameterizedMetaclass: Use method 'params' via param namespace 
```

jlstevens:
> I wonder if tests are failing because the order changed?

I think that is possible. Otherwise, I like this PR and would be happy to see it merged once the tests are green!

philippjfr:
Will merge and fix up the tests separately.

Issue #455, opened 2021-02-08
Incorrect validation when calling super in ParameterizedFunctions


jlstevens:
I would expect the following pattern to work, where the baseclass generalizes the set of allowable parameter values while a subclass has a stricter definition:

```python
import param

class Foo(param.ParameterizedFunction):
    a = param.ClassSelector(default=['foo'], class_=(dict, list))
    def __call__(self, **params):
        self.p = param.ParamOverrides(self, params)
        return 1
    
class Bar(Foo):
    a = param.List(default=['bar'])
    def __call__(self, **params):
        return super().__call__(**params) + 1
    
Bar.instance(a=[4])
```

What seems to be happening is that `class_=(dict, list)` is being used by `List` to validate that the elements of the list are lists or dictionaries (which is obviously wrong).

jbednar:
That does seem like. a bug, but it's also illustrating bad practice to the extent that parameters are considered part of an object's declaration, since Bar does not satisfy "is a" Foo (since it doesn't accept all values of 'a' that Foo does).

jlstevens:
> it's also illustrating bad practice ... since Bar does not satisfy "is a" Foo (since it doesn't accept all values of 'a' that Foo does).

I agree that if you were to write `Bar` given the above definition of `Foo` you would match the declarations. 

The use case described in the PR above is when you generalize a base class and you don't know about existing users who may have written subclasses before that generalization was made. Those people should not have their subclasses broken as long as they specify values that are an allowed subset of the base class (but a warning would be appropriate, suggesting that they should update to the generalized declaration).

Even if we were to be dogmatic about not allowing subclasses to be more specific, the current behavior is just confusing.



In the HoloViews use case,

PR #456, opened 2021-02-08
Avoid clashing class_ slot in List parameter

jlstevens:
Here is one way to address https://github.com/holoviz/param/issues/455, namely to make sure the `List` parameter does not use the same `class_` slot as `ClassSelector` (I called it `list_class` for now instead).

Generally, I'm not sure why parameters inherit previous slot values at all (when they have a value of None) tbh...

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1364,7 +1364,7 @@ class List(Parameter):
     def __init__(self,default=[],class_=None,item_type=None,instantiate=True,
                  bounds=(0,None),**params):
         self.item_type = item_type or class_
-        self.class_ = class_
+        self.class_ = item_type or class_
         self.bounds = bounds
         Parameter.__init__(self,default=default,instantiate=instantiate,
                            **params)
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1359,11 +1359,11 @@ class List(Parameter):
     items in the list are checked to be of that type.
     """
 
-    __slots__ = ['bounds', 'list_class']
+    __slots__ = ['bounds', 'item_type']
 
     def __init__(self,default=[],class_=None,instantiate=True,
                  bounds=(0,None),**params):
-        self.list_class = class_
+        self.item_type = class_
         self.bounds = bounds
         Parameter.__init__(self,default=default,instantiate=instantiate,
                            **params)
@@ -1396,9 +1396,9 @@ class List(Parameter):
         self._check_type(val)
 
     def _check_type(self,val):
-        if self.list_class is not None:
+        if self.item_type is not None:
             for v in val:
-                assert isinstance(v,self.list_class),repr(self.name)+": "+repr(v)+" is not an instance of " + repr(self.list_class) + "."
+                assert isinstance(v,self.item_type),repr(self.name)+": "+repr(v)+" is not an instance of " + repr(self.item_type) + "."
 
 
 class HookList(List):
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1363,7 +1363,7 @@ class List(Parameter):
 
     def __init__(self,default=[],class_=None,item_type=None,instantiate=True,
                  bounds=(0,None),**params):
-        self.item_type = class_ or item_type
+        self.item_type = item_type or class_
         self.bounds = bounds
         Parameter.__init__(self,default=default,instantiate=instantiate,
                            **params)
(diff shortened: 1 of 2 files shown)


jlstevens:
This is probably the real issue: https://github.com/holoviz/param/issues/97

jlstevens:
The tests are passing and this is a fairly simple workaround for #97. Even then, I would argue there is a difference between `class_` in class selector and `value_class` for the values in collections (lists, sets, dicts etc) justifying a separate slot.

philippjfr:
I'd agree it justifies a different slot but as you say this is really a workaround for a deeper issue. Also not sure about `list_class` as the name, I'd suggest `item_type` but don't like that much more. In the long run I'd like towards making Parameters composable rather than having `item_type`-like slots.

jlstevens:
Happy to rename to `item_type` and I also agree that a compositional model would be better for types. After the renaming would you be happy with this PR due to the difference in semantics or would you say #97 needs to be addressed first (which is bound to be a much bigger job)?

philippjfr:
Maybe @jbednar has a better suggestion for the naming. I'd at least like to see `class_` to be aliased properly so whatever new name we decide on is also accepted as a kwarg.

jbednar:
#97 does need addressing, but I agree it's a much larger job than putting in a workaround here, so we should do both. Regarding this specific workaround, I agree that the type for the items in a collection is semantically different than the type for a selector, so having a separate slot is very reasonable.  The name `item_type` conveys to me that it's the type of the item and not of the collection, so I'm happy with that.

param/__init__.py
```diff
@@ -1359,11 +1359,11 @@ class List(Parameter):
     items in the list are checked to be of that type.
     """
 
-    __slots__ = ['class_','bounds']
+    __slots__ = ['bounds', 'item_type']
 
-    def __init__(self,default=[],class_=None,instantiate=True,
+    def __init__(self,default=[],class_=None,item_type=None,instantiate=True,
                  bounds=(0,None),**params):
-        self.class_ = class_
```
jlstevens, line 1366: @jbednar @philippjfr Think it is worth adding a warning to tell people to use `item_type` here instead if they set `class_` or would that be too annoying/noisy?

jbednar: Too annoying.

jlstevens:
@jbednar @philippjfr Happy with this workaround fix for now?

param/__init__.py
```diff
@@ -1359,10 +1359,11 @@ class List(Parameter):
...
     """
 
-    __slots__ = ['class_','bounds']
+    __slots__ = ['bounds', 'item_type', 'class_']
 
-    def __init__(self,default=[],class_=None,instantiate=True,
+    def __init__(self,default=[],class_=None,item_type=None,instantiate=True,
                  bounds=(0,None),**params):
+        self.item_type = item_type or class_
         self.class_ = class_
```
philippjfr, line 1367: For backward compatibility this should also reflect the `item_type`. Is it possible to simply make `class_` a property?

philippjfr: Just as an example, this change will already break the serialization code if a user uses item_type rather than `class_`.

jlstevens: > Is it possible to simply make `class_` a property?

I don't think so...I hit an issue when importing holoviews where `class_` as a slot was expected which is why I re-added it. What I should do is keep them in sync, so I'll push this change:

```python
self.class_ = item_type or class_
```

Which should fix the serialization issue.

jlstevens: Addressed in 9096e39

param/__init__.py
```diff
@@ -1359,11 +1359,12 @@ class List(Parameter):
...
 
-    __slots__ = ['class_','bounds']
+    __slots__ = ['bounds', 'item_type', 'class_']
 
-    def __init__(self,default=[],class_=None,instantiate=True,
+    def __init__(self,default=[],class_=None,item_type=None,instantiate=True,
                  bounds=(0,None),**params):
-        self.class_ = class_
+        self.item_type = item_type or class_
+        self.class_ = item_type or class_
```
philippjfr, line 1367:
```suggestion
        self.class_ = self.item_type
```

param/__init__.py
jbednar: Won't there still be an issue when deserializing between different param versions, such that class_ was serialized previously but gets deserialized only to class_ and not iterm_type?  Seems like there needs to be the same sort of logic in get_state.

jlstevens: Annoying but yes, good point!

jlstevens:
@philippjfr I see you've merged but what about Jim's point about pickling and `__getstate__` needing updating?

philippjfr:
Missed that :/

PR #467, opened 2021-05-05
Adds step to Range

### LINKED ISSUE #466: Add step to param.Range
It could be nice if `param.Range` has a `step` argument like `param.Number`.  This would make it possible for me to do this:
``` python
import param 
import panel as pn

    a = param.Range(default=(2, 5), bounds=(0, 10), step=0.2)  #  <-- I want to be able to do this
    b = param.Number(default=2, bounds=(0, 10), step=0.2)
    
pn.panel(Test())
```

Where right now, I need to do something like this to get the same effect:

``` python
    a = param.Range(default=(2, 5), bounds=(0, 10))
    b = param.Number(default=2, bounds=(0, 10), step=0.2)
    
    def __init__(self, **params):
        super().__init__(**params)
        self.widgets = pn.Param(self, widgets={"a": {"step": 0.2}})
    
pn.panel(Test().widgets)
```

Hoxbro:
Fixes #466

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1933,14 +1933,15 @@ class Color(Parameter):
 class Range(NumericTuple):
     "A numeric range with optional bounds and softbounds"
 
-    __slots__ = ['bounds', 'inclusive_bounds', 'softbounds']
+    __slots__ = ['bounds', 'inclusive_bounds', 'softbounds', 'step']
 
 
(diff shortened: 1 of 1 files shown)


jbednar (approved):
This looks perfect to me, thanks! @philippjfr , do you anticipate any problems with this for Panel?

philippjfr:
Looks good!

Issue #471, opened 2021-05-13
param.Color should accept named colors


philippjfr:
Currently param.Color only accepts HEX colors but in almost all cases I can think of where param.Color is being used valid named colors are also supported. Therefore I propose that we extend param.Color to also accept all named colors which are part of the CSS Extended color keywords spec:

https://www.w3.org/TR/css-color-3/#svg-color

(which also happens to be identical to the list of colors supported by SVG). [Matplotlib also supports these colors](https://matplotlib.org/stable/gallery/color/named_colors.html).

jbednar:
I think we should definitely allow that, but I'm not sure about the details.  

1. Would you propose that Param store the mapping from name to hex colors, returning the hex color when the attribute is accessed? That would ensure that the given name always means the same thing whenever it is a Parameter value, regardless of the consumer of that value, but it would mean that someone setting "red" then checking the current state of the Parameter would find "#FF0000" instead of "red", which could be confusing. It's also a bit odd to store such a list of hex values in Param, given that Param is such a general-purpose library, but it's common enough that it's not _crazy_ to do so. 
2. If we do have the hex values, we could consider not modifying the parameter value but instead providing a special method on the parameter that always gives a hex code computed from the current parameter value. That would be awkward to use, but would let the original value stay the same to avoid that sort of confusion.
3. If we don't store the hex values, are you proposing just storing the list of names, or simply accepting any string? If storing the names, presumably we're just validating that the supplied string is a valid color. At that point would Panel let you choose either a standard color selector or a Selection widget to select by named color?

Anyway, I have no objection to any of these options, just noting that they need decisions.

philippjfr:
I think 1. is an interesting proposal but would not really match what we do elsewhere, i.e. param generally returns the same thing you give it. So while I kind of like the idea I think it's overcomplicating it. For 2. I kind of have no opinion, adding such a method seems reasonable but you're right that it's probably too awkward to use in practice.

>At that point would Panel let you choose either a standard color selector or a Selection widget to select by named color?

Not sure about this one, I think the browser color selector is still the most versatile, but we can make a decision on that separately.

jbednar:
Ok, sounds good!

PR #472, opened 2021-05-13
Allow named colors on param.Color

philippjfr:
Implements #471

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1915,6 +1915,7 @@ class Color(Parameter):
     prefix.
     """
 
+    # CSS3 color specification https://www.w3.org/TR/css-color-3/#svg-color
     _named_colors = [ 'aliceblue', 'antiquewhite', 'aqua',
         'aquamarine', 'azure', 'beige', 'bisque', 'black',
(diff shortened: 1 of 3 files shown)


param/__init__.py
```diff
@@ -1915,18 +1915,63 @@ class Color(Parameter):
     prefix.
     """
 
-    def __init__(self, default=None, **kwargs):
+    _named_colors = [ 'aliceblue', 'antiquewhite', 'aqua',
```
jbednar, line 1918: Should have a comment pointing to the definitive source for these names.

tests/API1/testcolorparameter.py
```diff
@@ -37,3 +37,8 @@ class Q(param.Parameterized):
         Q.q = '#fff'
         self.assertEqual(Q.q, '#fff')
 
+    def test_valid_named_color(self):
+        class Q(param.Parameterized):
+            q = param.Color(allow_named=True)
+        Q.q = 'indianred'
+        self.assertEqual(Q.q, 'indianred')
```
jbednar, line 44: Would be good to test for an invalid color name when `allow_named=True`, as well.

jbednar (approved):

PR #475, opened 2021-05-13
Improved error messages and cleaned up code

philippjfr:
Some attempt at:

1. Improving error messages so they state what the wrong value actually was (this is not always obvious)
2. Clean up line lengths
3. Clean up of spacing between arguments.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -972,22 +972,12 @@ class Boolean(Parameter):
     def _validate(self, val):
         if self.allow_None:
             if not isinstance(val, bool) and val is not None:
-                raise ValueError("Boolean '%s' only takes a Boolean value "
-                                 "or None, not type %s."
-                                 % (self.name, type(val)))
-
-            if val is not True and val is not False and val is not None:
-                raise ValueError("Boolean '%s' must be True, False, or None, "
-                                 "not %s."% (self.name, val))
-        else:
-            if not isinstance(val,bool):
-                raise ValueError("Boolean '%s' only takes a Boolean value, "
-                                 "not type %s."
-                                 % (self.name, type(val)))
-
-            if val is not True and val is not False:
-                raise ValueError("Boolean '%s' must be True or False, "
-                                 "not %s." % (self.name, val))
+                raise ValueError("Boolean parameter %r only takes a "
+                                 "Boolean value or None, not %s."
+                                 % (self.name, val))
+        elif not isinstance(val, bool):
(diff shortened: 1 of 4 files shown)


param/__init__.py
```diff
@@ -956,25 +964,30 @@ class Boolean(Parameter):
...
             if val is not True and val is not False and val is not None:
-                raise ValueError("Boolean '%s' must be True, False, or None."%self.name)
+                raise ValueError("Boolean '%s' must be True, False, or None, "
+                                 "not %s."% (self.name, val))
         else:
             if not isinstance(val,bool):
-                raise ValueError("Boolean '%s' only takes a Boolean value."%self.name)
+                raise ValueError("Boolean '%s' only takes a Boolean value, "
+                                 "not type %s."
+                                 % (self.name, type(val)))
```
jbednar, line 986: Is the type more informative here than the value? Should it maybe show both (here and below)? Not sure.

param/__init__.py
```diff
@@ -1379,7 +1399,8 @@ def _validate(self, val):
             return
 
         if not isinstance(val, list):
-            raise ValueError("List '%s' must be a list."%(self.name))
+            raise ValueError("List '%s' must be a list, not %s."
```
jbednar, line 1402:
```suggestion
            raise ValueError("List '%s' must be a list, not an object of type %s."
```

jbednar (approved):

param/__init__.py
philippjfr: Yeah, I'm vacillating here. I think it should just show the value.

param/__init__.py
```diff
@@ -956,25 +964,30 @@ class Boolean(Parameter):
...
         if self.allow_None:
-            if not isinstance(val,bool) and val is not None:
-                raise ValueError("Boolean '%s' only takes a Boolean value or None."
-                                 %self.name)
+            if not isinstance(val, bool) and val is not None:
+                raise ValueError("Boolean '%s' only takes a Boolean value "
+                                 "or None, not type %s."
+                                 % (self.name, type(val)))
 
             if val is not True and val is not False and val is not None:
```
philippjfr, line 979: @jbednar This seems like dead code, am I wrong? The `not isinstance(val, bool) and val is not None` should catch this already, right?

jbednar: I agree. There does not seem to be any possible value of `val` for which this second clause would fire that would not already have fired for the first one.

PR #476, opened 2021-05-13
Update FileSelector parameters when setting path

philippjfr:
Adds an `_on_set` method to all Parameters, this allows Parameters to trigger actions when one of its attributes changes. This is used to update the list of valid paths when `(Multi)FileSelector.path` is set.

Fixes https://github.com/holoviz/param/issues/412

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -750,16 +750,16 @@ class Number(Dynamic):
 
     """
 
-    __slots__ = ['bounds','_softbounds','inclusive_bounds','set_hook', 'step']
+    __slots__ = ['bounds', 'softbounds', 'inclusive_bounds', 'set_hook', 'step']
 
-    def __init__(self,default=0.0,bounds=None,softbounds=None,
+    def __init__(self, default=0.0, bounds=None, softbounds=None,
                  inclusive_bounds=(True,True), step=None, **params):
         """
         Initialize this parameter object and store the bounds.
 
         Non-dynamic default values are checked against the bounds.
         """
-        super(Number,self).__init__(default=default,**params)
+        super(Number,self).__init__(default=default, **params)
 
         self.set_hook = identity_hook
         self.bounds = bounds
@@ -832,7 +832,7 @@ class Number(Dynamic):
         return val
 
     def _validate_bounds(self, val, bounds, inclusive_bounds):
-        if bounds is None:
+        if bounds is None or (val is None and self.allow_None):
             return
         vmin, vmax = bounds
         incmin, incmax = inclusive_bounds
@@ -1004,6 +1004,9 @@ class Tuple(Parameter):
                              "not %r." % (self.name, type(val)))
 
     def _validate_length(self, val, length):
+        if val is None and self.allow_None:
+            return
+
         if not len(val) == length:
             raise ValueError("Tuple parameter %r is not of the correct "
                              "length (%d instead of %d)." %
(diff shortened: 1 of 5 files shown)


jlstevens:
My main question is whether this mechanism might be better used to generate warnings instead of errors?

philippjfr:
Reverted any changes to validation on setting but still factored out separate validation methods for different attributes which should make it straightforward to readd the validation later.

jlstevens:
Makes sense. Ready for review?

philippjfr:
Yes, note this PR also continues some style and exception cleanup. Should have kept that separate sorry.

param/__init__.py
```diff
@@ -750,46 +750,37 @@ class Number(Dynamic):
 
     """
 
-    __slots__ = ['bounds','_softbounds','inclusive_bounds','set_hook', 'step']
+    __slots__ = ['bounds', 'softbounds', 'inclusive_bounds', 'set_hook', 'step']
```
jlstevens, line 753: Getting rid of the underscore makes sense as the properties didn't really do anything, but couldn't this have backward compatibility implications?

jlstevens: I suppose stateful things like pickling shouldn't touch slots directly and the API is the same as it was when properties were used. It *should* be ok but I am always wary when messing with slots in param.

param/__init__.py
```diff
@@ -837,61 +827,56 @@ def crop_to_bounds(self,val):
...
-        if self.step is not None and not _is_number(self.step):
+    def _validate_step(self, val, step):
+        if step is not None and not _is_number(step):
             raise ValueError("Step parameter can only be None or a "
-                             "numeric value, not type %r." % type(val))
-
-        self._checkBounds(val)
+                             "numeric value, not type %r." % type(step))
 
+    def _validate(self, val):
```
jlstevens, line 872: Good to see the types of validation performed laid out like this!

param/__init__.py
```diff
@@ -837,61 +827,56 @@ def crop_to_bounds(self,val):
 
         else:
             # non-numeric value sent in: reverts to default value
-            return  self.default
+            return self.default
 
         return val
 
+    def _validate_bounds(self, val, bounds, inclusive_bounds):
```
jlstevens, line 834: I like this: more general than `_checkBounds` and makes the input explicit using arguments instead of `self`.

param/__init__.py
```diff
@@ -1112,33 +1097,40 @@ class Composite(Parameter):
...
+
+    def _validate_attribs(self, val, attribs):
+        if len(val) == len(attribs):
+            return
+        raise ValueError("Compound parameter %r got the wrong number "
+                         "of values (needed %d, but got %d)." %
+                         (self.name, len(attribs), len(val)))
 
     def _validate(self, val):
-        assert len(val) == len(self.attribs),"Compound parameter '%s' got the wrong number of values (needed %d, but got %d)." % (self.name,len(self.attribs),len(val))
```
jlstevens, line 1133: Guess I've never seen this assert raise an exception so I guess it is safe to remove...

param/__init__.py
```diff
@@ -1315,30 +1305,36 @@ class ClassSelector(SelectorBase):
...
-                    (self.name, class_name, val))
+                    "%s parameter %r value must be an instance of %s, not %r." %
+                    (param_cls, self.name, class_name, val))
         else:
-            if not (val is None and self.allow_None) and not (issubclass(val,self.class_)):
+            if not (issubclass(val, class_)):
                 raise ValueError(
-                    "Parameter '%s' must be a subclass of %s, not '%s'" %
-                    (val.__name__, class_name, val.__class__.__name__))
+                    "%s parameter %r must be a subclass of %s, not %r." %
```
jlstevens, line 1336: I like this improvement to the exception message...

jlstevens:
Reviewing with PR was a little tricky with all the formatting changes but every code change looks like an improvement to me. Happy to see this merged!

param/__init__.py
philippjfr: Yeah, this is also in preparation to supporting validate methods on the class itself which could be useful for composite parameters.

param/__init__.py
philippjfr: Not removed, just moved to `_validate_attribs` as a proper `ValueError`

Issue #74, opened 2015-01-14
allow_None is only documented for Number


ceball:
allow_None also appears in Boolean and ClassSelector.

PR #478, opened 2021-05-17
Edit Param docstring 

ppwadhwa:
Include explanation of all parameters in the Param class docstring.

PR #479, opened 2021-05-18
Expanded Param docstring

jbednar:
Most of the Parameter attributes accepted were undocumented, but they are now documented in the constructor. Fixes https://github.com/holoviz/param/issues/74 . Supersedes #478.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -719,34 +719,82 @@ class Parameter(object):
 
     _serializers = {'json':serializer.JSONSerialization}
 
-    def __init__(self,default=None,doc=None,label=None,precedence=None,  # pylint: disable-msg=R0913
-                 instantiate=False,constant=False,readonly=False,
+    def __init__(self,default=None, doc=None, label=None, precedence=None,  # pylint: disable-msg=R0913
+                 instantiate=False, constant=False, readonly=False,
                  pickle_default_value=True, allow_None=False,
                  per_instance=True):
-        """
-        Initialize a new Parameter object: store the supplied attributes.
 
-        default: the owning class's value for the attribute
-        represented by this Parameter.
+        """Initialize a new Parameter object and store the supplied attributes:
 
-        precedence is a value, usually in the range 0.0 to 1.0, that
-        allows the order of Parameters in a class to be defined (for
-        e.g. in GUI menus). A negative precedence indicates a
-        parameter that should be hidden in e.g. GUI menus.
+        default: the owning class's value for the attribute represented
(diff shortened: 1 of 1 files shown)


param/parameterized.py
```diff
@@ -719,34 +719,82 @@ class Foo(Bar):
...
+        instances of that class or superclass). For a mutable
+        Parameter value, the default of False is also appropriate if
+        you want all instances to share the same value state, e.g. if
+        they are each simply referring to a single global object like
+        a singleton. If each Parameterized should have its own
+        independently mutable value, instantiate should be set to
+        True, but note that there is then no simple way to change the
+        value of this Parameter at the class or superclass level,
+        because each instance will now have an independently
+        instantiated value.
```
philippjfr, line 760: Changing the default will still affect instances created after making that change right? Not sure if that's worth clarifying though.

jbednar: True. Feel free to propose a change to say that!

jbednar: Each *existing* instance?

philippjfr: Yeah, that works.

param/parameterized.py
```diff
@@ -719,34 +719,82 @@ class Foo(Bar):
...
+        at the Parameterized class or superclass level (affecting all
+        instances of that class or superclass). For a mutable
+        Parameter value, the default of False is also appropriate if
+        you want all instances to share the same value state, e.g. if
+        they are each simply referring to a single global object like
+        a singleton. If each Parameterized should have its own
+        independently mutable value, instantiate should be set to
+        True, but note that there is then no simple way to change the
+        value of this Parameter at the class or superclass level,
+        because each instance will now have an independently
```
jbednar, line 759:
```suggestion
        because each existing instance will now have an independently
```

param/parameterized.py
```diff
@@ -719,34 +719,82 @@ class Foo(Bar):
...
+
+        instantiate: controls whether the value of this Parameter will
+        be deepcopied when a Parameterized object is instantiated (if
+        True), or if the single default value will be shared by all
+        Parameterized instances (if False). For an immutable Parameter
+        value, it is best to leave instantiate at the default of
+        False, so that a user can choose to change the value at the
+        Parameterized instance level (affecting only that instance) or
+        at the Parameterized class or superclass level (affecting all
+        instances of that class or superclass). For a mutable
```
jbednar, line 751:
```suggestion
        existing and future instances of that class or superclass). For a mutable
```

PR #480, opened 2021-05-24
make _add_param method public

### LINKED ISSUE #9: Should _add_parameter be public?

ppwadhwa:
I'm not sure if you want to do this, but I think that making _add_param public seems like a good idea.  If there is more/less to do here, let me know!  

Closes #9 .

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1438,7 +1438,7 @@ class Parameters(object):
         setattr(cls,param_name,value)
 
 
-    def _add_parameter(self_, param_name,param_obj):
+    def add_parameter(self_, param_name,param_obj):
         """
         Add a new Parameter object into this object's class.
 
@@ -2807,8 +2807,8 @@ class Parameterized(object):
 
     @classmethod
     @Parameters.deprecate
-    def _add_parameter(cls, param_name,param_obj):
-        return cls.param._add_parameter(param_name,param_obj)
+    def add_parameter(cls, param_name,param_obj):
+        return cls.param.add_parameter(param_name,param_obj)
 
     @bothmethod
     @Parameters.deprecate


param/parameterized.py
```diff
@@ -2807,8 +2807,8 @@ def state_pop(self):
 
     @classmethod
     @Parameters.deprecate
-    def _add_parameter(cls, param_name,param_obj):
-        return cls.param._add_parameter(param_name,param_obj)
+    def add_parameter(cls, param_name,param_obj):
+        return cls.param.add_parameter(param_name,param_obj)
```
jbednar, line 2811:
```suggestion
    def _add_parameter(cls, param_name,param_obj):
        return cls.param.add_parameter(param_name,param_obj)
```

This one is a deprecated method, so we should leave it with the old name until it's deleted.  But we do need to have it call the newly public method on param, which I've left it doing.

jbednar (commented):
Looks good! I've updated it not to change `self._add_parameter`, which is old API already deprecated. With that in mind, we should consider whether we need to provide backwards compatibility for `self.param._add_parameter`, which is private but not deprecated. 

On the one hand, we never promise to keep private API the same, and a Google search for "param._add_parameter" finds only [two](https://discourse.holoviz.org/t/add-dependencies-between-functions-and-a-dynamic-number-of-params/1030) [examples](https://discourse.holoviz.org/t/parameterized-class-update-on-mouse-up/2225) of such usage "in the wild". So we are certainly within our rights to make no provision for backwards compatibility.

On the other hand, it's clearly in use at least by a couple of people, and it's easy enough to add backwards compatibility, either by adding `_add_parameter=add_parameter` directly after the method definition or by making a deprecated wrapper with the old name. Either way leaves the codebase more confusing for the large number of users not hackily accessing `param._add_parameter`, so I think my vote is for not providing such compatibility here. Any other opinions from @jlstevens, @ceball , or @philippjfr ?

Issue #470, opened 2021-05-11
Serialization of JSON, YAML, INI/ConfigParser


sdrobert:
#### Is your feature request related to a problem? Please describe.
Serialization to simple config files (rather than pickled files) is limited to JSON, and support is only partial.

#### Describe the solution you'd like
Integration of [pydrobert-param](https://pydrobert-param.readthedocs.io/en/latest/index.html) functionality into param. It already supports such serialization, as well as a bunch of argparse and CLI stuff.

#### Describe alternatives you've considered
N/A

#### Additional context
I worked separately on this code base a while back because it seemed like it wasn't going in the direction that param was heading. With the serializer submodule, maybe this has changed? The codebase is [here](https://github.com/sdrobert/pydrobert-param) and would need some tailoring, but I'm happy to work on it a bit if it's worthwhile.

jbednar:
Cool! Would you be able to summarize what pydrobert-param provides that's not already in Param? There are definitely some useful things I'd like to see in Param:

- Serialization to YAML (though presumably easily achieved with Param's current JSON support?)
- Serialization to .ini (but could/should that be TOML instead?)
- Parsing command-line arguments as Parameters (though I can't quite see if that's what the argparse stuff does?)

For the Optuna stuff, I'd be happy to see some convenience function in param that gives the right kind of data structure that Optuna would need, but I wouldn't want any specific Optuna support to be in param itself. Optuna could be an example in the docs of how this data structure would be used, though! Hope that distinction makes sense.

sdrobert:
Hi and thanks for spending the time!

When I wrote the code circa 2019 (I believe), I wanted a way to seamlessly translate between config files and `Parameterized` instances. This meant a) writing/reading parameterized values to/from a dictionary; b) (de)serializing that dictionary to a popular config file type; and c) providing `argparse.ArgumentParser` hooks to handle that all transparently so that, by the time I'm done parsing the command-line arguments, my `Parameterized` instances have been populated.

[...]

PR #482, opened 2021-05-28
Support for deserialization of file types into Array, DataFrame

sdrobert:
This PR follows #382, #470 as well as discussion with @jbednar and @jlstevens.

param currently models deserialization routines after serialization routines. This is fine if you're only ever deserializing something that has previously been automatically serialized. However, my use case (and I believe that of plenty of others) focuses on reading in hand-written config files for e.g. specifying neural network parameters. In such cases, writing a Series, Array, or DataFrame could involve dumping a giant list of numbers by hand into the configuration file.

A better solution would be to specify a path to a data file in the configuration, which, when deserialized, is quietly parsed into the value. The reference to the file can be thrown away and just the value stored in the parameter. The type of file and thus the routine for parsing is inferred by the file extension. In the `deserialize()` method of the relevant functions, before interpreting the value as arguments to a constructor (either a `ndarray` or `DataFrame`), we: first check if it matches a file on disk; then if it matches an extension we know; and then, if Numpy or Pandas has a routine to read that extension, we call it.

This is not a bulletproof solution - some files will have misleading or no extensions, or might require non-standard arguments to parse - it will yield correct results in most situations. The user can always overload or subclass the relevant Parameters if she needs a special method of deserialization.

I have so far handled the easiest file types: those with both read and write routines in either numpy or pandas. Some require extra dependencies. If those dependencies were easily installed, I added guards in the test file and added them to the test environment in tox.ini. For numpy, these are

- .npy (Numpy archive)
- .txt[.gz] (Numpy text file)

For pandas, these are

- .csv (comma-separated)
- .dta (stata)
- .feather
- .json
- .ods (OpenOffice sheet)
- .parquet
- .pkl (pickle)
- .tsv (tab-separated)
- .xls{m,x} (Excel sheet)

Pandas has a lot more [I/O routines](https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html) that are trivial to add but harder to test. The two glaring absences are:

- .hdf5
- .xls (Pre-2007 Excel)

Pandas does have a write routine for ".xls" but its backend has been deprecated. HDF5 support relies on PyTables. PyTables is easily installed on Conda but has only a source distribution on PyPI. To install that source distribution, you need access to hdf5's headers.

I hope this is a good starting point for this functionality.

Thank you for your time,
Sean


philippjfr:
Nice! I like this a lot, are you also anticipating implementing the other direction and adding serialization?

sdrobert:
@philippjfr In my use case there isn't a need to serialize back to file. @jbednar and @jlstevens bandied the idea back and forth and I think came to the conclusion that, whereas deserialization can be performed transparently from file to value, serialization would need some mechanism to tether the Parameter value to a specific file/encoding. This PR as-is makes no changes to param's API (except perhaps removing the ability to deserialize an existing path as an array of characters - which was probably not intended in the first place).

jbednar:
Right; at this point it only covers deserialization, and fully supporting transparent roundtripping (ensuring the file type doesn't change in the process) sounds complicated. Still, I think we'll be able to review and merge this and add serialization later. I'd guess the filetype won't be preserved, e.g. a .csv might turn into .parquet, which is what Intake does for caching, but that seems ok to me.

param/__init__.py
```diff
@@ -282,6 +288,27 @@ def _get_min_max_value(min, max, value=None, step=None):
...
+        # transparently
+        ext = os.path.splitext(root)[1]
+    # If the extension does match, this is likely the correct routine and we will error
+    # later on. Let's warn the user.
+    try:
+        if ext in ext_to_routine:
+            return ext_to_routine[ext](path)
+    except:
+        get_logger().warning(
+            "Could not parse file '{}' as {}".format(path, type_name), exc_info=True)
```
jlstevens, line 307: One thing I think would be useful would be to state the supported extensions. Something like:

```python
"Could not parse file '{}' as {}. Supported extensions: {}".format(path, type_name, ', '.join(supported_exts))
```

The supported extensions would then probably be in some global registry by `type_name`.

jlstevens:
Looks great!

> This is not a bulletproof solution - some files will have misleading or no extensions, or might require non-standard arguments to parse - it will yield correct results in most situations. 

I realize this is still WIP but I've made one comment that I think would help users when the file fails to load: essentially, it would be nice to state which extensions *are* supported for that type.

param/__init__.py
sdrobert: This specific warning occurs when the user tried to parse an extension we do support, but the routine failed to do so (e.g. the file was invalid). How about just below?

param/__init__.py
```diff
@@ -282,6 +288,27 @@ def _get_min_max_value(min, max, value=None, step=None):
...
+    raise ValueError(
```
sdrobert, line 308: Here might be a better point for a warning stating the possible extensions. At this point we've verified that the path is indeed a file on disk, but it wasn't listed in our extension list. On that note, we can use the keys of the dictionary for the list of supported extensions

``` python
get_logger().warning("Cannot parse file '{}' as {}: unknown extension '{}'. Supported extensions: {}".format(path, type_name, ext, ', '.join(sorted(ext_to_routine))
```

Keeping track of the supported extensions somewhere globally could also make sense. Where would we put it?

sdrobert:
Thanks for the review jlstevens. Letting users know what file types are supported is a really good idea.

I also wanted to mention a quiet bug that I'm glossing over right now in case you want me to handle it differently. Python 2.7 supports only up to Pandas `0.24`. In that version, `pandas.read_excel` did not support `.ods` files. I am currently just skipping the `.ods` test for Python 2.7 and the package will incorrectly report the ability to handle `.ods` files. A more correct solution would involve checking the Pandas version and excluding the file type appropriately, or nixing the type altogether. That said, it's Python 2.7. I'm sure there are also minimum version requirements to Pandas (pre 1.0) and Numpy that I've overlooked as well.

param/__init__.py
jlstevens: > Keeping track of the supported extensions somewhere globally could also make sense. Where would we put it?

It could just be a global variable at the top level of the namespace. The tricky bit is naming it!

jlstevens:
>  I am currently just skipping the .ods test for Python 2.7 and the package will incorrectly report the ability to handle .ods files.

I think we can just mention this in the release notes. Even though param will probably support 2.7 for a while, many of our downstream projects are now switching to Python 3. At this point, it isn't critical if there are a few holes in the Python 2 support.

param/__init__.py
```diff
@@ -282,6 +288,27 @@ def _get_min_max_value(min, max, value=None, step=None):
...
+        "No deserialization routine for files with '{}' extension".format(path))
```
jbednar, line 309: I don't think we currently use the word "routine" in Param; probably clearer as:

```suggestion
        "No deserialization method for files with '{}' extension".format(path))
```

param/__init__.py
```diff
@@ -1461,8 +1488,15 @@ def serialize(cls, value):
...
-        return asarray(value)
+        import numpy
+        try:
+            return _deserialize_from_path(
+                {'.npy': numpy.load, '.txt': lambda x: numpy.loadtxt(str(x))},
+                value, 'Array'
+            )
+        except:
+            pass
+        return numpy.asarray(value)
```
jbednar, line 1499: So, what happens if numpy.load raises an exception and is unable to parse it? Will the user get any error messages? Seems like they would get an obscure message here about value not being suitable for numpy.asarray.

param/__init__.py
```diff
@@ -1544,8 +1578,26 @@ def serialize(cls, value):
 
     @classmethod
     def deserialize(cls, value):
-        from pandas import DataFrame as pdDFrame
-        return pdDFrame(value)
+        import pandas
+        try:
+            # FIXME(sdrobert): pandas.read_hdf5 requires pytables which can be
+            # installed no prob with conda but requires hdf5 headers for pip.
```
jbednar, line 1584: I'd be inclined to add HDF5 to the list anyway, and just put the installation info into the docs, because if the user _has_ installed it it will work, and that's all we can say about _any_ of these options...

tox.ini
```diff
@@ -41,17 +41,55 @@ deps = {[testenv]deps}
        gmpy
 setenv = PARAM_TEST_GMPY = 1
 
+# xlrd is the reader for xls files in pandas. xlrd is also the reader
+# for xlsx and xlsm files for pandas<1.2, but this is only possible using
+# xlrd<2. If pandas>=1.2 is guaranteed, you can remove the version spec
```
jbednar, line 46: I'd be ok with testing this new file-reading functionality only on pandas>=1.2; it's a new feature and not supporting it with old versions wouldn't be a big deal.

jbednar (commented):
Looks great! I agree that putting the mapping from file extension to reader functions would be nice to have at a global level, but those reader functions are in code that might not have been imported and may not even exist in this environment. Thus we'd have to store the mapping as text, which is doable but a bit ugly. So I'd be inclined to leave those mappings where they are, and use the keys() of the dict they are in to list the available file types.

param/__init__.py
sdrobert: They will be warned with the "Could not parse file {} as {}" bit.

tox.ini
sdrobert: Can do, but I think this'll nix python 3.6? I think they only go up to 1.1?

param/__init__.py
sdrobert: You're okay with not having an associated test?

param/__init__.py
sdrobert: It also prints the stack trace viz. "exc_info=True", which includes the original error.

jbednar (commented):
Can you provide some usage examples? At first I thought a user would be able to supply "file.csv" directly to the constructor of param.DataFrame, but that doesn't appear to be the case; it's only used on an explicit serialization. Can you show how that's meant to be used, in a form that could later be put in the docs?

param/__init__.py
sdrobert: > Ok, thanks. I think I see that better now. However, I can't seem to trigger a meaningful error message like that:
> 
> Somehow it seems like there is an uncaught exception rather than a logging message, but I can't spot where that is.

Ah, that's because we don't even try to deserialize if we can't find the disk on file - it just falls through to the previous behaviour.

```
(D:\conda-envs\param) PS D:\Documents\Projects\Code\param> echo "fiiii" > a.pkl
(D:\conda-envs\param) PS D:\Documents\Projects\Code\param> python -c "import param; param.DataFrame.deserialize('a.pkl')"
WARNING:param: Could not parse file 'a.pkl' as DataFrame
  File "D:\Documents\Projects\Code\param\param\__init__.py", line 304, in _deserialize_from_path
    return ext_to_routine[ext](path)
  File "D:\conda-envs\param\lib\site-packages\pandas\io\pickle.py", line 203, in read_pickle
    return pickle.load(handles.handle)  # type: ignore[arg-type]
_pickle.UnpicklingError: invalid load key, '\xff'.
  File "<string>", line 1, in <module>
  File "D:\Documents\Projects\Code\param\param\__init__.py", line 1600, in deserialize
    return pandas.DataFrame(value)
  File "D:\conda-envs\param\lib\site-packages\pandas\core\frame.py", line 590, in __init__
    raise ValueError("DataFrame constructor not properly called!")
ValueError: DataFrame constructor not properly called!
```

> In any case, it would be nice to eliminate the two levels of exception handling that make the code verbose and difficult to follow. Seems like this method could be simply be:
> 
> ```
>     def deserialize(cls, value):
>         return numpy.asarray(_deserialize_from_path(
>             {'.npy': numpy.load, '.txt': lambda x: numpy.loadtxt(str(x))},
>             value, 'Array'))
> ```
> 
> and either we allow the underlying method to raise a useful exception, or else we have it print a warning and then return a dummy value. I don't see what's gained by the extra try/except, but could easily be missing something.

I was trying to be as conservative as possible with the assumptions: "is it a string? a path to an existing file? ends in such and such extension? Did we successfully parse said file?" If the answer is no to any of these we fall back at the cost of more cryptic failures.

Looking again, I'm probably being too paranoid about it. "Is it a string?" is probably enough. I think the only possible valid array we're clobbering is a scalar numpy string array

``` python
>>> import numpy as np
>>> np.asarray("ab")
array('ab', dtype='<U2')
```
Which I'd imagine would be a rare intent.

param/__init__.py
jbednar: My vote is that a global dictionary like that is too tricky due to imports, and would either involve hackery (eval, etc.) or would be out of sync with what the code actually does. But I do think we should list the valid extensions.

param/__init__.py
jbednar: Yes, I think we can explicitly consider such an array not a valid final value for the Parameter.  Given that in any application where filenames and paths are being used an incorrect or outdated path is the most likely error by far, I think we should optimize the error reporting around (a) indicating that we are treating the provided string as a filename and (b) indicating that the file is not found. In that circumstance, there should be a simple, straightforward error message.  All other cases can be lumped together a bit, hopefully being clear but not needing to be simple or tightly controlled like that. Parse errors on pickle files do happen, but they are very rare compared to not finding the file at all!

param/__init__.py
jbednar: A test skipped by default sounds perfect.

sdrobert: Done

param/__init__.py
sdrobert: Ok, I've updated the numpy code to the below (pandas is similar):

``` python
    @classmethod
    def deserialize(cls, value):
        import numpy
        if isinstance(value, unicode):
            return _deserialize_from_path(
                {'.npy': numpy.load, '.txt': lambda x: numpy.loadtxt(str(x))},
                value, 'Array'
            )
        else:
            return numpy.asarray(value)
```

`_deserialize_from_path` doesn't return if it can't deserialize. This is as per the discussion above: if you've got a string and it's not a path, it's an error.

param/__init__.py
sdrobert: Ok, I've updated the messages. They also error instead of warning as per our other discussions. Right now the list of supported extensions isn't in some global list - I'll leave you two to come to a consensus about that.

sdrobert:
I think I've addressed all the comments above that I wasn't blocked on. I'm waiting on a consensus about the global list of extensions and the python 3.6/pandas 1.1 stuff.

param/__init__.py
jlstevens: The import issue does make it tricky so I guess leaving the listing implicit (i.e. as it is now) is fine.

maximlt:
@jlstevens do you think you'll have time to review this for 2.0 or prefer to postpone that?

jlstevens:
I think this probably should go in param 2.0 but it would be good to have a little time to test this change as well.

If you could fix the merge conflict, I'll do a quick review then merge.

maximlt:
@jlstevens I fixed the merge conflict.

sdrobert:
@maximlt Sorry for the trouble! It's been a busy few weeks.

maximlt:
> @maximlt Sorry for the trouble! It's been a busy few weeks.

Oh no worries :) We're pushing to get Param 2.0 out, hence the recent movement on this PR and others.

maximlt:
@jlstevens fixed the conflicts that were introduced recently after a few big merged PRs.

tox.ini
jbednar: I'm ok with not supporting Python 3.6, since we are trying to merge this into Param 2, our forward-looking codebase.

jbednar (changes requested):
I'm eager to get this merged at last, but I remain confused about the expected user-level API. What I was expecting is for a user to be able to replace `pd.read_csv("file.tsv")` with simply `"file.tsv"` in code like:

```python
import param, numpy as np, pandas as pd

class A(param.Parameterized):
    f = param.DataFrame(pd.read_csv("file.tsv"), rows=(2,None), columns=set(['a','b']))

a = A()
a.f
```

That doesn't work, and I'm not sure what the conditions should be for a user to invoke this functionality, since replacing `pd.read_csv("file.tsv")` with `param.DataFrame.deserialize("file.tsv")` doesn't seem like a net win:
image

Presumably we need some examples for the docs before merging?

param/__init__.py
```diff
@@ -314,6 +314,28 @@ def _get_min_max_value(min, max, value=None, step=None):
     return min, max, value
 
 
+def _deserialize_from_path(ext_to_routine, path, type_name):
+    """Call deserialization routine with path according to extension"""
```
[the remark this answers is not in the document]
sdrobert: @jbednar I suppose there's some ambiguity in the title. Deserialization could refer to what Pandas is doing, i.e. interpreting the contents of a file as a data frame. However, the deserialization this PR refers to is the deserialization `param` is doing, i.e. interpreting the contents of a JSON file as a Parameterized instance. Pandas' deserialization from file only occurs when the DataFrame parameter is assigned a string in the JSON file.

I think it could make sense to elevate this functionality so that any time a `param.DataFrame` parameter is passed a string, it treats it as a file path and tries to read it. I can't think of any reason off the top of my head why you shouldn't, nor why the functionality shouldn't be extended to `param.Array` parameters. The additional functionality could open up new questions which are a lot more complicated, particularly w.r.t. any persistent relationship to the file. For example, should a DataFrame initially read from a file be updated if the `param.DataFrame` instance is mutated? If one serializes a parameterized instance into JSON, then, does the file name get serialized as well? The answer to both may be "no," but I remember having discussions with @jbednar and I believe @jlstevens in a similar vein a few years back when I made this PR.

In any event, I believe I delivered on the functionality agreed upon and resolved all issues circa 2021. Though I'm happy to make small changes to conform to style, etc., I'd rather not be involved in developing more functionality at this time.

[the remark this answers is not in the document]
sdrobert: Ah, apologies. It looks as though @jbednar did ask about this in 2021. I missed it; that's my fault. If there's a consensus that this is what's wanted, I can spend some time revamping this functionality. Is it what's wanted?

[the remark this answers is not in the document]
jlstevens: I think there are lots of extra features that could be implemented around this but I also think that not all these extra features are necessarily desirable or worth the extra code complexity. My feeling is that the basic idea behind this PR makes sense and that the job then is to clearly document the known limitations. No need to let perfect be the enemy of good! (and issues such as detecting whether a dataframe has been mutated are tricky enough problems in their own right)

.github/workflows/test.yaml
```diff
@@ -38,7 +38,8 @@ jobs:
       fail-fast: false
       matrix:
         platform: ['ubuntu-latest', 'windows-latest', 'macos-latest']
-        python-version: ${{ ( github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' || ( github.event_name == 'push' && github.ref_type == 'tag' ) ) && fromJSON('["3.7", "3.8", "3.9", "3.10", "3.11", "pypy3.9"]') || fromJSON('["3.7", "3.9", "3.11"]') }}
+        # python-version: ${{ ( github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' || ( github.event_name == 'push' && github.ref_type == 'tag' ) ) && fromJSON('["3.7", "3.8", "3.9", "3.10", "3.11", "pypy3.9"]') || fromJSON('["3.7", "3.9", "3.11"]') }}
```
maximlt, line 41: Oops this one is on me, I forgot to reset that.

jbednar:
@sdrobert , even after all this time, I'm still missing a key bit of the intended use case and motivation, because there still aren't any examples of actual usage. As best I can tell, since there is no serialization implemented, what this PR will address is someone who writes their own JSON file and wants to specify a filename rather than the actual contents of the DataFrame or Array. Can you give us any example of an actual JSON file that would be used in this way? I consider JSON to be a read-only format, and would never edit it by hand since that just leads to file-format errors, but I understand that editing JSON can be feasible for some people in an editor like VSCode that has better support than what I use. Is that really the intended use case? Directly authoring JSON? If so we need to include an example in the docs of doing that, or no one will ever use this functionality.

sdrobert:
@jbednar, to answer your immediate question, my use case has always been machine learning. You can find an example [here](https://pydrobert-param.readthedocs.io/en/latest/serialization.html) in my supplementary library to `param` (N.B. this is not a plug; I would rather all the functionality exist in `param` so I could sunset my library).  This example does not contain any arrays or data frames, but may easily and plausibly be augmented to include one, e.g.

``` json
{
  "training": {
    "lr": 1e-05,
    "max_epochs": 10,
    "model_regex": "model-{epoch:05d}.pkl"
  },
  "model": {
    "activations": "relu",
    "layers": [
      "conv",
      "conv",
      "fc"
    ],
    "mean": "mean.npy",
    "std": "std.npy"
  }
}
```

Where `mean.npy` and `std.npy` point to files where, e.g., feature means and standard deviations reside. When deserializing from file, the mean and std params are populated with the contents of the file. This configuration can be specified by hand; other methods would be tedious/impossible.

DataFrame parameters could provide an easy means of dynamically specifying training sets for smaller ML tasks, e.g. for [scikit-learn](https://scikit-learn.org/stable/index.html) routines. It could also be used similarly to script visualization routines for e.g., [seaborn](https://seaborn.pydata.org/) or possibly HoloViz, avoiding notebooks.

I'm not really sure how to answer the other question about modifying JSON by hand. I agree that JSON is unwieldy, which is why I have also implemented YAML (de)serialization in my repo and [am willing to make a PR for it here](https://github.com/holoviz/param/issues/520#issuecomment-1254059750). Based on the framing of the question, however, it seems perhaps that the JSON (de)serialization mechanism in `param` is intended solely for machine consumption? There doesn't appear to be a standard means of getting parameters into a `Parameterized` instance beyond manipulating them after instantiation programmatically. This is a broader question than that of just arrays and data frames. If the team doesn't see much value in ingesting manual configurations overall, this PR isn't going to do much. I'm sorry for the noise at that point.

jbednar:
That's perfect, thanks! Indeed, we do see much use for a declarative YAML spec, and now it makes sense. This PR is only half of what's needed, which is what was confusing me! Ok, now we can move forward. Thanks.

.github/workflows/test.yaml
maximlt, line 41:
```suggestion
        python-version: ${{ ( github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' || ( github.event_name == 'push' && github.ref_type == 'tag' ) ) && fromJSON('["3.7", "3.8", "3.9", "3.10", "3.11", "pypy3.9"]') || fromJSON('["3.7", "3.9", "3.11"]') }}
```

maximlt:
Given that you've reviewed this extensively @jbednar, I'll let you decide whether you want to hit merge.

PR #487, opened 2021-06-16
Param Promo Example

MarcSkovMadsen:
Used for solving https://github.com/holoviz/param/issues/486#issuecomment-862041017

Issue #486, opened 2021-06-13
Ideas for improving the new documentation


MarcSkovMadsen:
The upcoming revised documentation is a great improvement. I've already learned a lot and believe I understand Param and the ideas behind it much better. It's great.

The below is a list of ideas that I believe could improve the documentation for both new and experienced users.

## 1. Make it even more visible WHY `param` would be valuable to "me".

The documentation is text heavy. I think that is great as soon as my attention has been caught and I understand why I should take the time to read and learn. But the text is great.

My suggestion would be to add a 3 min video to the `Introduction` that explains the why with a real world, practical example. It should be a walk through of a very interactive Notebook that can be downloaded by the user right next to the video.

## 2. Make the first Getting Started Example even more friendly for users.

The first example I see looks like this. 

```python
import param
        
class A(param.Parameterized):
    title = param.String(default="sum", doc="Title for the result")
    
class B(A):
    a = param.Integer(2, bounds=(0,10), doc="First addend")
    b = param.Integer(3, bounds=(0,10), doc="Second addend")
    
    def __call__(self):
        return self.title + ": " + str(self.a + self.b)
```

```python
>> o1 = B(b=4, title="Sum")
>> o1.a = 5
>> o1()
'Sum: 9'
```

The main question for me is whether using `__call__` in the example is a good idea. I know about `__call__` but I never see any colleagues or blogs in data, engineering, trading or science use it. So I believe the big questions for users is: Why do you use `__call__`. I did not know I needed `__call__` in my code. What is the use case for `__call__`? Why is `__call__` useful to me?

## 3. Add Pydantic to the Comparison

Pydantic is taking the Python world with storm. Together with FastApi. I why like to know why it's valuable to me to build domain models with Param when

1. I can use Pydantic and not Param with FastAPI.
2. Pydantic works using type annotations which provides a much better development experience in an IDE than param.
3. The huge popularity of Pydantic means more colleagues would probably know and understand why I use Pydantic
4. There are (more) tools out there for Pydantic like PyCharm, MyPy, Hypothesis plugins. And more are probably coming.
5. Lots of orgs are building on Pydantic and what to be associated with it. See https://github.com/Kludex/awesome-pydantic

I believe Param and Pydantic could be integrated and that would help a lot of Pydantic users easily create visualizations, tools, apps and dashboards from their models. And Param users create modern Rest Apis via Fast Api. For a Streamlit+Fast API example check out https://testdriven.io/blog/fastapi-streamlit/.

[...]

PR #488, opened 2021-06-17
Promotion Initiative

MarcSkovMadsen:
This PR adds some of the ideas from https://github.com/holoviz/param/issues/486 like a Promo Notebook with an associated Video.

Replaces https://github.com/holoviz/param/pull/487

FYI @ea42gh @jbednar


jbednar:
Thanks, @MarcSkovMadsen! I've pushed some minor cleanups, but otherwise it looks great! 

The notebook wasn't formatting properly in classic Jupyter (all the logos were showing up in one vertical line). I fixed that by putting everything in a table, which works well in classic Jupyter; hope that also works in JupyterLab (and hope nothing else I did messed up the incremental display for JupyterLab).

The example looks good. Some notes:
1. It's annoying that the Parameterized class needs a constructor, because very often a Parameterized class does not, and this one is about the most complex code being shown here. I can't think of how to avoid it if we want c to be initialized at first, though. That may be a small hole in Param that we could fill somehow, i.e. add `init=True` alongside `watch=True` to ensure that the dependent code is run during initialization.
2. Do we need to cover the bit with a separate watcher?  I'm not sure that most users will be able to see why that's relevant to them immediately, and I think the limit for a video like this is about 5 minutes, so it would be good to find something to cut.
3. It's a shame to have to include try/except in the first intro video; not everyone is comfortable catching exceptions, and it's quite wordy. In my other docs I'm using a context manager `with exceptions_summarized()` for that reason, which is shorter, but then that has to be imported and people aren't necessarily comfortable with context managers either. I guess that's another hole in Param; we should have some variable we can set that gives us short exceptions printing, for cases like this.
4. There are lots of other things Param provides; obviously we can't include all of them, so this is as good a set as any to start with, I suppose!

ea42gh:
Fortunately, folding works in both jupyter classic and jupyter lab.
In classic, one needs to enable an nbextension...

maximlt:
ipython has a `%xmode` magic command that is a switch affecting the length of tracebacks displayed. Its minimal mode displays only one line. See its [doc](https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-xmode). That's a potential alternative to the context manager or the try ... except block.

jbednar:
True; that could work here! Thanks. I don't think it would help for the docs in general, because the docs need some way to indicate that the exception is expected and that execution should not halt.

ea42gh:
I took my copy of Promo.ipynb and used %xmode
The video will be cleaner that way, but I would leave the original version with the
block to catch the exception...

jbednar:
Looks good; merging!

PR #494, opened 2021-06-27
Fixed List validation

### LINKED ISSUE #491: `param.List` doesn't respect `bounds` and `class_`
This snippet works fine while it should raise errors (for l1, l2 and l3):
```python
import param

class A(param.Parameterized):
    l1= param.List([0, 1], bounds=(2, 4))
    l2 = param.List([0], bounds=(2, 4))
    l3 = param.List(["a"], class_=str)

a = A()
a.l1 = [0]
a.l3 = [1]
```

This is because `List._validate_bounds()` and `List._validate_item_type` are never called.

(Which attribute between `class_` or `item_type` should be used?)

jbednar:
Fixed validation for List parameters, which was broken in https://github.com/holoviz/param/commit/bf3d324ddaa597a64585177662e0d9e6287a2ece because the new `_validate_bounds()` and `_validate_item_type()` methods weren't called by the default implementation of `_validate()`. Closes #491.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1095,6 +1095,9 @@ class Composite(Parameter):
     in the order specified.  Likewise, setting the parameter takes a
     sequence of values and sets the value of the constituent
     attributes.
+
+    This Parameter type has not been tested with watchers and 
+    dependencies, and may not support them properly.
     """
 
     __slots__ = ['attribs', 'objtype']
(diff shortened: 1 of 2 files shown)


param/__init__.py
```diff
@@ -1376,6 +1383,15 @@ def __init__(self, default=[], class_=None, item_type=None,
                            **params)
         self._validate(default)
 
+    def _validate(self, val):
+        """
+        Checks that the value is numeric and that it is within the hard
+        bounds; if not, an exception is raised.
+        """
```
jbednar, line 1390: Oops; left a useless docstring in from where I copied this.  I'll delete in a later PR.

PR #495, opened 2021-06-27
Fixed Path to allow folders, as already documented

jbednar:
Path is documented as accepting both files and folders (directories), but it was only accepting files.  Fixed to make Path check only that the path exists, not whether it is a file or folder. Also simplified the logic overall. Marked API since the behavior has apparently never matched the documented API; now it should match the previously documented API.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1628,45 +1628,39 @@ class resolve_path(ParameterizedFunction):
         Prepended to a non-relative path, in order, until a file is
         found.""")
 
-    path_to_file = Boolean(default=True, pickle_default_value=False, doc="""
+    path_to_file = Boolean(default=True, pickle_default_value=False,
+                           allow_None=True, doc="""
         String specifying whether the path refers to a 'File' or a 'Folder'.""")
 
     def __call__(self, path, **params):
         p = ParamOverrides(self, params)
-
         path = os.path.normpath(path)
+        ftype = "File" if p.path_to_file is True \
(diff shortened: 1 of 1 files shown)


param/__init__.py
```diff
@@ -1628,45 +1628,39 @@ class resolve_path(ParameterizedFunction):
         Prepended to a non-relative path, in order, until a file is
         found.""")
 
-    path_to_file = Boolean(default=True, pickle_default_value=False, doc="""
+    path_to_file = Boolean(default=True, pickle_default_value=False,
+                           allow_None=True, doc="""
         String specifying whether the path refers to a 'File' or a 'Folder'.""")
```
jlstevens, line 1633: I was wondering how it could be anything other than `True` or `False` but looking at the code I see that `None` means *either* a file *or* a directory. I'll clarify this in a small PR.

Minor nitpick but I associate 'folder' with Windows terminology while I consider 'directory' to be the proper platform-independent terminology. Probably not worth updating as everyone knows what you mean!

jlstevens: Clarified docstring in #507. I won't mark this as resolved in case you have any opinion on 'Folder' versus 'Directory'.

jbednar: Send like the do string is still confusing as it is a boolean, not a strong as claimed. Thanks for the None clarification. I agree with your preference for "directory" over "folder", but I am just following the existing param.Filename and param.Foldername terminology.

Issue #307, opened 2018-12-07
(Multi)FileSelector cannot set check_on_set=True with non-empty/null default


sdrobert:
Hi,

Thanks again for the great package. As it stands, Selector initialization happens before the call to `update()`. ``FileSelector`` and ``MultiFileSelector`` can't actually set any non-empty defaults when ``check_on_set=True`` because ``objects`` has yet to be populated. To reproduce:

``` sh
echo "
import param

class Foo(param.Parameterized):
  a = param.FileSelector('foo.py', path='*', check_on_set=True)

" > foo.py
python foo.py
```

Sean

tonyfast:
I can reproduce problem and i receive the following error message.

```
~/miniconda3/lib/python3.7/site-packages/param/__init__.py in _validate(self, val)
   1240             items = '[' + ', '.join(items) + limiter
   1241             raise ValueError("%s not in Parameter %s's list of possible objects, "
-> 1242                              "valid options include %s"%(val,attrib_name, items))
   1243 
   1244     def _ensure_value_is_in_objects(self,val):

ValueError: foo.py not in Parameter None's list of possible objects, valid options include []
```

jbednar:
Fixed in #497

PR #497, opened 2021-06-28
Replace ObjectSelector with Selector

### LINKED ISSUE #493: Selectors should inherit from Selector, not ObjectSelector
ObjectSelector is old, deprecated API, replaced with Selector. However, FileSelector, ListSelector, and MultiFileSelector still inherit from ObjectSelector, making it difficult to remove and making those classes confusing to work with.  Swapping ObjectSelector and Selector in the class hierarchy shouldn't affect anything not testing those exact classes, and we should do that soon to avoid future confusion.

jbednar:
Fixes #493

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1153,7 +1153,7 @@ class Selector(SelectorBase):
     """
     Parameter whose value must be one object from a list of possible objects.
 
-    By default, if no default is specfied, picks the first object from 
+    By default, if no default is specfied, picks the first object from
     the provided set of objects, as long as the objects are in an ordered
     data collection.
 
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1181,8 +1181,8 @@ class Selector(SelectorBase):
     # Selector is usually used to allow selection from a list of
     # existing objects, therefore instantiate is False by default.
     def __init__(self, objects=None, default=None, instantiate=False,
-                  compute_default_fn=None, check_on_set=None,
-                  allow_None=None, empty_default=False, **params):
+                 compute_default_fn=None, check_on_set=None,
+                 allow_None=None, empty_default=False, **params):
 
         if is_ordered_dict(objects):
             autodefault = list(objects.values())[0]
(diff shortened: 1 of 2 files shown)


jbednar:
Note that ListSelector, FileSelector, and MultiFileSelector all still use the convention from ObjectSelector where the first argument is the default value, not the objects list, and where the default value must be specified since it won't be computed from the objects list. Changing those things breaks backwards compatibility, so shouldn't be done lightly.

jbednar:
Added a fix for https://github.com/holoviz/param/issues/307 .

param/__init__.py
```diff
@@ -1149,10 +1149,14 @@ def get_range(self):
         raise NotImplementedError("get_range() must be implemented in subclasses.")
 
 
-class ObjectSelector(SelectorBase):
+class Selector(SelectorBase):
     """
     Parameter whose value must be one object from a list of possible objects.
 
+    By default, if no default is specfied, picks the first object from
```
jlstevens, line 1156: Typo: `specfied -> specified`. I'll make a new PR with these minor fixes now this has been merged.

param/__init__.py
```diff
@@ -1174,11 +1178,27 @@ class ObjectSelector(SelectorBase):
 
     __slots__ = ['objects', 'compute_default_fn', 'check_on_set', 'names']
 
-    # ObjectSelector is usually used to allow selection from a list of
+    # Selector is usually used to allow selection from a list of
     # existing objects, therefore instantiate is False by default.
-    def __init__(self, default=None, objects=None,  instantiate=False,
+    def __init__(self, objects=None, default=None, instantiate=False,
```
jlstevens, line 1183: This order is indeed better but in theory could break anyone passing in the arguments by position. This causing an issue is very unlikely though as I don't think I've ever seen that style of signature so overall this is better.

jlstevens:
Looks good! 

Sorry for not reviewing before merge but I don't see anything I would have objected to (other than a minor typo fix that I'll push a PR for now).

param/__init__.py
jlstevens: Fixed in #507

Issue #396, opened 2020-04-15
Limit recursion in reprs to avoid stack overflow for circular references


MarcSkovMadsen:
# System Info
- param 1.9.3
- pytest 5.3.1
- Windows 8.1

#### My Pain

I'm developing a component based on `Param` and `Panel` and would like to test it with `pytest`. During development my tests stopped working/ running but without errors. After some debugging I discovered it was due to a stackoverflow that pytest did not report. It just stopped.

I've developed a smaller code example below. I've also included the stack trace. 

#### Stack Trace

```bash
$ python 'test_issue_multple_param_action.py'
Fatal Python error: Cannot recover from stack overflow.

Current thread 0x00001be8 (most recent call first):
  File "C:\Users\masma\source\repos\orsteddcsmarketsanalytics\api\.venv\lib\site-packages\param\parameterized.py", line 1535 in get_param_values
  File "C:\Users\masma\source\repos\orsteddcsmarketsanalytics\api\.venv\lib\site-packages\param\parameterized.py", line 2408 in __repr__
  File "C:\Users\masma\source\repos\orsteddcsmarketsanalytics\api\.venv\lib\site-packages\param\parameterized.py", line 2408 in <listcomp>

..........

  File "C:\Users\masma\source\repos\orsteddcsmarketsanalytics\api\.venv\lib\site-packages\param\parameterized.py", line 2408 in <listcomp>
  File "C:\Users\masma\source\repos\orsteddcsmarketsanalytics\api\.venv\lib\site-packages\param\parameterized.py", line 2408 in __repr__
  ...
(.venv)
```

#### Code

```python
import param

class InputDataComponent(param.Parameterized):
    update_data = param.Action()
    download_data = param.Action()

    def __init__(self, **params):
        super().__init__(**params)

        self.update_data = self._update_data
        self.download_data = self._download_data

    def _update_data(self, event):
        raise NotImplementedError()

    def _download_data(self, event):
        raise NotImplementedError()

repr(InputDataComponent())
```

MarcSkovMadsen:
#### Additional Context

It's no problem with only one `param.Action`

```bash
$ python 'test_issue_multple_param_action.py'
(.venv)
```

#### Code

```python
import param

class InputDataComponent(param.Parameterized):
    update_data = param.Action()
    # download_data = param.Action()

    def __init__(self, **params):
        super().__init__(**params)

        self.update_data = self._update_data
        # self.download_data = self._download_data

    def _update_data(self, event):
[...]

PR #499, opened 2021-06-30
Avoid recursive repr issues

jbednar:
Addresses #209 and #396 to avoid infinite recursion when printing a Parameterized object that contains an indirect reference to itself. 

First tried using `reprlib.recursive_repr`, which seemed to work for repr itself, but did not address `pprint`, because it did not allow the decorated method to have any extra arguments. Luckily the implementation is straightforward, so I copied it from https://github.com/python/cpython/blob/3.2/Lib/reprlib.py and changed it to handle args and kwargs.  I did have to change `from _thread import get_ident` to avoid private API, but it seems the same as `from threading import get_ident`, which appears to work the same.

I haven't run the test suite yet, but it addresses #209:

```python
import param
class A(param.Parameterized):
    
    b = param.Parameter()
    
class B(param.Parameterized):
    
    a = param.Parameter()
    
a = A()
b = B(a=a)
a.b = b
a.pprint()
```
'A(b=__main__.B(a=...))`

and #396:

```python
import param

class InputDataComponent(param.Parameterized):
    update_data = param.Action()
    # download_data = param.Action()

    def __init__(self, **params):
        super().__init__(**params)

        self.update_data = self._update_data
        # self.download_data = self._download_data

    def _update_data(self, event):
        raise NotImplementedError()

    def _download_data(self, event):
        raise NotImplementedError()

repr(InputDataComponent())
```
"InputDataComponent(name='InputDataComponent00004', update_data=<bound method InputDataComponent._update_data of ...>)"

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2504,10 +2504,34 @@ script_repr_reg[FunctionType]=function_script_repr
 dbprint_prefix=None
 
 
-
-
-
-
+# Copy of Python 3.2 reprlib's recursive_repr but allowing extra arguments
+if sys.version_info.major >= 3:
+    from threading import get_ident
+    def recursive_repr(fillvalue='...'):
+        'Decorator to make a repr function return fillvalue for a recursive call'
+    
+        def decorating_function(user_function):
+            repr_running = set()
+    
+            def wrapper(self, *args, **kwargs):
+                key = id(self), get_ident()
+                if key in repr_running:
+                    return fillvalue
+                repr_running.add(key)
+                try:
+                    result = user_function(self, *args, **kwargs)
+                finally:
+                    repr_running.discard(key)
+                return result
+            return wrapper
+    
+        return decorating_function
+else:
+    def recursive_repr(fillvalue='...'):
+        def decorating_function(user_function):
+            return user_function
+        return decorating_function
+    
 
 @add_metaclass(ParameterizedMetaclass)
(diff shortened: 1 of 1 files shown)


param/parameterized.py
```diff
@@ -2504,9 +2504,33 @@ def type_script_repr(type_,imports,prefix,settings):
...
+            return wrapper
 
-
-
+        return decorating_function
+else:
+    def recursive_repr(fillvalue='...'):
+        def decorating_function(user_function):
+            return user_function
+        return decorating_function
```
philippjfr, line 2533: Couldn't be bothered? 😆

philippjfr:
Finally! Thanks for tackling this one.

philippjfr (approved):
Looks good, and nice solution.

param/parameterized.py
jbednar: To fix this issue on py2? Definitely not!

param/parameterized.py
```diff
@@ -2677,13 +2702,15 @@ def __str__(self):
         return "<%s %s>" % (self.__class__.__name__,self.name)
 
 
+    @recursive_repr()
```
jbednar, line 2705: Actually, I don't think this one's needed, because it's not recursive; it just calls self.pprint, and self.pprint can handle the recursion check.
```suggestion
```

PR #509, opened 2021-07-08
Replace nose with pytest

maximlt:
In line with what was done partially in #423, the goal of this PR is to replace nose by pytest:

* [x] Minimal changes required to switch to pytest and get the CI green
* [x] Fix coverage

When this is done, some optional tasks could be tackled such as fixing the warnings already spotted in #423, switching to fixtures when relevant...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,3 +71,7 @@ jobs:
       - name: unit all_deps
         if: (contains(matrix.os, 'ubuntu') && !startsWith(matrix.python-version, 'py'))
         run: tox -e with_all
+      - name: run coveralls
+        env:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+        run: tox -e coveralls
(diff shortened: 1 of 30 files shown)


jbednar:
Thanks! If you can see a place in the new docs to put information about running the tests, it would be good to capture and communicate that to our (power) users. Could be worth having in a HoloViz developer's guide, as I believe the instructions will now be the same across most of our projects.

maximlt:
Yes the new docs could have a Developers page. Do you use pyctdev to develop param? (I haven't used it here)

jbednar:
Param was written long before pyctdev or any of the other HoloViz projects. I recently briefly attempted to move Param to use pyctdev (https://github.com/holoviz/param/pull/502/files), but reverted those changes as there were clearly a whole set of related changes that would need to be made for it to work. Making Param more consistent with the rest of the HoloViz projects using pyctdev would be great, if someone could do that.

maximlt:
coverage reporting with coveralls seems to be [working again](https://coveralls.io/jobs/83398226).

jbednar (approved):
Thanks!

PR #517, opened 2021-08-19
Calendardate accepts date values only

### LINKED ISSUE #489: CalendarDate accepts datetime.datetime values
This snippet works without any error:

```python
import datetime

from param import Parameterized, CalendarDate

class A(Parameterized):
    cd = CalendarDate(datetime.datetime.now())
```

I expected this Parameter type to accept dates (e.g. `datetime.date.today()`) only. If this is a bug, it may be caused by the fact that `datetime.datetime` objects inherit from `datetime.date` (see this [SO answer](https://stackoverflow.com/a/23298608/10875966)).

maximlt:
Fixes #489

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1942,8 +1942,8 @@ class CalendarDate(Number):
         if self.allow_None and val is None:
             return
 
-        if not isinstance(val, dt.date) and not (allow_None and val is None):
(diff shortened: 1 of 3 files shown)


jbednar:
Thanks!

PR #522, opened 2021-09-11
Cleaned up script_repr

jbednar:
Adds the `script_repr` function to the main param module namespace for ease of use, and provides default values for all parameters. All `script_repr` options now match the `pprint()` function, but have different defaults suitable for generating a script rather than a prettyprinted terminal representation. Also adds a `show_imports` argument (defaulting to true) which adds the collected import statements to the top of the script.

The only non-backwards-compatible change should be that the pprint function no longer checks that a Parameterized object has a script_repr method if it doesn't have a pprint method. Both methods have been implemented in Param for many years now, and so it seems unlikely that a script_repr would exist with no pprint. If it does (e.g. in topographica?), it will need renaming and updating to `pprint`.

Should eventually also remove the `script_repr` method (not function) altogether in Param 2.0, which is more likely to cause an impact but even then is probably not in wide usage since it's never been documented. I verified that commenting it out doesn't hurt anything now, so it can be removed whenever we decide to. It should now be considered deprecated, and should start warning in the next release.

@jlstevens, I think the `pprint` method should be moved to the `param` subobject, though the `script_repr` method can be left as is since it should be deleted in 2.0. Can `param.pprint` be added to this PR without breaking anything else? Ideally .pprint it would only be used internally via .param, but would not yet have a deprecated warning when used directly on self, since there hasn't been a release that supports both ways.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -29,7 +29,7 @@ from .parameterized import (
     Parameterized, Parameter, String, ParameterizedFunction, ParamOverrides,
     descendents, get_logger, instance_descriptor, basestring)
 
-from .parameterized import (batch_watch, depends, output, # noqa: api import
+from .parameterized import (batch_watch, depends, output, script_repr, # noqa: api import
                             discard_events, edit_constant, instance_descriptor)
 from .parameterized import logging_level     # noqa: api import
 from .parameterized import shared_parameters # noqa: api import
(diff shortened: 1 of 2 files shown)


jbednar:
@jlstevens, I went ahead and hard-coded `self.param.pprint` to invoke `self.pprint`; let me know if there is a cleaner way to handle that. Should be ok to release now with a compatibility note saying that users should update self.pprint to self.param.pprint in their code to avoid warnings in future releases (and to avoid breakage with 2.0).

PR #523, opened 2021-09-13
Added Outputs and Serialization and Persistence user guides

jbednar:
Added two new user guides:

- Serialization and Persistence: Saving the state of a Parameterized object to a text, script, or pickle file
- Outputs: Output types and connecting output to Parameter inputs

There was also planned to be a Customization user guide, but after writing it there was very little material, so that material was added to the Parameters user guide instead.

Other misc changes:
- Added basic coverage of pydantic in Comparisons
- Fixed MultiFileSelector example in Parameter_Types. Added classlist and descendents docs.
- Documented edit_constant, shared_parameters, and pprint in Parameters


examples/user_guide/Outputs.ipynb
```diff
@@ -0,0 +1,164 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "94463ebb",
```
philippjfr, line 5: Oh no, looks like you upgraded 😄

jbednar: Happy to have the 'id' field added to what we strip out...

PR #526, opened 2021-09-17
CI: disable pandas and numpy tests on pypy 

maximlt:


jbednar:
Wow, the tests pass much more quickly now! The pandas and numpy tests on pypy, particularly on Windows, had been taking so long that they timed out after 30 minutes (e.g. https://github.com/holoviz/param/pull/525/checks?check_run_id=3634139230). The pypy tests are still much slower (6m for pypy on windows and 4m for pypy on linux vs <3 min for every other test), but this gives a massive improvement and should make it work much nicer for development. Thanks!

PR #528, opened 2021-09-20
Give Range a proper docstring

maximlt:
It was just a regular string.


maximlt:
Bad luck, coveralls was down for maintenance when the CI was triggered. The test suite passed.

PR #529, opened 2021-09-20
Fix test_initialization_without_step_instance

maximlt:
It was wrongly a class rather than an instance.


maximlt:
Bad luck, coveralls was down for maintenance when the CI was triggered. The test suite passed.

PR #530, opened 2021-09-20
Update supported Python versions and add a badge in the README

maximlt:
The Python versions supported by param are reflected by those currently tested in the CI, i.e. 2.7, 3.6, 3.7, 3.8 and 3.9. This PR updates setup.py to list these versions. Note that Python 3.10 will be out soon and it'd be nice to add it to the CI and see how that goes.

This PR also adds a badge to the README that shows the Python versions supported by param. Since param is meant to be a foundational library of its kind, it's nice to show that it is compatible with many versions of Python. This badge could also be added to the landing page of the site if required.


jbednar:
Param is a foundational library and thus should be more open in version support than just about any other library. But like Python 2.7 itself, I think that concern only applies to Param 1.x, which should continue to be usable with Python 2.7 indefinitely. Once we switch to Param 2.x I don't see any reason to provide Python 2.7 support any longer.

PR #531, opened 2021-09-20
ListSelector restricted to list type objects

### LINKED ISSUE #527: ListSelector doesn't check the type
There seems to be two issues with `ListSelector`.

It raises a `TypeError` when the type is wrong (e.g. `p.ls = 1`). Which doesn't sound so bad but `param` actually raises a `ValueError` in those cases, e.g. for a `Number` parameter: `ValueError: Parameter 'n' only takes numeric values, not type <class 'str'>.`. For consistency, should it also raise a `ValueError`?

`ListSelector` doesn't check that the type is a `list` which means that this snippet doesn't fail:

```python
import param

class A(param.Parameterized):
    ls = param.ListSelector('ab', ['a', 'b', 'c', 'd'])

a = A()
a.ls = 'cd'
```

Is this a bug?

maximlt:
Fixes #527 

I chose to raise a `ValueError` when the type is not a list since raising `ValueError` seems to be what param does in this case. The tests used to catch `TypeError`s in some occasions because param was trying to iterate over non iterables.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1862,6 +1862,9 @@ class ListSelector(Selector):
     def _validate(self, val):
         if (val is None and self.allow_None):
             return
+        if not isinstance(val, list):
+            raise ValueError("ListSelector parameter %r only takes list "
+                             "types, not %r." % (self.name, val))
         for o in val:
             super(ListSelector, self)._validate(o)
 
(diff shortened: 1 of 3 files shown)


philippjfr:
We need to be consistent about `ValueError` and `TypeError` but in this case this is more consistent because it matches what `param.List` raises when given the wrong type.

PR #532, opened 2021-09-21
Fixed version regexp

jbednar:
Previous regexp would have matched "vim" and "vis" and "van" as valid version strings, which surely they are not. Presumably also needs fixing on autover.

### DIFF
--- a/param/version.py
+++ b/param/version.py
@@ -191,7 +191,7 @@ class Version(object):
                     except: pass
             if output is None:
                 output = run_cmd([cmd, 'describe', '--long', '--match',
-                                  "v[0-9]*.[0-9]*.[0-9]*", '--dirty'],
+                                  "v[0-9]*[.][0-9]*[.][0-9]*", '--dirty'],
                                  cwd=os.path.dirname(self.fpath))
             if as_string: return output
         except Exception as e1:


maximlt:
`v1.1` seems to be allowed, so how about `v[0-9]*[.][0-9]*[.]?[0-9]*`

jlstevens:
@maximlt Here the regular expression is what is accepted by `git describe` (it isn't a Python regexp). I tried your version and it didn't match any tags for HoloViews:

```
git describe --long --match v[0-9]*[.][0-9]*[.]?[0-9]* --dirty
fatal: No names found, cannot describe anything.
```

Jim's version works though so I'll go ahead and merge.

maximlt:
Alright it's not a regular expression at all but a glob so `*` matches any string.

jbednar:
D'oh! Yes, `git-describe --match <pattern>` matches a glob, not a regex, so my proposed "fix" isn't needed. Oops!

jlstevens:
Is it worth reverting the change? 

This regexp defines the possible space of version strings and all the ones we use in holoviz are compatible with what we had before and the change in this PR. I have no opinion other than this is not a Python regexp and you need to have a clear idea of what git is doing (which I don't right now). :-)

jbednar:
I think it's worth reverting, because it diverges from autover and looks even more like a regex than it did before. git describe just refers to `glob(7)`, which isn't a man page on my own system, but online it looks like it's just the usual glob syntax for bash.

jbednar:
Reverted in edc0528a520fdea132918814f398a0498b76b75e .

PR #533, opened 2021-09-22
Allow resolving dependencies dynamically

philippjfr:
Fixes https://github.com/holoviz/param/issues/410

- [ ] Add (tons of) tests
- [ ] Ensure old watchers are cleaned up


jbednar:
@philippjfr, can you please rebase this on master? There will be a few conflicts.

param/parameterized.py
```diff
@@ -518,29 +521,82 @@ def _output(*args,**kw):
...
-    return params
+                dependencies += _params_depended_on(subdep)
+    return dependencies
+
+
+def _skip_event(*events, changed=None):
+    """
+    Checks whether a subobject event should be skipped if all the
+    values on the subobject match the values on the previous
+    subobject.
```
jbednar, line 575:
```suggestion
    Checks whether a subobject event should be skipped.
    Returns True if all the values on the new subobject 
    match the values on the previous subobject.
```

param/parameterized.py
```diff
@@ -1496,6 +1562,57 @@ def _instantiate_param(self_, param_obj, dict_=None, key=None):
...
+                for dep in _resolve_mcs_deps(obj, [], [ddep]):
+                    grouped[(id(dep.inst), id(dep.cls), dep.what, dattr)].append((ddep, dep))
+
+            for (_, _, _, dattr), group in grouped.items():
+                watcher = self_._watch_group(obj, n, queued, group)
+                obj._dynamic_watchers[(n, dattr)].append(watcher)
+
+    def _watch_group(self_, obj, name, queued, group):
+        ddeps = [dd for dd, _ in group if dd is not None]
+        _, gdep = group[0] # Need to grab representative dep from this group
```
jbednar, line 1595: I wasn't able to trace through the logic to convince myself that group will always be non-empty. If it will,please put a comment explaining where that's ensured, so that [0] is safe to do.

jbednar (approved):
Thanks!

param/parameterized.py
philippjfr: Wouldn't be much of a group if it didn't have any members now would it? 😉

philippjfr: Groups here are basically the output of a `groupby` like operation so it makes no sense for them to be empty. Anyway the method now has a docstring.

param/parameterized.py
```diff
@@ -1957,45 +2164,59 @@ def outputs(self_):
                     outputs[name] = (otype, method, idx)
         return outputs
 
-
-    def _spec_to_obj(self_,spec):
-        # TODO: when we decide on spec, this method should be
-        # rewritten
-
+    def _spec_to_obj(self_, spec, dynamic=True):
```
jlstevens, line 2167: Definitely needs a docstring to say what it is supposed to be doing for such a tricky method!

param/parameterized.py
```diff
@@ -2049,43 +2276,54 @@ def watch(self_, fn, parameter_names, what='value', onlychanged=True, queued=Fal
 
         See help(Watcher) and help(Event) for the contents of those objects.
         """
+        if precedence < 0:
+            raise ValueError("User-defined watch callbacks must declare "
+                             "a positive precedence. Negative precedences "
+                             "are reserved for internal Watchers.")
+        return self_._internal_watch(fn, parameter_names, what, onlychanged, queued, precedence)
```
jlstevens, line 2304: This reads to me as "complain when you see negative precedences (as they should be internal only) which means `_internal_watch` gets called with positive precedences'! Not sure whether this is just a naming issue but the error contradicts what you the next line seems to be suggesting...

philippjfr: Can't quite follow this objection, `_internal_watch` allows any precedence while the public API is restricted to positive.

param/parameterized.py
```diff
@@ -2049,43 +2276,54 @@ def watch(self_, fn, parameter_names, what='value', onlychanged=True, queued=Fal
...
+        assert what == 'value'
+        if isinstance(parameter_names, list):
+            parameter_names = tuple(parameter_names)
+        else:
+            parameter_names = (parameter_names,)
         watcher = Watcher(inst=self_.self, cls=self_.cls, fn=fn,
                           mode='kwargs', onlychanged=onlychanged,
                           parameter_names=parameter_names, what=what,
-                          queued=queued)
+                          queued=queued, precedence=0)
```
jlstevens, line 2342: Looking at the diff, precedence is checked (make sure it is > 0) but then isn't used and hard coded to zero?

philippjfr: Good catch!

param/parameterized.py
philippjfr: Sure, does seem like an implementation detail users don't need to know about but I don't actually think saying negative precedences are **reserved** for internal watchers implies that they are exclusively negative in any way.

param/parameterized.py
```diff
@@ -1957,45 +2164,80 @@ def outputs(self_):
...
+        every part of the path, e.g. for a dependency specification
+        like "a.b.c" we return dependencies for sub-object "a" and the
+        sub-sub-object "b" in addition to the dependency on the actual
+        parameter "c" on object "b". This is to ensure that if a
+        sub-object is swapped out we are notified and can update the
+        dynamic dependency to the new object. Even if a sub-object
+        dependency can only partially resolved, e.g. if object "a"
+        does not yet have a sub-object "b" we must watch for changes
+        to "b" on sub-object "a".
+        """
```
jlstevens, line 2188: That helps clarify things a lot, thanks! That said, I would expect a single list of dependencies to be returned given this docstring but I see the actual return is two lists, `(deps, deferred)`. What is the difference and what is meant by `deferred`?

philippjfr: Thought the first paragraph captured that, so suggestions for improving that welcome:

> Resolves dependency specifications into explicit parameter dependencies and dynamic dependencies to be resolved when the sub-object whose parameters are being depended on is defined.

In other words `deferred` are dependency specs that have not yet been resolved. Either because its class creation time and we don't want to resolve them yet or because they cannot be (fully) resolved because the subobject(s) they are referencing are not yet defined.

philippjfr: Can rename deferred to `dynamic_deps` (I have done that in most other places).

param/parameterized.py
jlstevens: Is `_watch` already used? The underscore suggest that it isn't public API and therefore internal...

philippjfr:
@jbednar Please review and merge.

PR #536, opened 2021-09-24
Added Dependencies_and_Watchers user guide

jbednar:
Merging this will make the Param User Guide complete for the first time in history!  Still needs some polish and possible reorganization, but the docs at least now cover (or are intended to cover) all of the public API and functionality of param.

This PR includes some minor changes to Param itself that should not affect backwards compatibility:

1. The context manager `param.parameterized.batch_watch` accepted two optional arguments (`enable` and `run`) that shouldn't be part of the public API as they are difficult to explain and shouldn't be needed by end users. Because the name was confusing (it doesn't watch things in batch, it batches up events created when calling watchers), created a new public API version called `param.parameterized.batch_call_watchers`, accepting only the single required parameterized argument. `batch_watch` is kept as an alias for what is now called `_batch_call_watchers` internally, but that alias should be removed in a later Param version.
2. The `_watch` method was accepting an argument `operation` that appears never to have been implemented or used internally. Since that's a private API, simply removed it from the argument list.
3. The namedtuples PInfo, MInfo, Watcher, and Event all now have docstrings (Python3 only).

The rest of the PR is to add the guide and fix up the index to declare the User Guide complete. The new guide currently requires aiohttp and panel for that notebook to run, so those have been added to the docs requirements in setup.py, but I'm not 100% certain that it's appropriate to require those for every build.

@philippjfr and @jlstevens , please verify my changes to parameterized.py and then actually check out this PR to work through the Dependencies_and_Watchers notebook cell by cell, trying to see if you can spot any errors or omissions.

### DIFF
--- a/examples/user_guide/Dependencies_and_Watchers.ipynb
+++ b/examples/user_guide/Dependencies_and_Watchers.ipynb
@@ -32,7 +32,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "58661e91",
+   "id": "e8166ebf",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -71,7 +71,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "f14d35a5",
+   "id": "d308a16d",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -112,7 +112,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "7c2101bd",
+   "id": "9cab7a1b",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -146,7 +146,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "299991bf",
+   "id": "7f7f48f2",
    "metadata": {},
    "source": [
     "Here we have created an object `d` of type `D` with a unique ID like `D00003`. `d` has various parameters, including one nested Parameterized object in its parameter `n`. In this class, the nested parameter is set to our earlier object `c`, using `instantiate=False` to ensure that the value is precisely the same earlier object, not a copy of it. You can verify that it is the same object by comparing e.g. `name='C00002'` in the repr for the subobject in `d` to the name in the repr for `c` in the previous section; both should be e.g. `C00002`.\n",
@@ -167,7 +167,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "0576bcb2",
+   "id": "010df800",
    "metadata": {},
    "source": [
     "Here we can see that method `cb1` will be invoked for any value changes in `d`'s parameters `x` or `s`, for any value changes in `c`'s parameter `country`, and a change in the `constant` slot of `s`. These dependency relationships correspond to the specification `@param.depends('x', 's', 'n.country', 's:constant', watch=True)` above.\n",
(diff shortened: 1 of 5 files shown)


examples/user_guide/Dependencies_and_Watchers.ipynb
```diff
@@ -0,0 +1,820 @@
...
+    "\n",
+    "Each of these will be described in the following sections."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "86823025",
+   "metadata": {},
+   "source": [
+    "### `batch_call_watchers`\n",
```
philippjfr, line 528: Would be nice to mention `.param.set_param` here and that it batches by default.

examples/user_guide/Dependencies_and_Watchers.ipynb
```diff
@@ -0,0 +1,820 @@
...
+   "source": [
+    "(Notice that none of the callbacks is invoked, despite all the Watchers on `p.a` and `p.b`.)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "dd987890",
+   "metadata": {},
+   "source": [
+    "### `.param.trigger`\n",
```
philippjfr, line 607: Would be nice to mention that when modifying mutable parameters no events are triggered and that `.param.trigger` is frequently useful in such cases.

examples/user_guide/Dependencies_and_Watchers.ipynb
```diff
@@ -0,0 +1,820 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "822c7365",
+   "metadata": {},
+   "source": [
+    "### Async executor\n",
+    "\n",
+    "Param's events and callbacks described above are all synchronous, happening in a clearly defined order where the processing of each function blocks all other processing until it is completed. Watchers can also be used with the Python [asyncio](https://docs.python.org/3/library/asyncio.html) library to operate asynchronously. To do this, you can define `param.parameterized.async_executor` with an asynchronous executor that allows using coroutines and other asynchronous functions as `.param.watch` callbacks. These callbacks will be scheduled on the event loop from Tornado or asyncio. \n",
```
philippjfr, line 727: You don't need to mention `asyncio`. The important thing to mention is support for `async/await` syntax, which is part of the core language. Whether you're using an `asyncio`, [`curio`](https://github.com/dabeaz/curio),  [`trio`](https://github.com/python-trio/trio) or whatever other async event loop implementations exist doesn't much matter.

examples/user_guide/Dependencies_and_Watchers.ipynb
jbednar: Isn't that covered by the following paragraph already? Does it need more explanation?

philippjfr: Yes, sorry did not read in sufficient detail.

examples/user_guide/Dependencies_and_Watchers.ipynb
```diff
@@ -0,0 +1,820 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "346330d8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "d.x=5"
```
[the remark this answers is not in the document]
jbednar: Consistently adding the spaces would be good; I tend to omit them for very short expressions but that's probably not a distinction worth making.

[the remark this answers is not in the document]
philippjfr: Would much rather see consistency because "very short expression" does not have a strict definition.

jbednar:
Merging so I can tag an alpha version and doc build; further cleanup can be done on master.

PR #537, opened 2021-10-02
Fix conda build

maximlt:
`nose` was still expected by `conda build` to run the test suite, which couldn't work at all since it was replaced recently by pytest.

Part of this PR I've also decided to move the last doctest (param.Time) to a unit test. Why? Because conda build failed by simply replacing `nosetests tests` by `pytest tests`. It executes this from a folder that only includes the subfolder `tests`, and the command that was set to tell pytest to look for doctest to execute required the subfolders param and numberben to be also there (pytest tests param numbergen --doc-modules).

There may have been a way to configure pytest to cover this case but I thought it was just simpler to remove that last doctest. This makes even more sense now since param has some nice docs :)

Note that I've not removed the docstest code from the docstring, since it is still useful docs in itself. It's just no longer executed by pytest.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -32,7 +32,7 @@ test:
     - numbergen
   commands:
     # https://github.com/ioam/param/issues/219  
-    - nosetests tests
+    - pytest tests
 
 about:
   home: {{ sdata['url'] }}
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -31,7 +31,7 @@ test:
     - param
     - numbergen
   commands:
-    # https://github.com/ioam/param/issues/219  
+    # https://github.com/holoviz/param/issues/219  
     - pytest tests
 
 about:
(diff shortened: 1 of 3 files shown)


conda.recipe/meta.yaml
```diff
@@ -32,7 +32,7 @@ test:
     - numbergen
   commands:
     # https://github.com/ioam/param/issues/219
```
jbednar, line 34:
```suggestion
    # https://github.com/holoviz/param/issues/219  
```

tests/API1/testtimedependent.py
```diff
@@ -81,6 +81,26 @@ def test_time_int_change_type(self):
         self.assertEqual(t(), 1)
         self.assertEqual(t.time_type, fractions.Fraction)
 
+    def test_time_integration(self):
+        # This used to part a doctest of param.Time, moved
```
jbednar, line 85:
```suggestion
        # This used to be a doctest of param.Time; moved
```

jbednar (approved):
Thanks!

PR #538, opened 2021-10-02
Add binder

maximlt:
Add an environment spec for binder (pinning things to try to avoid to get the deployed env updated regularly, this is usually too slow IMO) and a badge on the README.

Once v2.0.0 is released, we could use this tag to have the docs and the content shown by binder in sync.


jbednar:
Thanks!

PR #539, opened 2021-10-04
Small improvements to the docs and its build

maximlt:
This PR fixes a lot of small issues with the docs, and brings down the count of warnings emitted during the build to almost zero:
* Added pandas to the docs dependencies
* Added titles to some notebooks
* Replaced an image that was embedded in a notebook with a concrete one saved in *assets* (had the same issue recently on panel's docs, the tools don't handle that
* Added a link to the favicon
* Stopped creating auto docs with nbsite_generate_modules to avoid duplicated references (removed most of the warnings!)
* Excluded the historical release notes files
* Adapted some docstrings to avoid sphinx warnings
* Added sphinx-copy button


philippjfr:
Fabulous, thanks @maximlt. LGTM

tests/API1/testfileselector.py
```diff
@@ -0,0 +1,176 @@
+# import os
+# import shutil
+# import tempfile
+
+# import param
+
+# from . import API1TestCase
+
+
+# class TestFileSelectorParameters(API1TestCase):
```
philippjfr, line 10: Why is this commented out?

maximlt: Arf good catch this wasn't mean to be committed, I'm working on fixing an issue with FileSelector and adding some tests, I'll remove that file now.

jbednar (approved):
Looks great, thanks so much! I'm a bit confused by all the outputs and metadata changes, but when I tried loading the notebooks locally and clearing them to get rid of those, I got all the nonsense ID fields added, which is worse. If you know why those are there (maybe check with @jlstevens ?) and can get rid of those changes, that would be great; otherwise it all looks very helpful and I'm happy to see it merged!

maximlt:
> Looks great, thanks so much! I'm a bit confused by all the outputs and metadata changes, but when I tried loading the notebooks locally and clearing them to get rid of those, I got all the nonsense ID fields added, which is worse. If you know why those are there (maybe check with @jlstevens ?) and can get rid of those changes, that would be great; otherwise it all looks very helpful and I'm happy to see it merged!

I've removed the ids from the notebooks with the new version of Jean-Luc's utility, which doesn't make the diff any cleaner for now but should be better for the future.

I've opened the notebooks with VScode and it seems that it's saving the JSON in a different order:
image
I'll have a look on that the next time I edit a notebook with VScode, but if so, that's annoying. What did you use to write the notebooks? JupyterLab?

jbednar:
I still use classic Jupyter, and I believe @jlstevens does as well, while @philippjfr uses JupyterLab. It's helpful to have different team members using different tools so that at least one of us spots problems specific to a particular editor, and generally we've been able to interoperate as long as we all run Jean-Luc's script to clear out metadata.  But having the field order change if one of us is using VSCode is very, very annoying, as those changes will quickly drown out any real changes to the content.  I don't know any other solution besides loading into some Jupyter version after editing in VSCode, which is painful but would presumably avoid the reordering.

jlstevens:
I can't promise anything yet but it *might* be possible to enforce a field order with `jq` - I'll have a look. It does seem that VSCode is the odd-one-out here which is quite annoying.

maximlt:
Before that @jlstevens let me make sure that VScode is indeed reordering the field order.

maximlt:
I created a simple notebook with Jupyter Lab, saved it, copied it and opened the copy with VSCode:
```diff
diff --git a/TestOrderJlab.ipynb b/TestOrderVSCode.ipynb
index 40ef85d..f2d29ad 100644
--- a/TestOrderJlab.ipynb
+++ b/TestOrderVSCode.ipynb
@@ -2,66 +2,61 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "d7a867e9-e87f-4c4d-806a-5617288ad6e0",
-   "metadata": {},
    "source": [
     "# Test"
-   ]
+   ],
+   "metadata": {}
   },
   {
    "cell_type": "code",
    "execution_count": 1,
-   "id": "070353d7-4e47-4f65-859e-4a44519c04a5",
-   "metadata": {},
+   "source": [
+    "print('hey')"
+   ],
    "outputs": [
     {
-     "name": "stdout",
      "output_type": "stream",
+     "name": "stdout",
      "text": [
       "hey\n"
      ]
     }
    ],
-   "source": [
-    "print('hey')"
-   ]
+   "metadata": {}
   },
+   "metadata": {}
   },
   {
    "cell_type": "markdown",
-   "id": "1d2bb87e-b2a9-4970-baeb-46b015db21b0",
-   "metadata": {},
    "source": [
     "Woups"
-   ]
+   ],
+   "metadata": {}
   },
   {
    "cell_type": "code",
    "execution_count": 2,
-   "id": "f61a2196-e244-4555-b7cf-1cb628bee651",
-   "metadata": {},
+   "source": [
+    "2 + 2"
+   ],
    "outputs": [
     {
+     "output_type": "execute_result",
      "data": {
       "text/plain": [
        "4"
       ]
      },
-     "execution_count": 2,
      "metadata": {},
-     "output_type": "execute_result"
+     "execution_count": 2
     }
    ],
-   "source": [
-    "2 + 2"
-   ]
+   "metadata": {}
   },
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "e7054905-b1ea-48be-a90d-6670642a064a",
-   "metadata": {},
+   "source": [],
    "outputs": [],
-   "source": []
+   "metadata": {}
   }
  ],
  "metadata": {
@@ -85,4 +80,4 @@
  },
  "nbformat": 4,
  "nbformat_minor": 5
-}
```
This confirms that they don't have the same ordering :(

jbednar:
Ugh! As the Billy-come-lately, I blame VSCode for not respecting the precedent set by Jupyter and JupyterLab.  I think _we_ should respect precedent, though, which I guess requires re-sorting the VSCode output. Ugh!

Issue #408, opened 2020-05-12
Feature: Support for binary strings


mdk73:
Currently param.String works with strings (`"string"`), but throws an error when a binary string `b"binary string"` is used.

When accessing hardware devices (e.g. a camera via usb), the returned strings are very often binary strings. 

So I would like to suggest a new param type:
param.BinaryString

(I had a look into the source code myself and tried to understand how it works, but I failed...)

jbednar:
Sure.  BinaryString or maybe BinaryBuffer or ByteArray.  I assume you'd copy the String implementation but then change the type to a bytearray instead.

mdk73:
Somehow I overlooked the String class while scrolling through the github file in the browser.
I found the definition of the class now, so I guess I will be able to adapt the code.

jbednar:
Oh, I should have warned you that unlike nearly every other Parameter, String is defined in `parameterized.py` and not `__init__.py`, because of a bootstrapping issue that means that we need it declared to be able to have Parameterized objects have a name parameter.  I think that's the only parameter that is special like that; ByteArray (or whatever) would go into `__init__.py` like all the rest.

mdk73:
My suggestion of a Bytes class (based on the String class):

```python
class Bytes(Parameter):
    """
    A Bytes Parameter, with a default value and optional regular expression (regex) matching.

    A class similar to the String class, but instead of type basestring (e.g. 'string')
    the type bytes (e.g. b'bytes') is used.

    """
    
    def __init__(self, default=b"", regex=None, allow_None=False, **kwargs):
        super(Bytes, self).__init__(default=default, allow_None=allow_None, **kwargs)
        
    def _validate(self, val):
        if self.allow_None and val is None:
            return

        if not isinstance(val, bytes):
            raise ValueError("Bytes '%s' only takes a byte string value."%self.name)

        if self.regex is not None and re.match(self.regex, val) is None:
            raise ValueError("Bytes '%s': '%s' does not match regex '%s'."%(self.name,val,self.regex))
```

Simple test:
```python
class Test(param.Parameterized):
    b = Bytes(b'bytes')

test = Test()
test.b
```

Test similar to the given test in the String documentation:
```python
class IPAddress(Bytes):
    '''IPv4 address as bytes (dotted decimal notation)'''
       
    def __init__(self, default=b"0.0.0.0", allow_None=False, **kwargs):
[...]

PR #542, opened 2021-10-07
Add Bytes parameter

philippjfr:
Implements https://github.com/holoviz/param/issues/408

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -54,6 +54,7 @@ except ImportError:
 
 if sys.version_info[0] >= 3:
     unicode = str
+else:
     bytes = str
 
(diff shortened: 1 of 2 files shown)


philippjfr:
>Would I be able to use the Bytes parameter in Panels ReactiveHTML to efficiently transfer files or Arrow dataframes between server and client?

Could look at that for sure but note that `param.DataFrame` parameters on ReactiveHTML are already serialized efficiently as a ColumnDataSource.

jbednar (approved):
Looks good! It's a little odd for the unit test to use an IP address instead of something that would be more naturally stored as bytes, but I guess that's fine, so feel free to merge as-is.

PR #546, opened 2021-10-08
Update objects when path changes for MultiFileSelector and FileSelector

### LINKED ISSUE #545: FileSelector/MultiFileSelector don't update objects when path is updated and FileSelector doesn't set its value
`FileSelector` behaves like `Selector` and should as such set its value to the first item of *objects*, *objects* being the result of a glob on *path*. When *path* is updated it should also update *objects*.

`MultiFileSelector` should also update *objects* when *path* is updated. Also the docs currently state that the default value is all of the matched files, but in practice it's None.

> A param.MultiFileSelector is the analog of ListSelector but for files, i.e., again supporting a path glob but allowing the user to select a list of filenames rather than a single filename. The default value in this case is all of the matched files, not just the first one.

```python
import param

class A(param.Parameterized):
    fs = param.FileSelector(path='/usr/share/*')
    mfs = param.MultiFileSelector(path='/usr/share/*')

a = A()

print(a.fs is None)  # True!
paths = a.param.fs.objects.copy()
a.param.fs.path = '/usr/bin/*'
print(a.param.fs.objects == paths)  # True!

print(a.mfs == a.param.mfs.default == None)  # True!
paths = a.param.mfs.objects.copy()
a.param.mfs.path = '/usr/bin/*'
print(a.param.mfs.objects == paths)  # True!
```

maximlt:
Fixes #545 (only addresses the update path issue, not the default value)

```python
import param

class A(param.Parameterized):
    fs = param.FileSelector(path='/usr/share/*')
    mfs = param.MultiFileSelector(path='/usr/share/*')

a = A()

print(a.fs is None)  # True
paths = a.param.fs.objects.copy()
a.param.fs.path = '/usr/bin/*'
print(a.param.fs.objects == paths)  # False

print(a.mfs is None)  # True
paths = a.param.mfs.objects.copy()
a.param.mfs.path = '/usr/bin/*'
print(a.param.mfs.objects == paths)  # False
```

### DIFF
--- a/examples/user_guide/Parameter_Types.ipynb
+++ b/examples/user_guide/Parameter_Types.ipynb
@@ -745,9 +745,11 @@
     "\n",
     "A `param.ListSelector` works just the same as a regular Selector, but the value is a _list_ of valid objects from the available objects, rather than just one. Each item in the list is checked against the `objects`, and thus the current value is thus a _subset_ of the `objects`, rather than just one of the objects.\n",
     "\n",
-    "A `param.FileSelector` works like a regular Selector with the value being a filename and the `objects` being computed from files on a file system. The files are specified as a `path` [glob](https://docs.python.org/3/library/glob.html), and all filenames matching the glob are valid `objects` for the parameter.\n",
+    "A `param.FileSelector` works like a Selector with the value being a filename and the `objects` being computed from files on a file system. The files are specified as a `path` [glob](https://docs.python.org/3/library/glob.html), and all filenames matching the glob are valid `objects` for the parameter. The default value is *None* unless specified.\n",
     "\n",
-    "A `param.MultiFileSelector` is the analog of ListSelector but for files, i.e., again supporting a path glob but allowing the user to select a list of filenames rather than a single filename. The default value in this case is _all_ of the matched files, not just the first one."
+    "A `param.MultiFileSelector` is the analog of ListSelector but for files, i.e., again supporting a path glob but allowing the user to select a list of filenames rather than a single filename. The default value in this case is _all_ of the matched files, not just the first one. The default value is `None` unless specified.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


philippjfr:
This looks good. @maximlt can you advise if this PR still **should** be merged and @jbednar can you please either approve and merge or suggest what other fixes are needed?

examples/user_guide/Parameter_Types.ipynb
```diff
@@ -745,9 +745,11 @@
...
+    "A `param.FileSelector` works like a Selector with the value being a filename and the `objects` being computed from files on a file system. The files are specified as a `path` [glob](https://docs.python.org/3/library/glob.html), and all filenames matching the glob are valid `objects` for the parameter. The default value is *None* unless specified.\n",
     "\n",
-    "A `param.MultiFileSelector` is the analog of ListSelector but for files, i.e., again supporting a path glob but allowing the user to select a list of filenames rather than a single filename. The default value in this case is _all_ of the matched files, not just the first one."
+    "A `param.MultiFileSelector` is the analog of ListSelector but for files, i.e., again supporting a path glob but allowing the user to select a list of filenames rather than a single filename. The default value in this case is _all_ of the matched files, not just the first one. The default value is `None` unless specified.\n",
```
jbednar, line 750: This appears to say two different values for the default, which presumably can't be true?

jbednar (approved):
Seems like an appropriate change to me. We might want to revisit after https://github.com/holoviz/param/pull/605 to make the default always be the first value or all values (for multi) unless the user has explicitly provided a default of None (currently not possible to detect).

jbednar:
I think is is ready for revison by @maximlt . The diffs are confusing right now, perhaps because of conflicting files, and I can't see where the changes that implement updating objects are done.

PR #548, opened 2021-10-08
Run the test suite with Python 3.10

maximlt:
Let's find out if it works :)


jbednar:
Skipping the pandas tests with 3.10 on Windows sounds very reasonable in this case!

PR #557, opened 2021-10-19
Add compatibility layer for Param watcher precedence

philippjfr:
Old versions of Panel created custom so called `LinkWatcher` which track more information about the specific link they were setting up. They did so by unpacking existing watchers, which now errors because the tuple unpacking now has one extra item - the precedence. For now we are dropping the precedence from the tuple unpacking. This is **HACKY** but the only way we can ensure backward compatibility and should not affect anyone else because a) accessing watchers is rare for users and b) iterating over them is even rarer.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -756,6 +756,11 @@ class Watcher(_Watcher):
         """
         return iter(self[:-1])
 
+    def __str__(self):
+        cls = type(self)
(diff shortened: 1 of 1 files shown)


jlstevens:
This is probably the best option out of a bag of horrible ones. I'm fine with this hack as long as we put in a plan for when it can be eventually removed.

PR #563, opened 2021-10-22
Try with installing graphviz instead of libgraphviz-dev

maximlt:


maximlt:
@jbednar dev site deployed with the graphviz plots: https://pyviz-dev.github.io/param/reference.html#

Issue #578, opened 2021-12-07
Date does not work with allow_None


MarcSkovMadsen:
Param: 1.12.0

I'm trying to help a user that wants to implement the Pydantic *getting started example* using Param. https://discourse.holoviz.org/t/creating-an-optional-datetime-parameter-using-param/3147.

He asks how to implement `signup_ts: Optional[datetime] = None`. I think its not possible with the existing parameters as I get the below error.

```bash
$ python 'script.py'
Traceback (most recent call last):
  File "script.py", line 12, in <module>
    deserialized_data = User.param.deserialize_parameters(json_data)
  File "C:\repos\private\panel-ai\.venv\lib\site-packages\param\parameterized.py", line 2100, in deserialize_parameters
    return serializer.deserialize_parameters(self_or_cls, serialization, subset=subset)
  File "C:\repos\private\panel-ai\.venv\lib\site-packages\param\serializer.py", line 113, in deserialize_parameters
    deserialized = pobj.param[name].deserialize(value)
  File "C:\repos\private\panel-ai\.venv\lib\site-packages\param\__init__.py", line 1928, in deserialize
    return dt.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f")
TypeError: strptime() argument 1 must be str, not None
```

```python
import param
import json

class User(param.Parameterized):
    signup_ts = param.Date(allow_None=True, doc="""The sign up time of the user""")


external_data = {
    'signup_ts': None,
}
json_data = json.dumps(external_data)
deserialized_data = User.param.deserialize_parameters(json_data)
User(**deserialized_data)
```

jbednar:
I've changed the title, because `Date` parameters _do_ support `allow_None`:

```python
>>> import param, json
>>> class User(param.Parameterized):
        signup_ts = param.Date(allow_None=True, doc="""The sign up time of the user""")
>>> u = User()
>>> print(u.signup_ts)
None
```

The example above is specifically about serializing Date Parameters whose value is None, and it does look like there is an issue there. @jlstevens?

> If you look at the Pydantic getting started example you will realize that its deserializing is much more flexible/ forgiving as it will be able to converts string to ints for integer parameters, not care so much about the date time format for datetime parameters etc.

[...]

PR #582, opened 2021-12-15
Updated JSON serialization of Date and CalendarDate to support None

jlstevens:
Simple fix that addresses https://github.com/holoviz/param/issues/578 issue for `None` valued `Date` parameters and does the same for `CalendarDate` parameters.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1017,10 +1017,15 @@ class Tuple(Parameter):
 
     @classmethod
     def serialize(cls, value):
-        return list(value) # As JSON has no tuple representation
+        if value is None:
+            return 'null'
+        else:
+            return list(value) # As JSON has no tuple representation
 
     @classmethod
     def deserialize(cls, value):
+        if value == 'null':
+            return None
         return tuple(value) # As JSON has no tuple representation
 
 
@@ -1474,10 +1479,15 @@ class Array(ClassSelector):
 
     @classmethod
     def serialize(cls, value):
-        return value.tolist()
+        if value is None:
+            return 'null'
+        else:
+            return value.tolist()
 
     @classmethod
     def deserialize(cls, value):
+        if value == 'null':
+            return None
         from numpy import asarray
         return asarray(value)
 
@@ -1557,10 +1567,15 @@ class DataFrame(ClassSelector):
 
     @classmethod
     def serialize(cls, value):
-        return value.to_dict('records')
+        if value is None:
+            return 'null'
+        else:
+            return value.to_dict('records')
 
     @classmethod
     def deserialize(cls, value):
+        if value == 'null':
+            return None
         from pandas import DataFrame as pdDFrame
         return pdDFrame(value)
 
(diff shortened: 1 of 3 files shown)


jlstevens:
@jbednar @philippjfr Ready for review assuming the tests go green on CI (they work fine locally).

jlstevens:
Tests have passed except for the pypy suite due to an unrelated issue.

jlstevens:
Thanks for prompting me to have another look!

I realize that my initial impression *was* correct and `JSONNullable` is applied even for the special schema methods. I'll push a commit now to revert that unnecessary change.

Otherwise it is about checking how the `serialize` and `deserialize` handle `'null`'. The full list of parameters using these methods are: `Tuple`, `Array`, `DataFrame`, `Date` (done) and `CalendarDate` (done).

I'll make sure `Tuple`, `Array`, `DataFrame` work properly now.

jlstevens:
>  I think users will expect those classes to behave the same in all cases apart from the type, so for futureproofing we could consider refactoring them into a BaseDate class with a strformat parameter ("%Y-%m-%d" vs. "%Y-%m-%dT%H:%M:%S.%f"), or just making Date be that baseclass and having very little code in CalendarDate so that those two classes always stay in sync.

At a glance that does look to be true but looking closer there isn't *that* much shared. The `isinstance` checks are different and other than the `strformat` format,   `CalendarDate` uses the [`.date()` method](https://github.com/holoviz/param/blob/master/param/__init__.py#L1966). So on balance I don't think there is *that* strong an incentive to make a base class.

jbednar:
Ah, didn't spot ".date()".  Ok, keeping them separate is fine. Having a general declarative implementation of the instance checks across many Parameter types would be a separate task, but would seem to simplify things.

jlstevens:
@philippjfr @jbednar Any idea what the rationale might have been to hardcode `allow_None=True` for `Array`? Is there a reason arrays always have to accept `None`?

```python
class Array(ClassSelector):
    """
    Parameter whose value is a numpy array.
    """

    def __init__(self, default=None, **params):
        from numpy import ndarray
        super(Array, self).__init__(ndarray, allow_None=True, default=default, **params)
```

jbednar:
I would assume that's because Pandas is not a required dependency and thus we can't have an array as the default value. That said, I think the right thing to do is to have the default be None but _not_ specify `allow_None`, so that when someone instantiates their own Array parameter they can put their own default in, typically non-None, and then separately decide whether to allow None. That's a breaking change, but I think it's the right thing to do, as I don't think `allow_None` should have been hardcoded except when using the default None value.

jlstevens:
Your suggestion about `allow_None` makes sense to me and should be put into another issue.

Otherwise, I have added None/'null' support for `Tuple`, `Array` and `DataFrame`. Ready for review/merge once more.

jbednar (approved):
Looks good, but please consistently choose _either_ an early return style:

```python
if value is None:
    return 'null'

return val...
```
_or_ an if/else:

```python
if value is None:
    return 'null'
else:
    return val...
```

Right now it's a mix of the two, and thus less easy to scan.

jbednar (approved):

PR #596, opened 2022-01-25
Create a new `objects` instead of appending a value when `check_on_set` is False

maximlt:
This is useful for Panel when an `AutocompleteInput` widget is created from a `Selector` Parameter, with `restrict` (widget) and `check_on_set` (Parameter) set to False. Before this change, a new value manually added to the list of completions/options by a user would just be appended to `objects`, and as such not reflected in the widget `options`. With this change `objects` and `options` are in sync.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1289,7 +1289,7 @@ class Selector(SelectorBase):
         to check each item instead.
         """
         if not (val in self.objects):
-            self.objects.append(val)
+            self.objects = self.objects + [val]
 
     def get_range(self):
         """


jbednar:
Hmm. Is there a different way that we can get Panel to realize that the object list has changed? Here it seems like the user may have their own reference to the list of objects and expect to be able to modify that list as needed to change the available options, especially outside the context of Panel.

maximlt:
Do you mean something like that?
image

philippjfr:
This should no longer be needed once https://github.com/holoviz/param/pull/598 is merged because that triggers events even when modifying `.objects` inplace.

philippjfr:
I'll close because even if we decide to remove the trigger from #598 we should use `.param.trigger` instead of replacing the list.

Issue #309, opened 2018-12-14
Keep `names` attribute and `objects` attribute of an `ObjectSelector` in sync


sdc50:
Now that the `ObjectSelector` accepts a mapping as the `objects` argument of the constructor it would be nice to also allow the `objects` attribute of the class to be set with a mapping. Otherwise, if the `objects` attribute is changed after instantiation then the `names` attribute is out of sync. For example:

![image](https://user-images.githubusercontent.com/6454472/50022610-59223900-ffa2-11e8-937e-c777d3d28e8b.png)

One suggestion would be to make `objects` a property that has a setter:
```python
class ObjectSelector(Selector):
...
     __slots__ = ['_objects','compute_default_fn','check_on_set','names']

    @property
    def objects(self):
        return self._objects

    @objects.setter
    def objects(self, objects):
        if objects is None:
            objects = []
        if isinstance(objects, collections.Mapping):
            self.names = objects
            self._objects = list(objects.values())
        else:
            self.names = None
            self._objects = objects

    # ObjectSelector is usually used to allow selection from a list of
    # existing objects, therefore instantiate is False by default.
    def __init__(self,default=None,objects=None,instantiate=False,
                 compute_default_fn=None,check_on_set=None,allow_None=None,**params):
        self.objects = objects
        ...
```

Then the behavior would be like I would expect:
![image](https://user-images.githubusercontent.com/6454472/50022669-92f33f80-ffa2-11e8-8346-0d0c06cb1647.png)

philippjfr:
This should be fixed before we release 1.9.

PR #598, opened 2022-01-31
Add SelectorObjects wrapper for forward and backward compatibility

philippjfr:
We have long struggled with an approach to improving `Selector`, `ListSelector` and `ObjectSelector` that would allow users to easily update the `objects`. Due to unfortunate past decisions this is quite difficult because users can supply both a list or a dict in the constructor and we decided to store `objects` as a simple list while storing the **optional** names as a dictionary on the `names` attribute. This meant that if you instantiated `objects` as a dictionary you had to manage updates to both of these attributes at the same time.

To handle this issue we create a wrapper object around the `objects` attribute which has hybrid behavior that allows both list-like and dict-like updates and warns appropriately if for instance you are trying to update `objects` with the list-like API but have supplied names for other objects previously. An additional benefit of this wrapper object is that if you use any of the `list` or `dict` API to modify the `objects` we can trigger an event, while in the past a user would manually have to trigger such an event if they wanted Panel or some other downstream library to update in response to the change in `objects`.

Fixes https://github.com/holoviz/param/issues/309
Fixes https://github.com/holoviz/param/issues/331
Fixes https://github.com/holoviz/param/pull/598
Fixes https://github.com/holoviz/param/pull/398

- [x] Add tests
- [x] Update docs


philippjfr:
Will need some internal remapping to allow users to subscribe to `parameter:objects` despite the fact the slot is now called `_objects`.

philippjfr:
@jbednar This is ready for review. I would suggest instead of reviewing the code, review the test cases and check whether they match your understanding and intuitions.

One thing to note is that I've decided to allow dictionary style updates to list-like objects by upgrading the list to a dictionary, i.e. you can do something like this:

```python
>>> p.selector.objects = ['A', 'B', 'C']

>>> p.selector.objects['D'] = 4

>>> dict(p.selector.objects)

{'A': 'A', 'B': 'B', 'C': 'C', 'D': 4}
```

examples/user_guide/Parameter_Types.ipynb
```diff
@@ -737,6 +737,8 @@
     "\n",
     "Otherwise, the objects should be provided as a _name_:_value_ dictionary, where the string name will be stored for use in such a UI, but is not otherwise accessed by Param. The values from setting and getting the parameter are always the actual underlying object, not the string names. Because the string name will need to be looked up from the value if this parameter is used in a UI, all objects need to be hashable via the `param.hashable()` function, which accepts Python literals plus list and dictionary types (treating them like tuples).\n",
     "\n",
+    "To make it easier to modify `objects` they are wrapped in a object that supports both list and dictionary style methods. This ensures that there is a consistent API for updating the `objects` and that modifying the objects in-place still triggers an event.\n",
```
jbednar, line 740:
```suggestion
    "To make it easier to modify the collection of `objects`, they are wrapped in a container that supports both list-style and dictionary-style methods. This approach ensures that there is a consistent API for updating the `objects` and that modifying the objects in place still triggers an event.\n",
```

param/__init__.py
```diff
@@ -1201,6 +1203,178 @@ def get_range(self):
         raise NotImplementedError("get_range() must be implemented in subclasses.")
 
 
+
+class SelectObjects(list):
+    """
+    Wrapper around Selector.objects which allows both list-like and
+    dictionary like updates to the set of objects, ensuring backward
+    compatibility for setting objects as lists and dictionaries.
```
jbednar, line 1211: I think a class should be documented by what it does and what it assumes, rather than by how it is used elsewhere in the code. That way this class can be defended and understood on its own, and potentially reused in other cases, rather than solely being tied to the specific way it's used in Selector.

```suggestion
    Container that supports both list-style and dictionary-style updates. Useful for replacing code that originally accepted lists but needs to support dictionary access (typically for naming items).
```

param/__init__.py
```diff
@@ -1201,6 +1203,178 @@ def get_range(self):
...
+                    )
+                k, v = o
+                self.__setitem__(k, v, trigger=False)
+            for k, v in items.items():
+                self.__setitem__(k, v, trigger=False)
+
+    def values(self):
+        if self._parameter.names:
+            return self._parameter.names.values()
+        return named_objs(self).values()
```
jbednar, line 1375: I have no way to tell if the above list of methods is exhaustive. It's certainly exhausting! :-) Hope it covers everything!

jbednar (approved):
Looks good to me. I have to admit that my eyes glazed over both on the list/dict methods and on the tests; they all look reasonable but I have low confidence I'd detect any problems. I think we'll have to test this with lots of existing code to see if we find any problems. Have you tried it against the Panel test suite?

philippjfr:
This is ready to merge once we drop py2.7 on main.

param/__init__.py
```diff
@@ -1201,6 +1203,176 @@ def get_range(self):
         raise NotImplementedError("get_range() must be implemented in subclasses.")
 
 
+
+class SelectObjects(list):
```
jlstevens, line 1207: This approach makes sense to me given the constraints, but I think the name doesn't reflect what is really happening here. The name that would have made the approach clearer to me at a glance might be something like `ListProxy`.

philippjfr: Agree.

maximlt:
Does it fix https://github.com/holoviz/param/issues/398 too?

philippjfr:
Yes, it absolutely should.

philippjfr:
Just confirmed that it does.

param/__init__.py
```diff
@@ -1275,6 +1445,19 @@ def __init__(self, objects=None, default=None, instantiate=False,
...
+    def objects(self):
+        return ListProxy(self._objects, self)
+
+    @objects.setter
+    def objects(self, objects):
+        if isinstance(objects, collections_abc.Mapping):
+            self.names = objects
+            self._objects = list(objects.values())
+        else:
+            self.names = {}
```
maximlt, line 1458: The default value of `names` is now an empty dict when `objects` is a list, instead of `None`. The test suite fails now, since I added a bunch of tests that check the default values.

Seems like this change was intentional, @philippjfr can you confirm? If so, I can update the tests accordingly.

maximlt: I assumed this was an intentional change and updated the tests accordingly. Can always amend before 2.0 is release if need be.

PR #600, opened 2022-02-03
Fix dataframe serde

### LINKED ISSUE #599: Error serializing dataframes with timestamps
Discovered "error serializing datetime" (or timestamps or others), described below, when serializing a complex class with several datetime and other parameters, and several dataframes.  Identified that the issue seems to be in how pandas df's are (de)serialized - the "to_dict" function is called (see [here](https://github.com/holoviz/param/blob/7d702adfb4ace592a9550a25e8ca49522813fef4/param/__init__.py#L1566-L1577)) instead of panda's own json functions... in a df with timestamps those are simply dumped out in the resulting dict and that is attempted to be serialized, which fails as the timestamps aren't serializable without handling.

I found that by monkey patching the functions to use pandas own to_json/read_json functions I'm able to get the desired functionality.  Also, by using the "table" [orient](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_json.html#pandas-dataframe-to-json) option we get the schema in the output, allowing better serde roundtrip.

If the code below looks OK I'll be happy to submit a PR to add

Versions:
pandas: 1.4.0
json: 2.0.9
param: 1.12.0
python 3.10
osx

Description of expected behavior and the observed behavior: 

Serialize and deserialize pandas dataframes with timestamps

Complete, minimal, self-contained example code that reproduces the issue

```
import datetime as dt
import json
import pandas as pd
import param

class TestClass(param.Parameterized):
    df_param = param.DataFrame()

df = pd.DataFrame({'a': [dt.datetime(2000, 1, 1), dt.datetime(2000, 1, 1), dt.datetime(2000, 1, 1)],
                   'b': [1, 2, 3],
                   'c': [1.0, 2.0, 3.0]})

test = TestClass(df_param=df)

test_serde_dict = TestClass.param.deserialize_parameters(test.param.serialize_parameters())
```

Stack traceback and/or browser JavaScript console output

```
  File "/Users/carlsoer/opt/anaconda3/envs/covid-model/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3251, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-212f05d3f170>", line 15, in <module>
    test_serde_dict = TestClass.param.deserialize_parameters(test.param.serialize_parameters())
  File "/Users/carlsoer/opt/anaconda3/envs/covid-model/lib/python3.10/site-packages/param/parameterized.py", line 2087, in serialize_parameters
    return serializer.serialize_parameters(self_or_cls, subset=subset)
  File "/Users/carlsoer/opt/anaconda3/envs/covid-model/lib/python3.10/site-packages/param/serializer.py", line 104, in serialize_parameters
    return cls.dumps(components)
  File "/Users/carlsoer/opt/anaconda3/envs/covid-model/lib/python3.10/site-packages/param/serializer.py", line 81, in dumps
    return json.dumps(obj)
  File "/Users/carlsoer/opt/anaconda3/envs/covid-model/lib/python3.10/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/Users/carlsoer/opt/anaconda3/envs/covid-model/lib/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Users/carlsoer/opt/anaconda3/envs/covid-model/lib/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/Users/carlsoer/opt/anaconda3/envs/covid-model/lib/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Timestamp is not JSON serializable
```

Example fix:

```
def df_serialize(cls, value):
    return json.loads(value.to_json(orient='table'))

def df_deserialize(cls, value):
    import pandas as pd
    return pd.read_json(json.dumps(value), orient='table')

param.DataFrame.serialize = df_serialize
param.DataFrame.deserialize = df_deserialize

test_serde_dict = TestClass.param.deserialize_parameters(test.param.serialize_parameters())
test_serde = TestClass(**test_serde_dict)

print(test_serde.__repr__())
print(test.__repr__())

```

output:
```
TestClass(df_param=           a  b    c
0 2000-01-01  1  1.0
1 2000-01-01  2  2.0
2 2000-01-01  3  3.0, name='TestClass00002')
TestClass(df_param=           a  b    c
0 2000-01-01  1  1.0
1 2000-01-01  2  2.0
2 2000-01-01  3  3.0, name='TestClass00002')
```

ektar:
Resolves #599

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1567,14 +1567,16 @@ class DataFrame(ClassSelector):
     def serialize(cls, value):
         if value is None:
             return 'null'
-        return value.to_dict('records')
+        import json
+        return json.loads(value.to_json(orient='table'))
 
     @classmethod
     def deserialize(cls, value):
         if value == 'null':
             return None
-        from pandas import DataFrame as pdDFrame
-        return pdDFrame(value)
+        import json
+        import pandas as pd
+        return pd.read_json(json.dumps(value), orient='table')
 
 
 class Series(ClassSelector):
--- a/tests/API1/testjsonserialization.py
+++ b/tests/API1/testjsonserialization.py
@@ -42,7 +42,7 @@ class TestSet(param.Parameterized):
     __test__ = False
 
     numpy_params = ['r','y']
-    pandas_params = ['s','t','u','z']
+    pandas_params = ['s','t','u','z','ab']
     conditionally_unsafe = ['f', 'o']
 
     a = param.Integer(default=5, doc='Example doc', bounds=(2,30), inclusive_bounds=(True, False))
@@ -73,6 +73,9 @@ class TestSet(param.Parameterized):
     y = None if np is None else param.Array(default=None)
     z = None if pd is None else param.DataFrame(default=None, allow_None=True)
     aa = param.Tuple(default=None, allow_None=True, length=1)
+    ab = None if pd is None else param.DataFrame(default=pd.DataFrame(
+         {'A':[datetime.datetime(year,1,1) for year in range(2020,2023)], 'B':[1.1,2.2,3.3]}),
+         columns=(1,4), rows=(2,5))
 
 
 test = TestSet(a=29)


philippjfr:
Thanks for pushing this fix! I have a few concerns unfortunately. First, it seems quite inefficient to serialize to a JSON string and then load it again and do the reverse on deserialization. Also a bit confused as even a very simple df doesn't roundtrip for me using this approach:

Screen Shot 2022-02-03 at 17 38 24

The other thing which is not necessarily a problem is that this includes a schema in the data declaration, e.g. for a simple table this might look like this:

```json
{"schema":{"fields":[{"name":"index","type":"integer"},{"name":0,"type":"number"},{"name":1,"type":"number"},{"name":2,"type":"number"}],"primaryKey":["index"],"pandas_version":"0.20.0"},"data":[{"index":0,"0":0.5529858391,"1":0.3236336179,"2":0.8076434981},{"index":1,"0":0.6460423326,"1":0.3008845818,"2":0.4971573288},{"index":2,"0":0.0956722025,"1":0.4527357795,"2":0.7199068404},{"index":3,"0":0.4972505762,"1":0.5035357012,"2":0.4823526223},{"index":4,"0":0.5667421155,"1":0.4730645023,"2":0.5405153226},{"index":5,"0":0.212481649,"1":0.3453984203,"2":0.9464833136},{"index":6,"0":0.287443691,"1":0.1836393869,"2":0.6931539396},{"index":7,"0":0.4739893758,"1":0.3493285907,"2":0.799432673},{"index":8,"0":0.7580364284,"1":0.354098563,"2":0.7434683717},{"index":9,"0":0.3159887857,"1":0.8479990624,"2":0.5183711565}]}
```

This is helpful so it can accurately decode the datetime type or other custom types but it's also a bit odd since we try to keep the schema separate from the data.

jlstevens:
I will have a closer look at this shortly but looking at the schema Philipp just posted, I have to agree...it is definitely weird to have a schema that includes so many data values like that.

philippjfr:
>I have to agree...it is definitely weird to have a schema that includes so many data values like that.

The opposite is the case, the data serialization includes the schema, not the other way around.

jlstevens:
Thanks for the correction, though either way, these two things shouldn't be mixed together like that!

ektar:
Good catch!  

Investigated, seems issue just with the pandas "table" orient - roundtrips work if df created using records({0: [], 1: []...}), fails if df was created just with a numeric array like you show. 

I added test with dataframe created with an array instead of records - test fails as expected, then removed the 'orient' option to just have default... the problem now is that since schema isn't present anymore, pandas has no way to read the date strings and infer they are supposed to be datetype vs string.

Summary thus far - without more significant modifications to give pandas type hints (or make panda's orient=table work for matrices), output dataframe won't match input.  

A few options:
  1. Use orient='table' work with pandas devs to fix orient='table' for matrix input.  Rationale - functionality currently broken for tables with well defined rows/columns that have dates, which seems a primary use of pandas dataframes.  Purely numeric tables seem mostly used for tests, as pandas functions for these offer little advantages over numpy.  Schema is generally helpful to provide to pandas during deser to ensure correct dtypes coming back, not rely on inference logic.
  2. Use default orient, specify date format (to preserve time zone), leave it to user or future capability to cast to appropriate types.

orient='table': dates and df's constructed with record format works, df's constructed with matrices fail
```python
def serde_func(df):
    orient='table'
    return pd.read_json(df.to_json(orient=orient), orient=orient)

def non_eq_disp(*dfs):
    for df in dfs:
        display(df)
    for df in dfs:
        display(df.dtypes)

def run_test(df):
    new_df = serde_func(df)
    eq = df.equals(new_df)
    print(f'Equal: {df.equals(new_df)}')
    if not eq:
        non_eq_disp(df, new_df)
        
print('Date test')
df = pd.DataFrame(
    {
        "A": [datetime.datetime(year, 1, 1) for year in range(2020, 2023)],
        "B": [1, 2, 3],
    }
)
run_test(df)

print('Numeric test')
df = pd.DataFrame([[3, 3], [4, 2]])
run_test(df)
```

orient='table' output:
```
Date test
Equal: True
Numeric test
IntCastingNaNError                        Traceback (most recent call last)
Input In [74], in <module>
     15 print('Numeric test')
     16 df = pd.DataFrame([[3, 3], [4, 2]])
...
File ~/opt/anaconda3/envs/dev-param/lib/python3.10/site-packages/pandas/core/dtypes/cast.py:1193, in astype_float_to_int_nansafe(values, dtype, copy)
   1189 """
   1190 astype with a check preventing converting NaN to an meaningless integer value.
   1191 """
   1192 if not np.isfinite(values).all():
-> 1193     raise IntCastingNaNError(
   1194         "Cannot convert non-finite values (NA or inf) to integer"
   1195     )
   1196 return values.astype(dtype, copy=copy)

IntCastingNaNError: Cannot convert non-finite values (NA or inf) to integer
```

default orient - matrices work, dates fail as they're exported as epoch times, imported as ints
```python
def serde_func(df):
    orient=None
    return pd.read_json(df.to_json(orient=orient), orient=orient)

def non_eq_disp(*dfs):
    for df in dfs:
        display(df)
    for df in dfs:
        display(df.dtypes)

def run_test(df):
    new_df = serde_func(df)
    eq = df.equals(new_df)
    print(f'Equal: {df.equals(new_df)}')
    if not eq:
        non_eq_disp(df, new_df)
        
print('Date test')
df = pd.DataFrame(
    {
        "A": [datetime.datetime(year, 1, 1) for year in range(2020, 2023)],
        "B": [1, 2, 3],
    }
)
run_test(df)

print('Numeric test')
df = pd.DataFrame([[3, 3], [4, 2]])
run_test(df)
```

default orient output
image

default orient, specify date output format - matrices work, dates interpreted as strings
```python
def serde_func(df):
    orient=None
    return pd.read_json(df.to_json(orient=orient, date_format='iso'), orient=orient)

def non_eq_disp(*dfs):
    for df in dfs:
        display(df)
    for df in dfs:
        display(df.dtypes)

def run_test(df):
    new_df = serde_func(df)
    eq = df.equals(new_df)
    print(f'Equal: {df.equals(new_df)}')
    if not eq:
        non_eq_disp(df, new_df)
        
print('Date test')
df = pd.DataFrame(
    {
        "A": [datetime.datetime(year, 1, 1) for year in range(2020, 2023)],
        "B": [1, 2, 3],
    }
)
run_test(df)

print('Numeric test')
df = pd.DataFrame([[3, 3], [4, 2]])
run_test(df)
```

specified date format output
image

ektar:
One last note on schema - Schema provides hints to pandas to interpret columns correctly (e.g. dates), also captures table structure in a way the others can't.  E.g. indexing:
```python
df = pd.DataFrame(
    {
        "A": [datetime.datetime(year, 1, 1) for year in range(2020, 2023)],
        "B": [1, 2, 3]
    },
    index=((1, 1), (1, 2), (2, 1))
)
display(df)
json_str=df.to_json()
print(json_str)
display(pd.read_json(df.to_json()))
print()
json_str=df.to_json(orient='table')
print(json_str)
display(pd.read_json(json_str, orient='table'))
```

output:
image

ektar:
As much as I like schemas, orient=table just seems not ready yet, fine to implement just casting dates to iso str and user can parse later - at least will stop current crashing behavior.

If that sounds right, comment and question:
- Earlier there was a question why do serialize->json load then reverse in serialization... this is just to fit with the overall framework of how json serialization seems to be working in param - all serialization funcs are expected to return a dict, but panda's to_json returns a string.  And purpose of using panda's to_json is so that the logic doesn't need to get implemented here.  I needed to do reverse in deser before due to table orient with schema.  I thought I could just do dataframe load from dict on the return, but it turns out the index gets cast to string instead of ints (tests failed), so needed to do the full reverse round trip to get types correct.  This seemed simplest change - thoughts?
- Current test of df.equals will fail - original dataframe will have datetimes, new dataframe will have string column.  Preferences for how to change test?  I just updated to exclude testing that date cols are equal.

param/__init__.py
```diff
@@ -1567,14 +1567,16 @@ def _validate(self, val):
     def serialize(cls, value):
         if value is None:
             return 'null'
-        return value.to_dict('records')
+        import json
```
jlstevens, line 1570: As `json` is in the standard library, I would just import it at the top of the file instead of inlining it.

jlstevens:
> all serialization funcs are expected to return a dict, but panda's to_json returns a string. 

It is possible that I would need to run the code to understand the real issue, but this statement isn't quite correct. The idea of the `serialize` and `deserialize` methods is to use an intermediate representation of some sort where the output of `serialize` is JSON serializable. As strings (including JSON strings!) are JSON serializable, I don't think you need to use `json.loads` or `json.dumps` for this to work. Essentially, you would be embedding pandas JSON output in the param JSON output which is a little weird, but should work.

If it weren't for potential performance concerns, I would indeed prefer the way you have done it already though!

param/__init__.py
ektar: Agreed, done

ektar:
> > all serialization funcs are expected to return a dict, but panda's to_json returns a string.
> 
> It is possible that I would need to run the code to understand the real issue, but this statement isn't quite correct. The idea of the `serialize` and `deserialize` methods is to use an intermediate representation of some sort where the output of `serialize` is JSON serializable. As strings (including JSON strings!) are JSON serializable, I don't think you need to use `json.loads` or `json.dumps` for this to work. Essentially, you would be embedding pandas JSON output in the param JSON output which is a little weird, but should work.
> 
> If it weren't for potential performance concerns, I would indeed prefer the way you have done it already though!

I agree it looks weird - I was copying original [implementation](https://github.com/holoviz/param/blob/3784a1dc7a677c60648511e87171727232c42920/param/__init__.py#L1566-L1577) to have same return type.  I think intent is that the final json be standard, not have long strings that are intended to be further processed.  This is used by upstream [jsonserializer](https://github.com/holoviz/param/blob/3784a1dc7a677c60648511e87171727232c42920/param/serializer.py#L96-L115) - serializes all components then calls dumps.

ektar:
@jlstevens @philippjfr - Any changes needed?  Happy to implement

droumis:
@ektar, do you intend on completing this?

maximlt:
In https://github.com/holoviz/param/pull/600/commits/7e5098c1f1dec7c8f79d27b41cbb1f5f72dd916c I've slightly updated the tests to remove code that was handling timezone conversion. This was failing when I ran it locally (newer Pandas version?), removing that code got the test to pass.

PR #601, opened 2022-02-03
CI: Run more tests on Python 3.10

maximlt:
The CI is already running on Python 3.10 but was avoiding running a few tests because at the time it was set a few packages were not yet available on Python 3.10


maximlt:
Ready to be merged.

philippjfr:
Thanks @maximlt!

PR #615, opened 2022-04-11
Look at whole spec name when determing dynamic watchers

### LINKED ISSUE #614: Depends is sensitive to widget name
#### ALL software version info
``` 
panel    0.13.0a46.post46+g255f41c9
param    1.12.1a1.post13+ga11767e
```

#### Description of expected behavior and the observed behavior
I have the following minimal example, which contains a function for loading data and then plotting data based on the `picker` widget.

I have added  `date_picker` in my example, which does nothing in the MRE but is used in my actual code. By having this widget with this name, the plot function is only called once, and then it becomes stale. 

https://user-images.githubusercontent.com/19758978/161953202-5f9fa236-c3be-47a0-8e91-2599394a02a8.mp4

If I remove the widget from `depends`, everything works as expected. But even more strange is it also works if I rename the widget from `date_widget` to another name like `adate_widget`.

https://user-images.githubusercontent.com/19758978/161953591-bd9c5f2b-8a81-4f21-a012-131b68b8d6c6.mp4

#### Complete, minimal, self-contained example code that reproduces the issue

``` python
import panel as pn
import param


class Example(param.Parameterized):
    da = param.Number(default=1)

    def __init__(self, **params):
        # Not used in this MRE
        self.date_picker = pn.widgets.MultiSelect(
            options=list("ABCD"),
        )
        self.picker = pn.widgets.MultiSelect(
            options=list("ABCD"),
        )
        super().__init__(**params)

    @param.depends(
        "date_picker.value",  
        "picker.value",
        watch=True,
    )
    def load_data(self):
        self.da += 1  # To trigger plot_time

    @param.depends("da")
    def plot_time(self):
        return self.picker.value

    def __panel__(self):
        return pn.Row(self.picker, self.plot_time)


pn.panel(Example().__panel__).servable()


```

Hoxbro:
Fixes #614 

TODO: 
- [x] Needs to add a unit test

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1698,7 +1698,7 @@ class Parameters(object):
         for method, queued, on_init, constant, dynamic in type(obj).param._depends['watch']:
             # On initialization set up constant watchers; otherwise
             # clean up previous dynamic watchers for the updated attribute
-            dynamic = [d for d in dynamic if attribute is None or d.spec.startswith(attribute)]
+            dynamic = [d for d in dynamic if attribute is None or d.spec.split(".")[0] == attribute]
             if init:
                 constant_grouped = defaultdict(list)
                 for dep in _resolve_mcs_deps(obj, constant, []):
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks! Happy to merge but a unit test would indeed be nice.

Hoxbro:
I'm not sure if it is the right place I added the unit test. I will move it if needed.

jbednar:
Thanks! That looks like just the right fix.

PR #620, opened 2022-04-27
Ensure that arguments to the constructor trigger watchers

### LINKED ISSUE #616: on_init should only trigger if the parameter is set in the constructor
My hope when suggesting #513 was that a parameter would be triggered if it were passed into the constructor. It appears that the `on_init` argument always triggers on init regardless of whether the parameter was passed or not.

In the following example when `Test` is instantiated without any arguments then it still triggers the method that depends on `s`. 

image

I think it should work equivalent to this:

```python
def __init__(self, **params):
        super().__init__(**params)
        if 's' in params:
            self.param.trigger('s')
```

It may be useful to always trigger a method on initialization, but since that behavior is independent of any parameters then it should be a separate decorator that isn't associated with parameters. For example:

```python
    s = param.String(default='default')
    
    @param.on_init
    def run_on_initialization(self):
        print(self.s)
```

Having the `on_init` argument in the `depends` decorator makes it seem like it should be associated with the parameters that are also passed in with `depends`.

philippjfr:
Fixes #616

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -449,8 +449,12 @@ def depends(func, *dependencies, **kw):
             group[0].owner.param.watch(cb, [dep.name for dep in group])
 
     _dinfo = getattr(func, '_dinfo', {})
-    _dinfo.update({'dependencies': dependencies,
-                   'kw': kw, 'watch': watch, 'on_init': on_init})
+    _dinfo.update({
+        'dependencies': dependencies,
+        'kw': kw,
+        'watch': watch,
+        'on_init': on_init
+    })
 
     _depends._dinfo = _dinfo
 
@@ -1966,7 +1970,7 @@ class Parameters(object):
         return pdict
 
 
-    def trigger(self_, *param_names):
+    def trigger(self_, *param_names, **kw):
         """
         Trigger watchers for the given set of parameter names. Watchers
         will be triggered whether or not the parameter values have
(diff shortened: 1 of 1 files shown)


param/parameterized.py
```diff
@@ -1997,7 +2002,7 @@ def _update_event_type(self_, watcher, event, triggered):
         Returns an updated Event object with the type field set appropriately.
         """
         if triggered:
-            event_type = 'triggered'
+            event_type = 'triggered' if triggered == True else triggered
```
jbednar, line 2005: The docs should explain what values for `triggered` will be accepted, because otherwise it's very hard to reason about what values event_type might end up with.

jbednar (commented):
I can't follow the logic, but maybe if you add tests I'll be able to?

param/parameterized.py
jbednar: I.e. is it simply a Boolean? If so, event_type can be 'triggered', 'changed', or False? Seems odd.

philippjfr:
Don't review this yet, I was on a call with Scott and just wanted to push something up.

maximlt:
@philippjfr do you need this in 2.0?

jbednar:
@philippjfr , please review and either finish or replace with something suitable. Thanks!

philippjfr:
I no longer think this should be merged.

PR #621, opened 2022-05-01
Fix the test workflow

maximlt:
Noticed two bad issues:
* Not actually running on Python 2 :(
* Not running on the required version on Macos

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -21,7 +21,6 @@ jobs:
       run:
         shell: bash -l {0} 
     env:
-      PYTHON_VERSION: ${{ matrix.python-version }}
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


maximlt:
I've done some clean-up and added the package `tox-gh-actions` as a dependency of the test workflow. I used the CI of this package as an inspiration to fix the issues reported in the original post, since it's running on multiple platforms and lots of Python versions. However I don't why exactly the CI was not running on Python 2 or on the correct Python versions on Macos (this was fine on Windows and Linux), one of the things I changed in this commit (https://github.com/holoviz/param/pull/621/commits/ee5abcd25462d41bf79292dd6a2f33bc326423fb) fixed that though.

I'll merge now, the CI is failing on Python 2.7 (at least) because of a `yield from` that raises an invalid syntax error. This unfortunately means that the latest releases of Param could be installed on Python 2.7 (as declared in the `setup.py`) but were actually not compatible with Python 2.7. The good news in a way is that we have had no bug report so Param isn't used much on Python 2.7 (at least no by users who don't pin their dependencies, or report bugs...).

cc @jbednar @philippjfr @jlstevens

jbednar:
Yikes! Good catch. Let's release a final py27 compatible 1.x version shortly before releasing 2.0, and update the release notes for any release not compatible with 2.7.

PR #624, opened 2022-05-10
Ensure async functionality does not cause python2 syntax errors

philippjfr:


maximlt:
This looks good to me. I'll try to add some tests because I think these code paths aren't tested at all.

maximlt:
@philippjfr this is ready to me, the tests I've added are really simple and are just meant to get 100% coverage of the `_async` module.

PR #625, opened 2022-05-11
Allow (de)serializing CalendarRange and DateRange Parameters

maximlt:
I got the following error while working on a Lumen app that has a DateRange filter widget synced to the URL:
```
  File "/Users/mliquet/miniconda3/envs/lumen-dev37/lib/python3.7/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type datetime is not JSON serializable
```

We found out that the `DateRange` and `CalendarRange` parameters inherit from the `Tuple` parameter (de)serialization methods, while they should override these methods as Python's `json` module can't serialize datetime objects (`import datetime as dt, json; json.dumps(dt.datetime.now())` raises the TypeError mentioned above).

This PR adds these methods to the two parameters, and test them. I've added tests, including tests that use `np.datetime64` and `pd.Timestamp` objects since they appeared to be supported (I learnt that `pd.Timestamp` objects inherits from `datetime.datetime`).

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -2127,6 +2127,37 @@ class DateRange(Range):
                              "is before start datetime %s." %
                              (self.name, val[1], val[0]))
 
+    @classmethod
+    def serialize(cls, value):
+        if value is None:
+            return 'null'
+        # List as JSON has no tuple representation
+        serialized = []
+        for v in value:
+            if not isinstance(v, (dt.datetime, dt.date)): # i.e np.datetime64
+                v = v.astype(dt.datetime)
+            # Separate date and datetime to deserialize to the right type.
+            if type(v) == dt.date:
+                v = v.strftime("%Y-%m-%d")
+            else:
+                v = v.strftime("%Y-%m-%dT%H:%M:%S.%f")
+            serialized.append(v)
+        return serialized
+
+    def deserialize(cls, value):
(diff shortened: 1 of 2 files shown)


jbednar (approved):
Looks good to me!

maximlt:
Actually any container Parameter that holds a datetime object won't be able to be serialized due to the limitations of the basic Python JSON encoder. As such the changes made in this PR aren't yet enough to fix the issue I observed in Lumen, which uses the `DateRangeSlider` widget offered by Panel whose `value` attribute isn't a `DateRange` parameter but a `Tuple`.

The way forward may be to subclass `json.JSONEncoder` and `json.JSONDecoder` to implement custom encoders/decoders that handle datetime values. There must be some existing implementations of that in the Python ecosystem.

maximlt:
JSON has no datetime representation so it's actually not possible to directly decode a Python datetime value encoded to a JSON string. Seems like a limitation that should be mentioned in the documentation. The Lumen/Panel issue needs to be fixed separately.

philippjfr:
This seems sufficient for the specific bug we are fixing here. In the longer run it would be nice to work out some more general solution but I suspect we'd need composite parameter types to make that feasible.

PR #627, opened 2022-05-13
Improve DateRange validation

maximlt:
Noticed that the `DateRange` parameter was not checking that the value provided is a `tuple`, which it should be given that `DateRange` inherits from Tuple.

I've also changed the error message reported when one of the values is not of the correct type:

    DateRange parameter 'a' only takes date/datetime values, not type float.

This is slightly different formatting compared to the messages reported by other Parameters, e.g. with a List:

    List parameter None must be a list, not an object of type <class 'str'>.

Notice `of type float` vs. `of type <class 'str'>`. If the latter is preferred, I could update the error messages of the other Parameters.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -2112,9 +2112,15 @@ class DateRange(Range):
     """
 
     def _validate_value(self, val, allow_None):
+        # Cannot use super()._validate_value as DateRange inherits from
+        # NumericTuple which check that the tuple values are numbers and
+        # datetime objects aren't numbers.
         if allow_None and val is None:
             return
 
+        if not isinstance(val, tuple):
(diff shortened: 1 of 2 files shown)


maximlt:
@jlstevens ready for review 🙃

jlstevens (approved):

jlstevens:
Looks good to me!

jbednar (approved):
Looks good to me, thanks! I agree that it would be nice to make messages more consistent, but don't have a terribly strong opinion between the ways that type information is printed. I think on balance I'd prefer a simpler, more-English like version (e.g. `type float`) over a syntax-y version (e.g. `<class 'float'>`).

philippjfr:
Agreed, consistency would be good but for now this is an improvement.

Issue #618, opened 2022-04-13
Provide more control over the call order of watched methods


sdc50:
Between versions 1.11.1 and 1.12.1 the order in which methods that are decorated with `param.depends` are triggered was changed. 

```python
In [1]: import param

In [2]: class Test(param.Parameterized):
   ...:     s = param.String(default='default')
   ...: 
   ...:     @param.depends('s', watch=True)
   ...:     def on_s_update(self):
   ...:         print(self.s)
   ...: 
   ...: class Sub(Test):
   ...: 
   ...:     @param.depends('s', watch=True)
   ...:     def more_on_s_update(self):
   ...:         print(self.s.upper())
   ...: 
   ...: s = Sub()
   ...: s.s = 'test'
test
TEST

In [3]: param.__version__
Out[3]: '1.11.1'
```
vs.

![image](https://user-images.githubusercontent.com/6454472/163219221-ceadcea5-d7ba-41ea-a6c6-5ad0f27ae376.png)

Without a way to explicitly control the call order much of my code was in a delicate balance, and this has just wreaked havoc on it all.

There should be a way explicitly control the call order of depended on methods.  In the past I had discussed with @philippjfr about adding a `precedence` argument to the `depends` decorator (and the `watch` method) so that callbacks can be sorted and called in a specified order when a parameter is triggered.

philippjfr:
Will reopen since the precedence proposal hadn't been added yet.

philippjfr:
I'm somewhat hesitant to add `precedence` as a reserved kwarg to `depends` since it's not an unlikely argument name. Any other suggestions?

jbednar:
I guess `priority` has the same issue. But any compound word should be ok, e.g. call_prority or call_order or calling_order.

maximlt:
Oh. I looked that up but could not find it, is it documented somewhere that there are reserved kwargs? Does this just apply to when `@param.depends` is used to decorate a function? As in:

```python
import param

class P(param.Parameterized):

    x = param.String()
    w = param.String()


p = P()

@param.depends(x=p.param.x, watch=p.param.w)
def foo(x, watch):
    print(x)

p.x = 'test'
```

philippjfr:
I don't know if we explicitly discuss reserved kwargs, certainly we do discuss `watch` and now we'd be proposing to add another argument. I don't think there are any others but I might be wrong.

maximlt:
Btw this is the error I get with the snippet above:

```python
Traceback (most recent call last):
[...]

PR #628, opened 2022-05-17
Fix regression in param depends execution ordering

philippjfr:
Fixes https://github.com/holoviz/param/issues/618


jbednar (approved):
Looks good to me, assuming this test would pass in the previous Param version.

philippjfr:
>Looks good to me, assuming this test would pass in the previous Param version.

It would pass in param<1.12

PR #629, opened 2022-05-18
Support comparing date-like objects

maximlt:
Discovered this bug while working on https://github.com/holoviz/lumen/pull/276

The `Comparator` class didn't support `datetime.date`, `datetime.datetime` and `np.datetime64` object. This meant that, if a parameter value was set to its old value, Param would still see it as a new value and would let potential watchers be executed, while they should not. This caused problems for Panel bidirectional links which don't check for equality but rely on Param.

I see that other types aren't supported, like Numpy Arrays, Pandas DataFrame, or frozensets. If more types should be added, let's come up with a list and add them in another PR.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -27,7 +27,7 @@ import collections
 
 from .parameterized import (
     Parameterized, Parameter, String, ParameterizedFunction, ParamOverrides,
-    descendents, get_logger, instance_descriptor, basestring)
+    descendents, get_logger, instance_descriptor, basestring, dt_types)
 
 from .parameterized import (batch_watch, depends, output, script_repr, # noqa: api import
                             discard_events, edit_constant, instance_descriptor)
@@ -46,16 +46,6 @@ try:
     __version__ = str(Version(fpath=__file__, archive_commit="$Format:%h$", reponame="param"))
 except:
     __version__ = "0.0.0+unknown"
-
-
-dt_types = (dt.datetime, dt.date)
-
-try:
-    import numpy as np
-    dt_types = dt_types + (np.datetime64,)
-except:
-    pass
-
 
 try:
     import collections.abc as collections_abc
@@ -201,12 +191,13 @@ def guess_param_types(**kwargs):
                 params[k] = Tuple(**kws)
         elif isinstance(v, list):
             params[k] = List(**kws)
-        elif isinstance(v, np.ndarray):
-            params[k] = Array(**kws)
         else:
+            from numpy import ndarray
             from pandas import DataFrame as pdDFrame
             from pandas import Series as pdSeries
 
+            if isinstance(v, ndarray):
+                params[k] = Array(**kws)
             if isinstance(v, pdDFrame):
                 params[k] = DataFrame(**kws)
             elif isinstance(v, pdSeries):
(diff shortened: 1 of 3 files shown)


param/__init__.py
```diff
@@ -201,12 +191,13 @@ def guess_param_types(**kwargs):
                 params[k] = Tuple(**kws)
         elif isinstance(v, list):
             params[k] = List(**kws)
-        elif isinstance(v, np.ndarray):
-            params[k] = Array(**kws)
         else:
+            from numpy import ndarray
```
jbednar, line 195: This function seems to assume that numpy and pandas can be imported. Seems like that ought to be checked, not assumed?

maximlt: Yes you're right, this utility was assuming that pandas and numpy were always available. See https://github.com/holoviz/param/pull/630

maximlt:
@philippjfr I've seen you added the `Comparator` class, in https://github.com/holoviz/param/pull/629/commits/6fe66a8b84b0274c267d9bc5568e2db275327d7a I added an identity check to short cut the equality comparisons. This I think would have fixed the issue I was having with bidirectional links in Panel, since the same attribute object is shared between the linked Parameterized instances.

However this test is failing because the code did not handle any other type than the ones declared in the `equalities` dictionary.

```
  _____________ TestWatch.test_triggered_when_unchanged_complex_type _____________
  
  self = <tests.API1.testwatch.TestWatch testMethod=test_triggered_when_unchanged_complex_type>
  
      def test_triggered_when_unchanged_complex_type(self):
          def accumulator(change):
              self.accumulator += 1
      
          obj = SimpleWatchExample()
          obj.param.watch(accumulator, 'a')
          subobj = object()
          obj.a = subobj
          self.assertEqual(self.accumulator, 1)
          obj.a = subobj
  >       self.assertEqual(self.accumulator, 2)
  E       AssertionError: 1 != 2
```

param/__init__.py
philippjfr: Fixed in a separate PR.

param/parameterized.py
```diff
@@ -1398,11 +1407,14 @@ class Comparator(object):
...
         bytes: operator.eq,
-        type(None): operator.eq
+        type(None): operator.eq,
     }
+    equalities.update({dtt: operator.eq for dtt in dt_types})
 
     @classmethod
     def is_equal(cls, obj1, obj2):
+        if obj1 is obj2:
+            return True
```
philippjfr, line 1417: While this is of course technically correct it is a major change in behavior. Since complex types cannot be compared properly in many cases this will mean that objects that have been changed inplace now no longer trigger an event. Please open a new issue and let's not mix this change into this PR.

philippjfr, line 1417:
```suggestion
```

philippjfr:
I've removed the identity check, not because I think we shouldn't do it but because it's a change in behavior and we need to consider it carefully. Please open a new issue or PR to discuss further.

philippjfr:
MacOS tests passed but coveralls errored for some reason. Merging.

maximlt:
This was a codecov issue (https://github.com/codecov/codecov-action/issues/745) that is now resolved.

PR #630, opened 2022-05-19
Don't assume numpy/pandas is available in `guess_param_types` 

maximlt:
Implementing the suggestion in https://github.com/holoviz/param/pull/629#discussion_r876410658, **needs https://github.com/holoviz/param/pull/629 to be merged first**

`guess_param_types` is actually not used at all across HoloViz and not documented. The PR that added this utility - https://github.com/holoviz/param/pull/317 - says that it "has cropped up several times and has resulted in some level of code duplication" and points to HoloViews [Streams.define](https://github.com/pyviz/holoviews/blob/master/holoviews/streams.py#L85) classmethod which indeed could make use of this utility, since it's basically inlining a partial implementation of it.

https://github.com/holoviz/param/pull/317 provides more details about the usefulness of this utility, and on the conversions it doesn't yet support.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -203,24 +203,22 @@ def guess_param_types(**kwargs):
         else:
             try:
                 from numpy import ndarray
-            except ImportError:
-                pass
-            else:
                 if isinstance(v, ndarray):
                     params[k] = Array(**kws)
                     continue
+            except ImportError:
+                pass
             try:
                 from pandas import DataFrame as pdDFrame
                 from pandas import Series as pdSeries
-            except ImportError:
-                pass
-            else:
                 if isinstance(v, pdDFrame):
                     params[k] = DataFrame(**kws)
                     continue
                 elif isinstance(v, pdSeries):
                     params[k] = Series(**kws)
                     continue
+            except ImportError:
+                pass
             params[k] = Parameter(**kws)
 
     return params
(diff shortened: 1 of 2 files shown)


param/__init__.py
```diff
@@ -195,24 +185,34 @@ def guess_param_types(**kwargs):
...
+                pass
             else:
-                params[k] = Parameter(**kws)
+                if isinstance(v, pdDFrame):
+                    params[k] = DataFrame(**kws)
+                    continue
+                elif isinstance(v, pdSeries):
+                    params[k] = Series(**kws)
+                    continue
+            params[k] = Parameter(**kws)
```
jbednar, line 215: Thanks! I have trouble following the logic flow; is there any reason to use `else` here as opposed to putting that code directly into the `try` block after the import?

maximlt: I've removed the `else` blocks by putting the `isinstance` checks in the `try` blocks, as you suggested. The reason why I did it this way in the first place was that I usually try to have as less code as possible in `try` blocks. This made the code arguably ugly though!

The code is still a little ugly after the change, it's this way because I want to avoid as much as possible to have to import numpy or pandas if not necessary.

jbednar: I think it's slightly easier to follow now, though yes, definitely good to reduce `try` code where possible.

jbednar (approved):

PR #631, opened 2022-05-23
WIP: raise ValueError if param.depends is used without arguments or without defining dependencies

maximlt:
To raise an error in these cases where the `depends` decorator is used either without arguments or without providing any dependencies (same on methods on Parameterized classes).

```python
@param.depends
def foo():
    pass

# ValueError: The depends decorator on the callable 'foo' cannot be used without arguments.

@param.depends()
def foo():
    pass

# ValueError: The depends decorator on the callable 'foo' require dependencies to be defined.
```

WIP because as such the following would raise an error (since internally Panel uses the `depends` decorator when it binds a callable) while the approach retained is to warn in this case (https://github.com/holoviz/panel/pull/3545).

```python
import panel

def foo(val):
    return val

pn.panel(pn.bind(foo, val='bob'))
```

Emitting a warning instead of raising an error when no dependencies are defined may be a better idea (that Panel could silence).

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -372,8 +372,7 @@ def get_method_owner(method):
     return method.__self__ if sys.version_info.major >= 3 else method.im_self
 
 
-@accept_arguments
-def depends(func, *dependencies, **kw):
+def depends(*dependencies, **kw):
     """
     Annotates a function or Parameterized method to express its
     dependencies.  The specified dependencies can be either be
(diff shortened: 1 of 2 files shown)


philippjfr:
I don't think this is correct, `@param.depends()` is documented and the behavior is well defined in the [user guide](https://param.holoviz.org/user_guide/Dependencies_and_Watchers.html#watch-false-dependencies):

> Conversely, if you want to declare that a given method does not depend on any parameters at all, you can use @param.depends().

So I'm perfectly happy to disallow `param.depends` without calling it but we should not disallow `param.depends()`.

jbednar:
I'm not sure what the interaction with Panel is, but I agree with @philippjfr that `depends()` is not an error but `depends` is.

maximlt:
I no longer see much value in the changes I made in this PR. As stated above`@param.depends()` is valid. `@param.depends` could either raise an error or be accepted and made equivalent to `@param.depends()`. As I don't want to decide that now and this is definitely super low priority, I'm just going to close this. Let's revisit when this comes up again, if ever!

PR #632, opened 2022-05-25
Ensure named_objs does not fail on unhashable objects

philippjfr:
Not all objects are hashable so the `named_objs` implementation has started causing some errors (e.g. because mpl colormaps aren't hashable anymore). For unhashable objects we try to reverse map the objects by identity.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -144,7 +144,7 @@ def named_objs(objlist, namesdict=None):
 
     for obj in objlist:
         if reverse_lookup and hashable(obj) in reverse_lookup:
-            k = objtoname[hashable(obj)]
+            k = reverse_lookup[hashable(obj)]
         elif any(obj is v for (_, v) in unhashables):
             k = [k for (k, v) in unhashables if v is obj][0]
         elif hasattr(obj, "name"):
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -133,18 +133,18 @@ def named_objs(objlist, namesdict=None):
     """
     objs = OrderedDict()
 
-    reverse_lookup = {}
+    objtoname = {}
     unhashables = []
     if namesdict is not None:
         for k, v in namesdict.items():
             try:
-                reverse_lookup[hashable(v)] = k
+                objtoname[hashable(v)] = k
             except TypeError:
                 unhashables.append((k, v))
 
     for obj in objlist:
-        if reverse_lookup and hashable(obj) in reverse_lookup:
-            k = reverse_lookup[hashable(obj)]
+        if objtoname and hashable(obj) in objtoname:
+            k = objtoname[hashable(obj)]
         elif any(obj is v for (_, v) in unhashables):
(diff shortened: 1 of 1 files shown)


jbednar:
>y main confusion here is why it even attempts a reverse mapping at all if an explicit namesdict has already been provided.

Even if an explicit namesdict has been provided, it isn't required to be _complete_. Where that matters is  if `check_on_set=False` for this Selector, in which case someone can do `Obj.x=val`, for any `val`, whether that value is in the `namesdict` or not. The namesdict will then be incomplete, and so `named_objs` attempts to complete it by providing _some_ representation for every item in the object list.  Here you could maybe skip any unhashable items while still attempting to provide a representation for hashable items? But you'd then need the rest of the code to gracefully handle the case of an object not being in the `namesdict`.

philippjfr:
>But you'd then need the rest of the code to gracefully handle the case of an object not being in the namesdict.

The rest of the code being what in this case? Panel already does and afaik that's the only thing that actually uses the names.

param/__init__.py
```diff
@@ -132,12 +133,20 @@ def named_objs(objlist, namesdict=None):
...
-        objtoname = {hashable(v): k for k, v in namesdict.items()}
+        for k, v in namesdict.items():
+            try:
+                objtoname[hashable(v)] = k
+            except TypeError:
+                unhashables.append((k, v))
 
     for obj in objlist:
-        if namesdict is not None and hashable(obj) in objtoname:
+        if objtoname and hashable(obj) in objtoname:
```
jbednar, line 146:
```suggestion
        if hashable(obj) in objtoname:
```

(checking that the dict is non-empty doesn't seem like it achieves anything)

param/__init__.py
```diff
@@ -132,12 +133,20 @@ def named_objs(objlist, namesdict=None):
...
             k = objtoname[hashable(obj)]
+        elif any(obj is v for (_, v) in unhashables):
+            k = [k for (k, v) in unhashables if v is obj][0]
```
jbednar, line 149: This seems to be handling the case where the same unhashable object is in the list multiple times with different names, but then it returns only the first name?  If so what's the use of `unhashables` being a list rather than a dict?

philippjfr: Well it can't be a dict since the obj is unhashable.

param/__init__.py
philippjfr: Sure, it avoids the potentially expensive `hashable` conversion.

param/__init__.py
jbednar: A k->v dict.

jbednar: Ah, I guess that would make for a nicer data structure but then awkward to use. Never mind.

param/__init__.py
jbednar: Ok

jbednar (approved):
As revised this seems ok to merge, though really Selector needs to just store proper mappings as outlined in other PRs and issues...

Issue #608, opened 2022-02-22
Math error in BinaryPower example in User Guide 


tzelleke:
The `BinaryPower` class in the *Specialized Parameter types* subsection of the user guide does not validate binary powers but merely divisibility by 2.

[https://param.holoviz.org/user_guide/Parameters.html#specialized-parameter-types
](https://param.holoviz.org/user_guide/Parameters.html#specialized-parameter-types
)

Another issue is that this `Parameter` class does not respect `None` values.

If you want to keep this example based on `param.Parameter` you might update it to:

```python
    def _validate_value(self, val, allow_None):
        if allow_None and val is None:
            return
        
        super(BinaryPower, self)._validate_value(val, allow_None)
        if not isinstance(val, numbers.Number):
            raise ValueError("BinaryPower parameter %r must be a number, "
                             "not %r." % (self.name, val))
        
        if (not float(val).is_integer()) or (val < 2) or (bin(int(val)).count("1") is not 1):
            # in Python >= 3.10 use `int(val).bit_count() is not 1`
            raise ValueError("BinaryPower parameter %r must be a power of 2, "
                             "not %r." % (self.name, val))

```

I would rather suggest to have the example based on `param.Integer` with `default=2`... but that does not read so well with the way this section is structured currently ?

jlstevens:
Thanks for reporting this!

In https://github.com/holoviz/param/pull/634 I address this by keeping the correct code and changing the example to one that only allows even integers. The purpose of the docs here is to show that parameters can be defined easily and not to have people think too much about the precise validation logic itself.

jlstevens:
Addressed in #634 which is now merged.

PR #634, opened 2022-05-30
Renamed incorrect BinaryPower example to EvenInteger

jlstevens:
Addresses #608 by keeping the simple code example but renaming the parameter to `EvenInteger` which expresses the actual check that is performed.


jlstevens:
@jbednar ready for review/merge.

PR #638, opened 2022-06-14
Preserve existing Random seed behavior in Python 3.11

### LINKED ISSUE #602: numbergen - DeprecationWarning: Seeding based on hashing is deprecated since Python 3.9
From Python 3.9 onward a deprecation warning is emitted if a seed is set from a tuple, as done here in `numbergen`: https://github.com/holoviz/param/blob/32060b24736fa732417103a12ea05f941cf1f660/numbergen/__init__.py#L310

```
 _________________ ERROR collecting tests/API1/testnumbergen.py _________________
tests/API1/testnumbergen.py:4: in <module>
    import numbergen
numbergen/__init__.py:277: in <module>
    class TimeAwareRandomState(TimeAware):
numbergen/__init__.py:310: in TimeAwareRandomState
    default=random.Random((500,500)), doc=
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/random.py:123: in __init__
    self.seed(x)
/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/random.py:158: in seed
    _warn('Seeding based on hashing is deprecated\n'
E   DeprecationWarning: Seeding based on hashing is deprecated
E   since Python 3.9 and will be removed in a subsequent version. The only 
E   supported seed types are: None, int, float, str, bytes, and bytearray.
```

@jlstevens had some idea about how to fix that. Note that if the warning is eventually turned into an error in future Python versions `numbergen` will no longer be importable, since this `randon.Random((500, 500))` is executed on import time.

musicinmybrain:
Fixes #602. See https://github.com/holoviz/param/issues/602#issuecomment-1155147795 for rationale and caveats.

You might instead prefer to use a fixed seed of `13594734499543488215`. That would change the current initial state on older CPython versions (particularly Python 2.7) and on 32-bit builds of CPython; these would become consistent with the initial state on 64-bit builds of supported CPython 3.x versions.

### DIFF
--- a/numbergen/__init__.py
+++ b/numbergen/__init__.py
@@ -8,6 +8,7 @@ import hashlib
 import struct
 import fractions
 
+from ctypes import c_size_t
 from math import e,pi
 
 import param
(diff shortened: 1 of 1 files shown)


musicinmybrain:
Added an explanatory comment and force-pushed.

jlstevens:
Looks good to me! I'll let @jbednar make the final decision to merge though.

PR #655, opened 2022-10-25
Allow JSON serialization to work with json.dumps

### LINKED ISSUE #578: Date does not work with allow_None

Hoxbro:
Fixes #578

The changes in #582 made it possible to round-trip serialize <-> deserialize within param. 

A completely valid way to deserialize is to have the serialized values created outside param like this:
``` python
import json
import param

class User(param.Parameterized):
    A = param.Date(default=None)

data = json.dumps({"A": None})    # gives: '{"A": null}' <- note that null is not a string
User.param.deserialize_parameters(data)
```

Which raised this error before these changes:
``` python-traceback
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3378, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "/tmp/ipykernel_70470/342607310.py", line 8, in <module>
    User.param.deserialize_parameters(data)
  File "/home/shh/Development/holoviz/repos/param/param/parameterized.py", line 2121, in deserialize_parameters
    return serializer.deserialize_parameters(self_or_cls, serialization, subset=subset)
  File "/home/shh/Development/holoviz/repos/param/param/serializer.py", line 113, in deserialize_parameters
    deserialized = pobj.param[name].deserialize(value)
  File "/home/shh/Development/holoviz/repos/param/param/__init__.py", line 1954, in deserialize
    return dt.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f")
TypeError: strptime() argument 1 must be str, not None
```

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1022,12 +1022,12 @@ class Tuple(Parameter):
     @classmethod
     def serialize(cls, value):
         if value is None:
-            return 'null'
+            return None
         return list(value) # As JSON has no tuple representation
 
     @classmethod
     def deserialize(cls, value):
-        if value == 'null':
+        if value == 'null' or value is None:
             return None
         return tuple(value) # As JSON has no tuple representation
 
@@ -1483,12 +1483,12 @@ class Array(ClassSelector):
     @classmethod
     def serialize(cls, value):
         if value is None:
-            return 'null'
+            return None
         return value.tolist()
 
     @classmethod
     def deserialize(cls, value):
-        if value == 'null':
+        if value == 'null' or value is None:
             return None
         from numpy import asarray
         return asarray(value)
@@ -1570,12 +1570,12 @@ class DataFrame(ClassSelector):
     @classmethod
     def serialize(cls, value):
         if value is None:
-            return 'null'
+            return None
         return value.to_dict('records')
 
     @classmethod
     def deserialize(cls, value):
-        if value == 'null':
+        if value == 'null' or value is None:
             return None
         from pandas import DataFrame as pdDFrame
         return pdDFrame(value)
@@ -1942,14 +1942,14 @@ class Date(Number):
     @classmethod
     def serialize(cls, value):
         if value is None:
-            return 'null'
+            return None
         if not isinstance(value, (dt.datetime, dt.date)): # i.e np.datetime64
             value = value.astype(dt.datetime)
         return value.strftime("%Y-%m-%dT%H:%M:%S.%f")
 
     @classmethod
     def deserialize(cls, value):
-        if value == 'null':
+        if value == 'null' or value is None:
             return None
         return dt.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f")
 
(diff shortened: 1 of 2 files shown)


param/__init__.py
```diff
@@ -1022,12 +1022,12 @@ def _validate(self, val):
     @classmethod
     def serialize(cls, value):
         if value is None:
-            return 'null'
+            return None
```
philippjfr, line 1025: serialize should not actually return the JSON form but simply a form that can be cleanly serialized. So I don't think this is right.

Hoxbro: I don't understand your comment. 

As far as I see this change does not return the JSON form, but a form that can be serialized.

philippjfr: Sorry I'm just now looking at the handling of None in the serialize/deserialize methods and I believe it to be totally wrong based on the discussion we had in the original serialization PR (see https://github.com/holoviz/param/pull/414).

The `serialize` and `deserialize` methods are meant to return a Python intermediate representation (IR) that should be agnostic to the actual serialization format and it's the responsibility of the `Serializer` class to (de)serialize Python types from/to the underlying format. 'null' is specific to the JSON (and YAML) representation of None, so many serializer won't know what to do with it.

I think most `deserialize` implementations are wrong to check for `value == 'null'` and should instead check for `value is None`.

This being said it might be that it's too late and this should have been caught much earlier.

philippjfr:
@Hoxbro Your fixes here look good, the issue I noted above precede you. So to finish this I'd like to ask you to change all `serialize` implementations to return `None` if the value is `None` and for all `deserialize` implementations to handle `None` as input.

philippjfr:
Okay, sorry again. I think you already did that. Sorry for the noise.

PR #658, opened 2022-11-06
Add Python 3.11 in the test matrix

maximlt:
This PR adds Python 3.11 in the test matrix. To get the test to pass on Python 3.11 I had to replace usages of `asyncio.coroutine` (deprecated since Python 3.8) by using `async/await` keywords. Since they're not supported in Python 2 I had to move some of the tests to their own module to avoid them being collected when the test suite runs with Python 2.

The last issue is related to gmpy that cannot get installed on Python 3.11 at the moment. Shouldn't Param instead use [gmpy2](https://pypi.org/project/gmpy2/)?

### DIFF
--- a/tests/API1/testparamdepends.py
+++ b/tests/API1/testparamdepends.py
@@ -677,24 +677,23 @@ class TestParamDepends(API1TestCase):
         self.assertEqual(pinfo.name, 'a')
         self.assertEqual(pinfo.what, 'value')
     
-    @pytest.mark.skipif(sys.version_info.major == 2, reason='asyncio only on Python 3')
-    def test_async(self):
-        try:
-            param.parameterized.async_executor = async_executor
-            class P(param.Parameterized):
-                a = param.Parameter()
-                single_count = param.Integer()
+    # @pytest.mark.skipif(sys.version_info.major == 2, reason='asyncio only on Python 3')
+    # def test_async(self):
+    #     try:
+    #         param.parameterized.async_executor = async_executor
+    #         class P(param.Parameterized):
+    #             a = param.Parameter()
+    #             single_count = param.Integer()
 
-                @param.depends('a', watch=True)
-                @asyncio.coroutine
-                def single_parameter(self):
-                    self.single_count += 1
+    #             @param.depends('a', watch=True)
+    #             async def single_parameter(self):
+    #                 self.single_count += 1
 
-            inst = P()
-            inst.a = 'test'
-            assert inst.single_count == 1
-        finally:
-            param.parameterized.async_executor = None
+    #         inst = P()
+    #         inst.a = 'test'
+    #         assert inst.single_count == 1
+    #     finally:
+    #         param.parameterized.async_executor = None
 
 
 class TestParamDependsFunction(API1TestCase):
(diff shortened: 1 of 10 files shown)


jbednar:
I don't think Param relies on anything introduced in gmpy2, but if both work, feel free to switch it!

maximlt:
In https://github.com/holoviz/param/pull/658/commits/c08ea2640a7e7ca2bb6bb34def1e99e80564e7e7 I tried to use `gmpy2` but on Linux there was a failing test and some warnings:

```
   __________ TestTimeDependentDynamic.test_time_hashing_rationals_gmpy2 __________
  
  self = <tests.API1.testtimedependent.TestTimeDependentDynamic testMethod=test_time_hashing_rationals_gmpy2>
  
      @pytest.mark.skipif(gmpy2 is None, reason="gmpy2 is not installed")
      def test_time_hashing_rationals_gmpy2(self):
          """
          Check that hashes of fractions and gmpy2 mpqs match for some
          reasonable rational numbers.
          """
          pi = "3.141592"
          hashfn = numbergen.Hash("test", input_count=1)
          self.assertEqual(hashfn(0.5), hashfn(gmpy2.mpq(0.5)))
  >       self.assertEqual(hashfn(pi), hashfn(gmpy2.mpq(3.141592)))
  E       AssertionError: 238769760 != 243050811
  
  tests/API1/testtimedependent.py:317: AssertionError
  ----------------------------- Captured stderr call -----------------------------
  WARNING:param.main: Casting type 'float' to Fraction.fraction
  WARNING:param.main: Casting type 'mpq' to Fraction.fraction
  WARNING:param.main: Casting type 'str' to Fraction.fraction
  WARNING:param.main: Casting type 'mpq' to Fraction.fraction
  ------------------------------ Captured log call -------------------------------
  WARNING  param.main:parameterized.py:2529 Casting type 'float' to Fraction.fraction
  WARNING  param.main:parameterized.py:2529 Casting type 'mpq' to Fraction.fraction
  WARNING  param.main:parameterized.py:2529 Casting type 'str' to Fraction.fraction
  WARNING  param.main:parameterized.py:2529 Casting type 'mpq' to Fraction.fraction
```

Also similarly to `gmpy`, it wouldn't install on Linux and fail with:

```
        In file included from src/gmpy2.c:500:
        src/gmpy2.h:82:10: fatal error: mpfr.h: No such file or directory
           82 | #include <mpfr.h>
              |          ^~~~~~~~
        compilation terminated.
        error: command '/usr/bin/gcc' failed with exit code 1
        [end of output]
```

Last update of `gmpy` on PyPI was in July 2013, it was in January 2022 for `gmpy2`. I'll open an issue to suggest supporting `gmpy2` instead of `gmpy`.

tests/API1/testparamdepends.py
```diff
@@ -677,24 +677,23 @@ def test_param_external_param_instance(self):
...
-                @asyncio.coroutine
-                def single_parameter(self):
-                    self.single_count += 1
-
-            inst = P()
-            inst.a = 'test'
-            assert inst.single_count == 1
-        finally:
-            param.parameterized.async_executor = None
+    # @pytest.mark.skipif(sys.version_info.major == 2, reason='asyncio only on Python 3')
```
jlstevens, line 680: @maximlt I think it might just be worth adding a comment to say this test can uncommented when python 2 is dropped.

PR #672, opened 2022-12-29
master to main

maximlt:


maximlt:
@jlstevens `param.version` has a reference to *master*. Does it need to be updated to also try *main*?

https://github.com/holoviz/param/blob/897687f71ac4ca7da6d35361ffe832f93a64a128/param/version.py#L403

jlstevens:
> @jlstevens `param.version` has a reference to _master_. Does it need to be updated to also try _main_?
> 
> https://github.com/holoviz/param/blob/897687f71ac4ca7da6d35361ffe832f93a64a128/param/version.py#L403

Yes. I think it should be sufficient to just add `'main'` to that list.

PR #674, opened 2023-01-03
create initial project governance docs

droumis:
The files in this pull request should act as examples of how other HoloViz projects could quickly adopt the default project docs using references to the project docs at holoviz/holoviz.

The maintainer list is based on the HoloViz Duties spreadsheet. All three maintainers should participate and/or approve.

I'll mention this at the next docs meeting, but I've written the [guidance to create project docs](https://docs.google.com/document/d/1DSPq7oyR5VyV20g7Ez52tVAkwm12ilfMKZtZumFzAaI/edit?usp=sharing). I'll update this if there are any further changes to this PR.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,5 +1,5 @@
 # Code of Conduct
 
-For the code of conduct, see [HoloViz/HoloViz - CODE_OF_CONDUCT.md](https://github.com/holoviz/holoviz/blob/8257fe24af9d18598cee9f55055b78301307bd65/CODE_OF_CONDUCT.md).
+For the code of conduct, see [HoloViz/HoloViz - CODE_OF_CONDUCT.md](https://github.com/holoviz/holoviz/blob/param-gov/CODE_OF_CONDUCT.md).
 
 The Param Project’s equivalently named documents take precedence over any external materials referenced within this linked document above.
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,3 @@
+# Code of Conduct
+
+This project adopts the [HoloViz.org code of conduct](https://github.com/holoviz/holoviz/blob/8257fe24af9d18598cee9f55055b78301307bd65/CODE_OF_CONDUCT.md).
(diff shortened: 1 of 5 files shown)


jbednar (approved):
Looks good. Should we maybe have a tag like `param-governance` on the holoviz repo, so that these documents could point to that tag as an unambiguous version, but then we could move that tag whenever we wanted to inherit the new changes from holoviz?

jlstevens:
Looks good to me!

> Should we maybe have a tag like param-governance on the holoviz repo, so that these documents could point to that tag as an unambiguous version, but then we could move that tag whenever we wanted to inherit the new changes from holoviz?

A tag is a good idea though if we want to point to a different version for different projects then yes, we will need a `<project>-governance` tag for every project. Still probably better than pointing to an obscure SHA.

droumis:
> Should we maybe have a tag like param-governance on the holoviz repo, so that these documents could point to that tag as an unambiguous version, but then we could move that tag whenever we wanted to inherit the new changes from holoviz?

I've never done something like this, but I assume that in order to use a consistent `param-gov`-tagged link like `https://github.com/holoviz/holoviz/blob/param-gov/file.md` (not a real link) which starts on commit 'A',  then when it is time to update the `param-gov` tag to a more recently committed version of this document, commit 'B', then someone with merge permissions for holoviz/holoviz repo would have to first delete the tag from the commit 'A' and then push the tag to 'commit B'.

I'm not a maintainer for param so don't consider this a blocker, but I'm not sure I agree with this approach since the history of such a reference update would not live within the Param repo. Alternatively, an update to the SHA links would just be a matter of updating 4 such links and the history would be a pull request within the Param repo (and I imagine that this would happen pretty rarely)

jbednar:
Moving a git tag should be straightforward for any holoviz maintainer.

We're already depending on an external repo, because the explicit sha being referenced currently has no meaning outside of the external repo. It's true that it's not difficult to replace the sha in four files, but it _is_ difficult to know how many files, and which ones, need to be updated. Each time we need to do that there will be an investigation where we find an original PR like this or grep the repo for that sha or something of the like, all to make a trivial change. 

I do agree that it's helpful to have a history of the governance versions, and moving a tag doesn't immediately provide such a history, which is a downside. But as we indeed don't expect many changes to the governance, and generally look forward rather than backward, I'd favor having it be straightforward and obvious how to update to the latest version from holoviz, rather than optimizing for preserving past links. Other maintainers, please weigh in if you disagree or have a better suggestion, and not just for Param but for other HoloViz projects, as it would be best to handle them all similarly by default.

philippjfr (approved):
Looks good!

droumis:
[`param-gov`](https://github.com/holoviz/holoviz/releases/tag/param-gov) tag has been created and updated the links in this PR and the instructions on the google doc linked in the description at the top. I'll merge now

PR #636, opened 2022-06-07
Add utilities for auto-generating param Python types

philippjfr:
This is an attempt at implementing a solution for https://github.com/holoviz/param/issues/376

Major, major HACKALERT for now as there a large number of unsafe/incomplete details here:

- [ ] Implementations of `pytype` properties not complete
- [ ] Mapping of `Parameterized` to `ast.ClassDef` node not entirely safe
- [ ] Mapping of `Parameter` to `ast.Assign`/`ast.AnnAssign` node not entirely safe
- [ ] Formatting of types returned by `pytype` property into valid type definition string not safe
- [ ] Insertion of type definition assumes:
    - Default spacing `pname =`
    - No existing type definitions, i.e. `pname: int =` will not work
- [ ] Missing handling of imports for inserted type definitions
    - Assumes `typing` module has been imported
    - Any custom types are not imported

## Example

Here is an example of the types generated for `panel.viewable.Layoutable` (`doc` definitions omitted for readability):

```python
class Layoutable(param.Parameterized):
    """
    Layoutable defines shared style and layout related parameters
    for all Panel components with a visual representation.
    """

    align: typing.Any = param.ClassSelector(default='start',
                                class_=(str, tuple))

    aspect_ratio: typing.Any = param.Parameter(default=None)

    background: typing.Any = param.Parameter(default=None)

    css_classes: typing.Any = param.List(default=None)

    width: typing.Optional[numbers.Number] = param.Integer(default=None, bounds=(0, None))

    height: typing.Optional[numbers.Number] = param.Integer(default=None, bounds=(0, None))

    min_width: typing.Optional[numbers.Number] = param.Integer(default=None, bounds=(0, None))

    min_height: typing.Optional[numbers.Number] = param.Integer(default=None, bounds=(0, None))

    max_width: typing.Optional[numbers.Number] = param.Integer(default=None, bounds=(0, None))

    max_height: typing.Optional[numbers.Number] = param.Integer(default=None, bounds=(0, None))

    margin: typing.Any = param.Parameter(default=5)

    width_policy: typing.Any = param.ObjectSelector(
        default="auto", objects=['auto', 'fixed', 'fit', 'min', 'max'])

    height_policy: typing.Any = param.ObjectSelector(
        default="auto", objects=['auto', 'fixed', 'fit', 'min', 'max'])

    sizing_mode: typing.Any = param.ObjectSelector(default=None, objects=[
        'fixed', 'stretch_width', 'stretch_height', 'stretch_both',
[...]

PR #677, opened 2023-01-09
Add pytypes property to parameters to return Python type declaration

philippjfr:
Decided to split out the less experimental part of https://github.com/holoviz/param/pull/636. This PR adds a `pytypes` property to all Parameters that returns a Python type declaration for a specific parameter. This can be used in a number of ways, e.g. for documentation but also eventually for a tool that can automatically generate types for Parameterized classes (which I had prototyped in #636).


maximlt:
@philippjfr do you need this in 2.0?

jbednar:
I'll be glad to have this mapping between Parameter types and py3 types, but I can't comment on whether this is the best way to achieve that. Up to @philippjfr .

philippjfr:
I don't think it's ready, I want to revisit this as part of a more general plan for composite types as well as dataclass and pydantic integrations.

PR #680, opened 2023-01-16
Fix CI failures

Hoxbro:
Attempting to see why CI fails... hard.


philippjfr:
Do we know what about setuptool 66 broke things?

Hoxbro:
I just succeeded in recreating this on my own computer with docker. 

I'm still investigating, but if you want to merge this into the main branch I will open a new PR with a more permanent fix.

PR #683, opened 2023-01-17
Fetch tags before running tests

Hoxbro:
The test suite did not fetch the tags before the test and would therefore have a version of `"None"`.

With the latest release of `setuptools`, `"None"` was no longer a valid version tag, which was why the tests were failing.


philippjfr:
Ah, that makes sense, thanks!

Issue #689, opened 2023-03-06
`allow_None` not respected for `Parameter`


maximlt:
Despite `allow_None` being set to `False`, either automatically when a default value different than `None` is provided or explicitly, it's possible setting the parameter's value to `None` without any error being raised.

```python
import param

class Foo(param.Parameterized):
    p1 = param.Parameter('test', allow_None=False)
    p2 = param.Parameter('test')

f = Foo()

print(f.param.p1.allow_None, f.param.p2.allow_None)

f.p1 = None
f.p2 = None

print(f.p1, f.p2)
```

<img width="458" alt="image" src="https://user-images.githubusercontent.com/35924738/223045750-aa09025b-6d8e-4c04-aaeb-edc69e944e58.png">

jlstevens:
I would say this is indeed a bug: while `Parameter` lets you specify anything as a value, if you explicitly say it should not be `None`, then that should be respected.

jbednar:
Funny. Such a Parameter declaration literally  allows every possible value _except_ None, so it's not likely ever to be useful. But yes, it certainly shouldn't accept None in that case!

maximlt:
This was implemented in https://github.com/holoviz/param/pull/691 and reverted in https://github.com/holoviz/param/pull/712 as it broke Panel. I'm just going to close this as, as said above:

> it's not likely ever to be useful

PR #691, opened 2023-03-06
Raise an error when setting `None` to a base Parameter with `allow_None=False`

maximlt:
Fixes https://github.com/holoviz/param/issues/689


philippjfr (approved):
Looks good!

jlstevens:
Yup, looks good to me too!

PR #710, opened 2023-03-12
Add more tests

maximlt:
This PR doesn't touch any file in `param` or `numbergen`, it only adds more tests. Adding tests was motivated by the Sentinel PR, that in its current implementation is dynamically computing the default value of Parameter attributes (e.g. `instantiate`), opposed to know where most of them are just statically declared in the signature definitions. The tests added captures the current default values of most of the Parameters. This PR also adds some rather simple tests to previously untested Parameters such as `Path`.

### DIFF
--- a/tests/API1/testbooleanparam.py
+++ b/tests/API1/testbooleanparam.py
@@ -1,0 +1,155 @@
+"""
+Unit test for Boolean parameters.
+"""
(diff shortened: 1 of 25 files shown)


maximlt:
Going to self-merge as this just includes changes to the test files.

PR #716, opened 2023-03-14
Add more tests

maximlt:
- No change to the code source
- Add more tests:
  - in particular on inheriting metadata in a hierarchy of Parameterized objects
  - but also capturing some current behavior (that could be considered as buggy) that may change with the sentinel PR
- Soft removal of Python 2.7 and 3.6. A complete and cleaner drop of these two versions would require more work, this is just to be able to use newer features of pytest.


examples/user_guide/Outputs.ipynb
```diff
@@ -48,6 +48,7 @@
    ]
   },
   {
+   "attachments": {},
```
maximlt, line 51:
```suggestion
```

Issue #700, opened 2023-03-07
`Boolean` does no validate its default value


maximlt:
The `Boolean` Parameter does not validate its default value.

```python
import param

class P(param.Parameterized):
    b = param.Boolean(default='foo')

p = P()
print(p.b)
# 'foo'
```

maximlt:
Fixed in https://github.com/holoviz/param/pull/722

PR #722, opened 2023-03-23
Validate type of param.Boolean default value

minimav:
I've had a go at adding validation for the default for `param.Boolean` (see https://github.com/holoviz/param/issues/700). I was not sure whether the validation should occur before or after the `super` call, let me know if after that make more sense.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -998,7 +998,14 @@ class Boolean(Parameter):
     # no effect since values are either False, True, or None (if allowed).
     def __init__(self, default=False, bounds=(0,1), **params):
         self.bounds = bounds
+        self._validate_default(default)
         super(Boolean, self).__init__(default=default, **params)
+
+    def _validate_default(self, default):
+        if not isinstance(default, bool) and default is not None:
+            raise ValueError("Boolean parameter only takes a "
+                             "Boolean default or None, not %s."
+                             % default)
 
     def _validate_value(self, val, allow_None):
         if allow_None:
--- a/tests/API1/testbooleanparam.py
+++ b/tests/API1/testbooleanparam.py
@@ -79,6 +79,12 @@ class TestBooleanParameters(API1TestCase):
         with self.assertRaisesRegex(ValueError, msg):
             p.e = 'test'
 
+    def test_bad_default_type(self):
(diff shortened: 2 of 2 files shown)


param/__init__.py
```diff
@@ -998,8 +998,15 @@ class Boolean(Parameter):
...
         self.bounds = bounds
+        self._validate_default(default)
         super(Boolean, self).__init__(default=default, **params)
 
+    def _validate_default(self, default):
+        if not isinstance(default, bool) and default is not None:
+            raise ValueError("Boolean parameter only takes a "
+                             "Boolean default or None, not %s."
+                             % default)
+
```
jbednar, line 1009: That logic looks good, thanks! I think we should follow the example of other Parameters like String and just call _validate in the constructor:

```
        self._validate(default)
```

minimav: Thanks for the feedback :) I had a go at following what is done in Tuple in the latest commit. The only issue was having to modify the error message since `self.name` is not set when the validation happens and the message you would get, `"Boolean parameter None must be True or False, not {val}."`, seems confusing to me.

param/__init__.py
```diff
@@ -1007,8 +1008,15 @@ def _validate_value(self, val, allow_None):
...
         elif not isinstance(val, bool):
-            raise ValueError("Boolean parameter %r must be True or False, "
-                             "not %s." % (self.name, val))
+            if self.name is not None:
+                raise ValueError("Boolean parameter %r must be True or False, "
+                                 "not %s." % (self.name, val))
+            else:
+                raise ValueError("Boolean parameter must be True or False, "
+                                 "not %s." % val)
+
```
jbednar, line 1017:
```suggestion
            name = " " + self.name if self.name else ""
            raise ValueError("Boolean parameter%r must be True or False,  not %s." % (name, val))
```

jbednar (changes requested):
Looks good! Minor simplification.

param/__init__.py
minimav: 👍 Nice. There was an extra space in the suggestion so I did in locally, had to deal with %r on `" " + self.name` being slightly off as well.

jbednar:
Thanks!

Issue #647, opened 2022-08-06
Add Developer Guide to the web site


MarcSkovMadsen:
## Request

Add *Developer Guide* to Param web site similar to Panel and hvPlot

## Motivation

I would like to contribute to Param. The first thing I need to know is how to install for development. I cannot find a guide anywhere.

Thus I risk waisting time because I might not be able to install for development, install it in a way that causes issues or just cannot properly test the changes I make before creating the PRs.

Similarly a Developer Guide would be a sign that the project is welcoming contributions and is prepared to help new contributors.

MarcSkovMadsen:
Hi. Potential contributor here :-) Cannot find the Developer Guide?

PR #724, opened 2023-03-26
Adds support for required/ mandatory parameter values

MarcSkovMadsen:
Addresses #1

The first commit is a Proof of concept to spark the design discussion. I've added a `required` argument as suggested by @philippjfr. I can see that @jbednar also has an interested in this issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ coverage.xml
 
 # autover
 param/.version
+.venv/
+script.*
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -970,7 +970,7 @@ class Parameter(object):
     __slots__ = ['name', '_internal_name', 'default', 'doc',
                  'precedence', 'instantiate', 'constant', 'readonly',
                  'pickle_default_value', 'allow_None', 'per_instance',
-                 'watchers', 'owner', '_label']
+                 'watchers', 'owner', '_label', "required"]
 
     # Note: When initially created, a Parameter does not know which
     # Parameterized class owns it, nor does it know its names
@@ -983,7 +983,7 @@ class Parameter(object):
     def __init__(self,default=None, doc=None, label=None, precedence=None,  # pylint: disable-msg=R0913
                  instantiate=False, constant=False, readonly=False,
                  pickle_default_value=True, allow_None=False,
-                 per_instance=True):
+                 per_instance=True, required=False):
 
         """Initialize a new Parameter object and store the supplied attributes:
 
@@ -1051,6 +1051,8 @@ class Parameter(object):
         allowed. If the default value is defined as None, allow_None
         is set to True automatically.
 
+        required: If True a value must be provided on instantiation.
+
         default, doc, and precedence all default to None, which allows
         inheritance of Parameter slots (attributes) from the owning-class'
         class hierarchy (see ParameterizedMetaclass).
@@ -1070,6 +1072,7 @@ class Parameter(object):
         self.allow_None = (default is None or allow_None)
         self.watchers = {}
         self.per_instance = per_instance
+        self.required = required
 
     @classmethod
     def serialize(cls, value):
@@ -1601,6 +1604,16 @@ class Parameters(object):
         self = self_.param.self
         self.param._set_name('%s%05d' % (self.__class__.__name__ ,object_count))
 
+    @as_uninitialized
+    def _validate_params(self_, **params):
+        self = self_.param.self
+        params_to_instantiate = self.__class__.param._parameters
+                
+        for p,v in params_to_instantiate.items():
+            if v.required and not v.name in params:
+                message = f"__init__() missing 1 required positional argument: '{v.name}'"
+                raise TypeError(message)
+                
 
     @as_uninitialized
     def _setup_params(self_,**params):
@@ -3170,6 +3183,7 @@ class Parameterized(object):
         self._dynamic_watchers = defaultdict(list)
 
         self.param._generate_name()
+        self.param._validate_params(**params)
         self.param._setup_params(**params)
         object_count += 1
 
--- a/tests/API1/testparameterizedobject.py
+++ b/tests/API1/testparameterizedobject.py
@@ -813,3 +813,31 @@ def test_inheritance_class_attribute_behavior():
     # Should be 2?
     # https://github.com/holoviz/param/issues/718
     assert B.p == 1
+
+@pytest.mark.parametrize(["default", "value"], [
+    (None,"v"), ("d", "v")
+])
+def test_constant_parameter_not_raises(default, value):
+    """Test that you can make a parameter required and it will not raise
+    an error if provided."""
+    class TestConstant(param.Parameterized):
+        value = param.Parameter(default=default, required=True)
+
+    po = TestConstant(value=value)
+    assert po.value==value
+
+import re
+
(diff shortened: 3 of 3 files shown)


MarcSkovMadsen:
As I don't know the inner workings of the `param` code base and there is [no developer guide](https://github.com/holoviz/param/issues/647), I don't know what is required from the code or tests. I would need some strong guidance to proceed. Thanks.

MarcSkovMadsen:
An alternative to the required argument could be a options argument.

Something that talks in favorit of this is that options is the word used normally in python

MarcSkovMadsen:
There is a Long discussion about required for typing here https://mail.python.org/archives/list/typing-sig@python.org/thread/PYHMQEL7KKV7CJOPLY5BR6PTR4O5K7VU/

jbednar:
Thanks. Let's hold off on this until #605 is merged, which should make this go more smoothly.

maximlt:
Considering the API only of this feature, there are two options available with a) passing a sentinel value to `default` (`Parameter(default=REQUIRED)`) and b) adding a new `required` slot (`Parameter(required=True)`), b) being the approach chosen in this PR.

Overall I think I prefer b), the API looks cleaner to me eyes, it's easy to document, and I believe the implementation should be more straightforward (not having to fiddle with another sentinel value now #605 is merged and has added `Undefined`), the implementation suggested in this PR seems almost ready.
On the other hand, adding a new global slot does have some performance implications, using more memory.

@jlstevens what's your opinion?

maximlt:
I have updated this PR, most importantly to make sure all the class Parameters are evaluated. I have also added some more tests and updated the existing ones.

The behavior allowed by `required` is very similar to what Python allows with `*` added to a function signature before a list of required keyword arguments:
image

The error raised by Param is similar to the one emitted by Python. The only difference being that it includes the class name (`P.__init__()`, note the `P`).

image

jlstevens:
I'll review this PR in more detail tomorrow...

That said, my initial reaction is that while I don't object to `required` being an optional flag to parameters, I don't think I would ever use it myself: positional arguments seem to me to be the idiomatic way to ensure arguments are provided. Generally, I would lean towards enforcing required parameters that way at the Python level instead of relying on param for this sort of thing...

maximlt:
I see your point. I think the point made by Marc on [Discourse](https://discourse.holoviz.org/t/how-to-make-parameter-values-mandatory/5124) is that he wanted to avoid having to write boiler plate code.

image

What this PR doesn't address is what it means to pass the default value, e.g.
```python
import param

class P(param.Parameterized):
    x = param.Parameter(required=True)  # default is None

p = P(x=None)  # should this raise an error?
```

Maybe you have an opinion on this @MarcSkovMadsen ?

maximlt:
@jlstevens a *potential* advantage of having `required` as a new slot is that it could be leveraged by GUIs to label a widget as "mandatory" (with that red asterisk, you know)? Although I'm not sure the semantics match, as in a form you usually validate that the mandatory fields have been filled at some submit time, while in Param's case it has an effect earlier at the instantiation time. Now I'm not actually sure the two concepts can be combined in a single option! Might be worth considering that when deciding the fate of this PR.

philippjfr:
Good point, that is definitely a very common use case for this functionality. It really requires a secondary piece of information, since it implies some action you need to know **what** the parameter is required for. The most common case is a form submission. For that case I could see an `Event` or an `Action` parameter having a list of required parameters associated with it.

MarcSkovMadsen:
> I see your point. I think the point made by Marc on [Discourse](https://discourse.holoviz.org/t/how-to-make-parameter-values-mandatory/5124) is that he wanted to avoid having to write boiler plate code.
> 
> 
> 
> image
> 
> 
> 
> What this PR doesn't address is what it means to pass the default value, e.g.
> 
> ```python
> 
> import param
> 
> 
> 
> class P(param.Parameterized):
> 
>     x = param.Parameter(required=True)  # default is None
> 
> 
> 
> p = P(x=None)  # should this raise an error?
> 
> ```
> 
> 
> 
> Maybe you have an opinion on this @MarcSkovMadsen ?

I can live with having to write a custom __init__ function. I believe it would improve Param if this was documented as the recommended approach (for now).

I think that long term/ theoretically Params Parameters should have an api similar to Dataclasses/ Pydantic such that eventually static type checking can be supported. If that enables supporting required parameters in forms it would be perfect.

MarcSkovMadsen:
Just for reference there is a request here for Panel to support an api for forms here https://github.com/holoviz/panel/issues/3687.

maximlt:
> I can live with having to write a custom **init** function for now. I believe it would improve Param if this was documented as the recommended approach (for now).

I'm not sure I understand. Are you saying that you actually don't need the feature you started to implement in this PR?

> I think that long term/ theoretically Params Parameters should have an api similar to Typing/ Dataclasses/ Pydantic such that eventually static type checking can be supported. If that enables supporting required parameters in forms it would be perfect.

Indeed, but that is out of scope of this PR and discussion I believe.

To rephrase what I was asking you, do you believe that in the example below instantiating `P` with `x` set to `None`, which is the default value of that Parameter, should raise an error? I'm asking you in particular as in your example on Discourse you're mentioning that you have a class for which the default Parameters don't make sense.

```python
import param

class P(param.Parameterized):
    x = param.Parameter(required=True)  # default is None

p = P(x=None)  # should this raise an error?
```

MarcSkovMadsen:
The example should not raise an error because of required=True as you provide a parameter.

But if allow_None=False it should of course raise an exception.

maximlt:
> The example should not raise an error because of required=True as you provide a parameter.
> 
> But if allow_None=False it should of course raise an exception.

Ok thanks for your feedback, this is the behavior implemented in this PR.

jbednar:
Thanks, @Marc for the implementation and @Maxime for the adjustments to it. I'm happy for this PR to be merged as-is; looks good to me!

For completeness, though, I think there is a third way to enforce required parameters not considered above, beyond (1) specifying a `param.Required` sentinel value for the default or (2) supporting `required=True` as in this PR. Specifically, now that `Undefined` lets us determine whether an argument has been passed in or not, we could consider (3) using `allow_None` to cover this case. Regardless of this PR or any thoughts about required parameters, I think an error should be raised if the user provides a default of `None` but _explicitly_ specifies `allow_None=False`. We haven't previously raised such an error, because the default for `allow_None` is False, so we have silently overridden that declaration to match the new information that apparently None _is_ allowed. Now that we can tell the user has _explicitly_ set `allow_None` to False, I think we can reliably detect and signal this combination as an error, regardless of any considerations about required parameters. But then if we do raise an error there, we already have a poor-man's support for required Parameters, i.e. specifying the impossible combination `default=None, allow_None=False` that an instantiator must then override.

I'm not actually arguing for approach (3), because e.g. it wouldn't allow someone to accept `None` as a value for a required Parameter. But I did want to list it for the record.

jlstevens:
In some ways I don't think something like this would be insane:

```python
   foo = param.Number(default=param.RequiredValue(52.4))
```

The main reason I say this is that for all the years I have used param, I think this may be the first time where a new argument has been proposed to apply for all parameters. 

While I am not suggesting that I want to block this PR given that some people may find it useful, I know that (personally) I will continue to enforce mandatory arguments in the class constructor explicitly. I think I find this a lot easier than having to scan through all the parameters for the required flag, especially as parameters can be inherited from super classes imported from different spots in the code.

Issue #698, opened 2023-03-07
Stricter validation for `Range`


maximlt:
The `Range` Parameter could be more strictly validated:

- [x] Verify that that `b >= a` in `(a, b)`
- [x] Verify that `step` is a number

PR #725, opened 2023-03-30
Validate range

minimav:
This implements the validation on the `Range` param suggested in https://github.com/holoviz/param/issues/698. I tried to match what I could see in other params which had such logic already. A couple of minor things:

* `'q's` in the order error message for a named param was a compromise since `%r's` results in `'q''s` which looks a bit awkward.
* The order error message could perhaps be tweaked to `Range bounds...` when validating the bounds rather than `Range parameter ...` for both bounds and the param value.

Fixes https://github.com/holoviz/param/issues/698


param/__init__.py
```diff
@@ -2180,6 +2180,26 @@ def __init__(self, default=Undefined, bounds=Undefined, softbounds=Undefined,
     def _validate(self, val):
         super(Range, self)._validate(val)
         self._validate_bounds(val, self.bounds, self.inclusive_bounds)
+        self._validate_step(val, self.step)
+        self._validate_order(self.bounds, allow_None=True)
```
maximlt, line 2184: If the intention was to validate the order of `bounds`, then I would suggest to code that in `_validate_bounds`. As it is in this PR `Range(bounds=(1,0))` will indeed raise an error but that is going to be a bit confusing as it doesn't say that it's because of `bounds`: `ValueError: Range parameter end 0 is less than its start 1.`.

maximlt (commented):
Thanks @minimav for this well appreciated PR! I have left one inline comment, otherwise it looks good, and I am very pleased that you made the effort to add some tests 👍

minimav:
Thanks @maximlt, I have been a little busy and didn't manage to make the change you suggested.

param/__init__.py
```diff
@@ -2378,6 +2378,25 @@ def __init__(self, default=Undefined, bounds=Undefined, softbounds=Undefined,
...
+                             "numeric value, not type %r." % type(step))
+
+    def _validate_order(self, val, allow_None):
+        if val is None and allow_None:
+            return
+        elif val is not None and (val[0] is None or val[1] is None):
+            return
+
+        start, end = val
+        if not start <= end:
```
maximlt, line 2396: Actually, @jlstevens, I'm having a bit of a doubt here. Should `Range` enforce `a <= b` or is it actually fine to have `a < b`? Like in Python `range(10, 0, -1)` which would "translate" in Param to `param.Range(default=(10, 0), step=-1)`?

The code in the existing `_validate_bounds` seems to assume `a <= b`: https://github.com/holoviz/param/pull/725/files#diff-b4344305e71345c3583ee7c047c524b6b272f31f6af55e2b14c54914b1476cd1R2406-R2408

jbednar: What seems natural is to enforce a<=b if step>0, enforce b=>a if step<0, and enforce b=a if step=0.

maximlt: Yes that makes sense. Even if the case where step=0 is a little weird. Python `range` does not accept a step of 0.

jbednar: Disallowing step=0 altogether also works; as you wish.

jlstevens: I think the semantics Jim suggested are good and I would vote for disabling `step=0` entirely.

minimav: Thanks for the comments 👍 Latest commit should implement what is described above.

jbednar (approved):
Looks great, thanks!!

philippjfr (approved):
Looks good, let's get this merged!

maximlt:
Yep let me have a last look and merge after that.

maximlt:
Thanks a lot for your contribution @minimav ! This fix is going to be included in the release of Param 2.0.

Issue #583, opened 2021-12-15
Array should not force `allow_None=True`


jbednar:
As mentioned by @jlstevens in https://github.com/holoviz/param/pull/582 , it's odd that an Array parameter forces `allow_None` to True:

```python
class Array(ClassSelector):
    """
    Parameter whose value is a numpy array.
    """

    def __init__(self, default=None, **params):
        from numpy import ndarray
        super(Array, self).__init__(ndarray, allow_None=True, default=default, **params)
```

I'm guessing that the original reason for that was to make the code able to work when NumPy is not available, but because we always assume `allow_None` to be True if the default is `None`, here having a `None` default should be sufficient already, without forcing all `Array` instances to allow None. Fixing this will be a breaking change, since any user of Array that does want to allow None will now have to supply that flag explicitly, but it seems the right thing to do.

PR #726, opened 2023-04-05
Don't hard-code allow_None to True in Array

maximlt:
Fixes https://github.com/holoviz/param/issues/583

This is effectively a breaking change. The code below used to work fine, while after this change `p.a = None` will raise an error.

```python
import param, numpy as np

class P(param.Parameterized):
    a = param.Array(np.zeros((2, 2)))

p = P()

p.a = None
```


maximlt:
Thanks for the review! Merging.

Issue #174, opened 2017-07-16
Stop setting class attributes during initialization


ceball:
```
>>> import param
>>> class X(param.Parameterized):
...     y = 1
...     z = param.Parameter(10)
... 
>>> x = X(y=9)
WARNING:root:X00003: Setting non-parameter attribute y=9 using a mechanism intended only for parameters
```

jlstevens:
I would not want to set class attributes this way and I would rather have an error than the current behavior (I want to handle setting class attributes myself!).

ceball:
I thought this was the warning you were complaining about? If not this one, then which one did you mean? If it is this one, but you're getting it in some other way, could you give me an example? I'm a bit confused.

I created this issue not because I think we should allow what I described in my opening comment, but to record whatever decision we make...so that when you complain about param warnings again in the future, I can refer you back to it ;) But that will only work if we're discussing the right situation in which you get the warning!

jlstevens:
The problem is that param is assuming it has anything to do with class attributes. I don't want param to have any say about class attributes that aren't parameters and I certainly don't want it to set them!

I think the warning behavior should be optional and off by default.

jlstevens:
To be clear, I would prefer an exception if I misspell a parameter name along the lines of 'Cannot set parameter 'fooz' as no such parameter exists'.

A fancier implementation could say something like:

"Cannot set parameter 'fooz' as no such parameter exists. Similar parameters: 'foo','foz'"

ceball:
Ok, I think I get you now, but let's check! You want this:

```
>>> import param
>>> class X(param.Parameterized):
...     y = 1
...     z = param.Parameter(10)
... 
>>> x = X(y=9)
Error: X has no parameter 'y' 
>>> X.y = 9
>>> X.y
9
```

Whereas you currently get a warning in both cases (``X(y=9)`` and ``X.y=9``).

Have I described what you want?

jlstevens:
> Have I described what you want?

Yes, I think so! Though @jbednar and @philippjfr may also want to chime in...

philippjfr:
Yes I think that's more sensible.

ceball:
> Though @jbednar and @philippjfr may also want to chime in...

Just because I finally described what you want doesn't mean I agree with it, nor does it imply anything about whether or not I would implement it even if did I agree with it ;)

jbednar:
[...]

PR #729, opened 2023-04-05
Don't set class attributes during initialization and don't warn when setting non-Param class attributes

maximlt:
Fixes https://github.com/holoviz/param/issues/174

- Setting a non-Parameter class attribute no longer raises a warning
- Setting a non-Parameter in the constructor raises a `TypeError` (breaking change)

image

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1636,7 +1636,10 @@ class Parameters(object):
         for name, val in params.items():
             desc = self.__class__.get_param_descriptor(name)[0] # pylint: disable-msg=E1101
             if not desc:
-                self.param.warning("Setting non-parameter attribute %s=%s using a mechanism intended only for parameters", name, val)
+                raise TypeError(
+                    f"{self.__class__.__name__} has no Parameter '{name}'; "
+                    f"cannot set '{name}' in the constructor"
+                )
             # i.e. if not desc it's setting an attribute in __dict__, not a Parameter
             setattr(self, name, val)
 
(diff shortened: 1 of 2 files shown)


maximlt:
Oh wait, need to resolve conflicts...

Hoxbro:
Is `TypeError` the best type of error to raise? Maybe `AttributeError`? I could be wrong, though, or there could be an even better error.

maximlt:
`TypeError` is the error raised by Python when you pass a non expected keyword argument in a function call. That's why I chose it.

image

Do you have suggestions for a better error?

Hoxbro:
Maybe a custom one? That sets it apart from normal Python behavior? `ParametrizedError`

maximlt:
It seems to be this is a general Python behavior, you shouldn't be able to pass kwargs to a callable that don't expect them. I don't really see a reason to raise a different error here. One could actually argue that Param should raise the exact same error message.

image

Hoxbro:
> One could actually argue that Param should raise the exact same error message.

I agree.

maximlt:
The error message now mirrors Python more closely.

Pure Python:
image

Param on this branch:
image

Hoxbro:
I like it. I just ran this PR across the holoviz ecosystem and got two failing tests related to this change:

```
FAILED panel/tests/pane/test_holoviews.py::test_holoviews_with_widgets_not_shown - TypeError: HoloViews.__init__() got an unexpected keyword argument 'show_widgets'
---
FAILED hvplot/tests/testinteractive.py::test_interactive_pandas_frame_bind_operator_out_widgets - TypeError: Select.__init__() got an unexpected keyword argument 'default'
```

This does not necessarily mean these changes are wrong, but some tests need to be updated.

Hoxbro (approved):

jbednar (approved):
I think this is the correct change, thanks!

maximlt:
Great thanks! Merging.

PR #733, opened 2023-04-12
Allow type change for DateRange

Hoxbro:
Fixes https://github.com/holoviz/panel/issues/4566

This adds support for having both `datetime.date` and `datetime.datetime` values in `param.DateRange`.


maximlt:
Thanks for that @Hoxbro! I've made some changes, mostly to add the same behavior to `Date` so that Param offers a consistent behavior for these two Parameters. Let me know if you want to review it?

maximlt:
Thanks for the review!

Issue #702, opened 2023-03-09
`List.class_` is deprecated


maximlt:
`List.class_` is documented as deprecated in the docstring. Opening an issue to record that it should be programmatically deprecated.

https://github.com/holoviz/param/blob/cd1b4ae50c01c14c0eb22721e41a9c8fb1e8f738/param/__init__.py#L1420-L1422

Issue #692, opened 2023-03-06
Deprecating methods to be removed for 2.0?


maximlt:
This PR https://github.com/holoviz/param/pull/559 removed from the documentation few methods on the `.param` namespace in preparation for Param 2.0. It also announced that

> [the methods are] expected to warn in the next 1.x release after 1.12)

We are still on`1.12.x` but working actively towards Param 2.0. If we really intend to remove these methods, we should make a release **ASAP** including these deprecation warnings.

- [ ] `.param.defaults()`: one occurrence in HoloViews
- [x] `.param.deprecate()`: None
- [ ] `.param.params()`: occurrences in Lumen, HoloViews and hvPlot (1 in Panel but `objects()` used if found, which is the right thing to do I guess to cover multiple Param versions)
- [x] `.param.print_param_defaults()`: None
- [x] `param.print_param_values()`: None
- [x] `.param.set_default()`: None

I'll open PRs in the listed packages to update them.

maximlt:
Panel has this piece of code, that uses `param.objects()` if available instead of the deprecated `param.params()`.

```python
if hasattr(_config.param, 'objects'):
    _params = _config.param.objects()
else:
    _params = _config.param.params()
```

However the docstring of `objects` indicates that it takes an `instance` argument that can either be `True` (default), `False` or `'existing'`:

>         If instance=True and called on a Parameterized instance it
>         will create instance parameters for all Parameters defined on
>         the class. To force class parameters to be returned use
>         instance=False. Since classes avoid creating instance
>         parameters unless necessary you may also request only existing
>         instance parameters to be returned by setting
>         instance='existing'.

The release notes for 1.12.0 include erroneous information, I think:

> `.param.params()`: Use `.param.values()` or `.param['param']` instead ([#559](https://github.com/holoviz/param/pull/559))

What should be the recommendation to replace `param.params()`? Is it `param.objects()` or `param.objects(instance='existing')` or `param.objects(instance=False)`?

cc @philippjfr as I believe you implemented instance Parameters.

PR #734, opened 2023-04-12
Add programmatic and docstring deprecations

maximlt:
Our plan for Param 2.0 was to improve its API by renaming and removing some methods/functions. This has already been partially achieved, however some API we deprecated was just deprecated by adding a comment in the release notes (of Param 1.12.0 mostly) and in the code, sometimes in docstrings.

We deem that this was not enough to completely remove that API in Param 2.0, as users tend to rely on programmatic warnings. For instance, in some of the HoloViz libraries there were still usages of the deprecated `.param.params()`.

This PR:
- adds programmatic warnings using the `warnings` module of Python to the deprecated methods/functions/attributes
- adds the `deprecated` Sphinx directive to the relevant docstrings
- updates the test suite to no longer use the deprecated API

Fixes https://github.com/holoviz/param/issues/692
Fixes https://github.com/holoviz/param/issues/702


maximlt:
I'm going to merge this PR that is just adding programmatic deprecation warnings to some API that's already been docs-deprecated for a while.

Issue #459, opened 2021-02-17
allow np.integer in validation of param.Integer


LeonvanKouwen:
param                     1.10.1             pyhd3eb1b0_0

This is a very small issue,  but I bumped into it. 

The validate method of `param.Integer` currently does this:

```
if not isinstance(val,int):
    raise ValueError("Parameter '%s' must be an integer."%self.name)
```

But for example when you get integers from a random number generator, very often they are of some numpy int type, which do not inherit from built-in `int`. However, they are just fine integers for `param.Integer` purposes I would say. 

So maybe the validation can be changed to:

```
if not isinstance(val, (int, np.integer)):
    raise ValueError("Parameter '%s' must be an integer."%self.name)
```

[here](https://stackoverflow.com/questions/37726830/how-to-determine-if-a-number-is-any-type-of-int-core-or-numpy-signed-or-not) is a relevant thread on stack overflow about numpy int checking.

jlstevens:
Thanks for the suggestion! 

The proposed change looks good to me (will some additional logic needed to ensure `np` is an optional dependency).

jbednar:
Right; we can't assume Numpy is installed, as it's not a dependency.  I think we have a couple of options here, both of which , require special logic to avoid a dependency on `np`:
1. Simply allow Numpy ints as suggested here  
2. Coerce any provided Numpy int into a supported int type. 

For option 1, I'd be concerned that downstream code already written for `param.Integer` could break if someone provides a Numpy integer to Parameterized code not expecting it. At a minimum, if that parameter value is ever passed to `isinstance(val,int)` in Parameterized code, it would now fail. 

I see Param as offering a contract to code authors, where Param takes care of ensuring that it only lets through a particular universe of values, and then the code using that parameter need only ever consider the items in that universe, never needing to check or test other Python objects that a user might supply. Here, I don't know how big an issue this might be; I personally have found Numpy ints to behave like regular ints in general, but there will definitely be cases where they don't, and the question is whether the benefit of accepting such values outweighs those potential problems.

Option 2 seems safer to me, i.e. to convert the provided int-compatible value into an int.  That way downstream code sees only a regular int, and users can still pass in Numpy ints as desired.  Such conversions can cause problems for mutable data structures, but I don't think they'd be an issue for an int, right?  So I think I'd favor option 2, but am open to arguments in support of option 1 or some other option 3.

LeonvanKouwen:
[...]

Issue #205, opened 2017-11-01
Better behavior for Integer defaults


jbednar:
```
class P(param.Parameterized):
    val = param.Integer(4.0)
```

Should either coerce the provided value into an integer (which would be safe in this case, but may not be desirable if e.g. provided 4.3), or it should have a sensible error message. Instead it results in:

```
ValueError                                Traceback (most recent call last)
<ipython-input-32-f6658de6e3bd> in <module>()
----> 1 class P(param.Parameterized):
      2     val = param.Integer(4.0)
      3 
      4 

<ipython-input-32-f6658de6e3bd> in P()
      1 class P(param.Parameterized):
----> 2     val = param.Integer(4.0)
      3 
      4 

~/anaconda/envs/ds/lib/python3.6/site-packages/param/__init__.py in __init__(self, default, bounds, softbounds, inclusive_bounds, **params)
    588         self.inclusive_bounds = inclusive_bounds
    589         self._softbounds = softbounds
--> 590         if not callable(default): self._check_value(default)
    591 
    592 

~/anaconda/envs/ds/lib/python3.6/site-packages/param/__init__.py in _check_value(self, val)
    750 
    751         if not isinstance(val,int):
--> 752             raise ValueError("Parameter '%s' must be an integer."%self._attrib_name)
    753 
    754         self._checkBounds(val)

ValueError: Parameter 'None' must be an integer.
```

All things considered, I think that since we are explicitly declaring it to be an Integer, we should try converting whatever we are given into an integer.   In any case, we need to fix the broken attrib name here.

ceball:
> I think that since we are explicitly declaring it to be an Integer, we should try converting whatever we are given into an integer

To me, that feels like making things more complex (in implementation and behavior) for a benefit I can't quite see.

Also, where would you draw the line? You mention `4.3` as not being acceptable, but what about e.g. `"4"`?

> In any case, we need to fix the broken attrib name here.

If you were to just type in `val = param.Integer(4.0)` at a prompt, would you expect param.Integer to know about the name `val`? 

I'm not saying what you're asking for is impossible, just that it's not an accidental oversight...

jbednar:
[...]

PR #735, opened 2023-04-12
Allow `Integer` to accept `numpy.integer` values

maximlt:
Fixes https://github.com/holoviz/param/issues/205
Fixes https://github.com/holoviz/param/issues/459

This PR adds `np.integer` as the types supported by the `Integer` Parameter. Chose not to cast to Python `int` not to set a precedent compared to `Number`, `Date` (and others) that also support Numpy types and don't cast.


jlstevens:
Looks good to me! 

I like that it is handled consistently with out `dt_types` does it.

Issue #730, opened 2023-04-05
Make all (but one?) of the Parameter arguments keywords only


maximlt:
Parameters can currently be constructed with a variable number of positional arguments.

```python
param.List([1, 2], int, int, True, (2, 4))
param.ObjectSelector(1, [1, 2, 3])
```

[PEP 3102](https://peps.python.org/pep-3102/) implemented in Python 3.0 added support for declaring arguments that are keywords-only:

```python
def compare(a, b, *, key=None):
    pass

# key must be passed as a keyword argument
compare(1, 2, key=operator.eq)
```

Param dates from way before this PEP was accepted and does not benefit from it, while I think it should :)

There's however a remaining question, that is about the first argument. @jbednar explained me that the spirit of the current state is that the first argument is the value that makes the most sense for the Parameter at hand. For instance, for `Number` it is the `default` number, or for `Boolean` the `default` boolean value. In most cases, `default` is indeed the first argument and can be passed by position, for example `param.Boolean(True)`. There are a few Parameters that don't have `default` as their first argument:

- `ClassSelector` has `class_`
- `Selector` has `objects`. Note that its subclasses - `ListSelector`, `FileSelector`,  and `MultiFileSelector` - still have `default` as a first argument :/ Same for `ObjectSelector` that is deprecated but still used a lot.
- `Composite` has `attribs`
- `Dynamic` actually has default as its first argument but it accepts it as keyword only

Now there are multiple ways to approach that, all have their pros and cons, and all are more or less a breaking change:

1. all arguments keyword only
2. all arguments keyword only but the first one that is always `default`
3. all arguments keyword only but the first one that is the one that makes sense (mostly `default`, but can also be `objects`, `class_`, ...)

This is a timely question with the upcoming Param 2.0 and the Sentinel work (https://github.com/holoviz/param/pull/605) that touched the signatures.

Will need the opinion of Param users on this one to get a feeling on what the best API could be and what the extent of the breaking change you would be willing to accept: @jlstevens, @philippjfr, @Hoxbro, @MarcSkovMadsen

MarcSkovMadsen:
My vote goes to 2.

I believe this keeps things simple.

Hoxbro:
My vote is also on number 2 for the same reason as @MarcSkovMadsen.

jlstevens:
I would personally be happy with option 1 and that seems like the simplest way to me as well.

[...]

Issue #165, opened 2017-07-01
Performance test suite


ceball:
We used to implicitly test param's performance as part of Topographica. However, we should really add a performance test suite of param's own.

Issue #73, opened 2014-12-16
Add support for unicode (with python 2)?


ceball:

jlstevens:
This is fairly important and quite an annoyance. I promise our HoloViews users that it will be fixed in the next release. See the last entry of the HoloViews FAQ: ioam/holoviews@d2dec608e

I'm sure we will have enough improvements by then to justify an updated release of param. :-)

vascotenner:
I normally do something like:
```python
from __future__ import unicode_literals
import numpy as np
import holoviews as hv
```
But then params does not work anymore.

Related: https://github.com/ioam/holoviews/issues/29

jlstevens:
I agree this is an important issue. Will we have a new param bug-fix release before param 2.0 where we can have this fixed?

jbednar:
Let's discuss param releasing around 14 April; I'm waiting to hear some things about param from Chris.

ablythed:
Closing. Works in Python 3 and Python 2 is no longer a priority.

PR #737, opened 2023-04-12
Deprecate passing positional arguments to Parameters

maximlt:
Following the discussion in https://github.com/holoviz/param/issues/730 I am opening this PR for the other maintainers to check how they feel about this change. This PR deprecates passing positional arguments to all the Parameters. I found this very handy decorator in xarray (that they adapted from scikit-learn) here https://github.com/pydata/xarray/blob/7c7e3838685bb49a38303c9d8a91daf62b4c0ebf/xarray/util/deprecation_helpers.py#L44.

I have also made the changes required in `numbergen` and the whole test suite to avoid the warnings. It'd be interesting to see how much work would be required to update Panel and HoloViews to have keyword arguments only, I suspect it would actually be less than what was required in the test suite.


param/parameterized.py
```diff
@@ -58,6 +60,51 @@
...
+    @functools.wraps(func)
+    def inner(*args, **kwargs):
+        name = func.__qualname__.split('.')[0]
+        n_extra_args = len(args) - len(pos_or_kw_args)
+        if n_extra_args > 0:
+            extra_args = ", ".join(kwonly_args[:n_extra_args])
+
+            warnings.warn(
+                f"Passing '{extra_args}' as positional argument(s) to 'param.{name}' "
+                "was deprecated, please pass them as keyword arguments.",
```
Hoxbro, line 93:
```suggestion
                "is deprecated and will stop working in Param X.X. Please pass them as keyword arguments.",
```

param/parameterized.py
```diff
@@ -58,6 +60,51 @@
...
+                DeprecationWarning,
+                stacklevel=2,
```
Hoxbro, line 95: I think it would be beneficial to have a custom warning here. This will make it possible for our other project to raise the warning as an error in our test suite. 

And have you checked if the `stacklevel` is correct?

Hoxbro (commented):
I think we should create a new file `_warnings.py` or similar with contains the `_deprecate_positional_args` function.

maximlt:
> I think we should create a new file _warnings.py or similar with contains the _deprecate_positional_args function.

Yep there's an issue on refactoring Param generally, which I would prefer to do at once in another PR. For now I just make sure any new symbol that isn't meant to be public starts with an `_`.



> Have you tested whether `_deprecate_positional_args` will have any significant cost concerning execution time?

I just did a very simple check, timing running the test suite on *main* and on this branch, I couldn't see any significant difference. That is a pretty poor benchmark though. I never looked into how to set up a performance test suite. There's of course an issue for that :) https://github.com/holoviz/param/issues/165.

param/parameterized.py
maximlt: Yes not mentioning the exact Param release is intentional. I think Param 2.0 took a long time in the making because some changes were announced to happen specifically in Param 2.0 and it took a long time to happen. Oldest mention I can find on Github is in https://github.com/holoviz/param/issues/73#issuecomment-90621960 and it is 8 years old! cc @jlstevens and it's from him and it's kind of funny :)

In practice what I envision is that this would emit a `DeprecationWarning` and about 3 to 6 months before the change (in Param 3.0 I guess) the warning emitted would be a `FutureWarning` so that all users get it.

param/parameterized.py
maximlt: Yep `stacklevel` is correct, I ran the test suite and got correct warnings (lots of them!).

> I think it would be beneficial to have a custom warning here. This will make it possible for our other project to raise the warning as an error in our test suite.

Did you already figure out how to run a test suite and make it fail on some specific warnings only?

Hoxbro: > Yep stacklevel is correct, I ran the test suite and got correct warnings (lots of them!).

Sounds good, and this is also the case for packages outside Param? 

> Did you already figure out how to run a test suite and make it fail on some specific warnings only?

It is the same syntax as ignoring a file in `filterwarnings`. See [here](https://github.com/holoviz/holoviews/blob/main/pyproject.toml). But I can see that there is the possibility to reference the "path" with something like this `error::DeprecationWarning:param.parameterized`. 

Though, this will raise errors for all `DeprecationWarning`s. By having custom warnings, we will have finer control.

param/parameterized.py
Hoxbro: Most people would not know that 2.0 took a long time and will just install the package. 

And my feeling is that most people are lazy, and without a real deadline, there is no urgency and will therefore postpone it.

jlstevens: > Most people would not know that 2.0 took a long time and will just install the package.

I agree.

Simon's suggestion also seems sensible: giving the relevant param version in the warning.

> Oldest mention I can find on Github is in #73 and it is 8 years old!

But is 8 years enough? :-)

maximlt:
> In a big refactor PR, a lot of code will be moved around, making it hard to judge that if something like `_warnings.py` is the best filename. This PR is adding this as a new feature, so I don't see any problem with putting it in the "correct" location to start with.

Really I'd prefer to keep that for a later refactor. That will probably end up in a `_utils` module.

param/parameterized.py
maximlt: Here are a couple counter-arguments.

I have been bitten enough by having to support Python 2 (syntax, test suite, etc.) for such a long time, because it was somehow announced, that I would now be very careful with announcing anything concrete. To me, announcing something is deprecated and programmatically is the most important piece of information. Having the version in the warning message also sets a precedent, making it more likely to have to add it to other warnings, reinforcing the probability for us to tie our hands more than we would like to, while I believe not providing much value to Param users.

Yes, people are lazy, and they usually know it's a bad thing! I would **never** complain to a maintainer for a breaking change, for which a warning was emitted for a long period of time, I'll just blame myself for being too lazy. All the more so when that breaking change happens in a major release, which we would do here.

I would not be so picky if there was not history, but I bet that if asked 8 years ago, none of the Param maintainers would have guessed that it would have taken so long for the release to happen (it's not over by the way, who knows how long it is going to take for it to really happen!).

param/parameterized.py
maximlt: > Sounds good, and this is also the case for packages outside Param?

Tested it in the context of other packages and it looked fine.



> It is the same syntax as ignoring a file in `filterwarnings`. See [here](https://github.com/holoviz/holoviews/blob/main/pyproject.toml). But I can see that there is the possibility to reference the "path" with something like this `error::DeprecationWarning:param.parameterized`.

Great, I've added a `ParamDeprecationWarning` class, being a subclass of `DeprecationWarning`. The test suite fails now if this warning is emitted. I noticed that unfortunately the *examples* tests (I have just added that in this PR) don't fail when this warning is emitted, I presume this has to do with `nbval` :/

maximlt:
I had forgotten to update `String`, which I have fixed now. And I have also updated the docs. This is ready for another round of review.

maximlt:
Superseded by https://github.com/holoviz/param/pull/751

Issue #201, opened 2017-10-26
Enable warnings as errors for param's test suite


ceball:
Once the current warnings have first been displayed (#86), then fixed, we should make warnings be errors.

Set param.parameterized.warnings_as_exceptions=True before nose runs, maybe.

tonyfast:
Let us decide if this is something we really want to do. There are a lot of warnings. This one could hurt folks.

ceball:
To be clear, I'm suggesting to enable param's own "warnings as errors" for one of param's test environments. This does not even need to be the default mode for running the test suite, just so long as at least one test environment on travis does it. I.e. this would have no impact on anyone else (external user, packager, or even tester of param).

In the same PR, would need to fix any errors being raised (i.e. cases where param's own test suite is generating warnings from param), e.g.

![Screenshot from 2020-09-13 15-34-43](https://user-images.githubusercontent.com/1929/93020734-b19ec400-f5d6-11ea-826d-183cae71bacf.png)

This env would only be running the API1 tests, I assume.

If possible/practical, I would also like to go further and catch other kinds of warnings, too, e.g. deprecation and future warnings. Seems like that might become possible after #423 - adding something like `-W error::DeprecationWarning -W error::FutureWarning` to the pytest call. Again, this would not be the default mode of running the test suite, just one test environment on travis.

jbednar:
Sounds good to me. I assume it would be API2 tests, though; not much point in testing API1 anymore.

ceball:
The tests call them API0 and API1 (I guess it was a comment on amount of thought/actual APIness 😅)

jlstevens:
I was wondering about this today when I saw:

```
nosetests testwatch.py
......................WARNING:param.SimpleWatchExample00023: No such watcher Watcher(inst=SimpleWatchExample(a=0, b=0, c=0, d=0, e=False, f=False, name='SimpleWatchExample00023'), cls=<class 'tests.API1.testwatch.SimpleWatchExample'>, fn=<function TestWatch.test_warning_unwatching_when_unwatched.<locals>.accumulator at 0x104b1ec80>, mode='args', onlychanged=True, parameter_names=('a',), what='value', queued=False) to remove.
...................
----------------------------------------------------------------------
Ran 41 tests in 0.016s

OK
```

I can't tell at a glance if that warning is a mistake in the test definitions or expected to happen (if the latter, there should be a way to suppress it). I support warnings as errors to avoid this kind of confusion!

PR #738, opened 2023-04-12
Consider warnings as exceptions in the test suite

maximlt:
Fixes https://github.com/holoviz/param/issues/201


maximlt:
@Hoxbro I'm interested in your opinion on this one! I really wasn't sure about creating specific warning types and spent some time thinking about that. I saw that indeed some libraries implement custom warnings that are usually just subclasses of the builtin warnings. Notably Bokeh, which is maybe where you saw it in the first place? I saw two advantages of having these specific Param warnings:

- It allows other libraries to more easily filter these warnings, as you've started to do in GeoViews. I haven't (yet?) put any of these warnings in the `__init__.py` module as my goal is to clean that namespace and not add to it. So right now you'd turn these warnings into errors with this filter: `error::param._utils.ParamWarning`. Do you think that `ParamWarning` should really be exposed in `__init__.py`? Ideally our test suite would turn all warnings into errors so we wouldn't need to filter these warnings, so my idea is that using the ugly filter above is fine as it's meant to be temporary. Oh and yes I've created a base warning `ParamWarning` so that other libraries can just set up one filter.
- It improves the warning messages by giving them some more context as the warning name precedes the warning message (ush this temporary file path, yuk):
image

The disadvantage is that it is more code to maintain of course , which is why I'm interested in keeping it as private as possible, if that's sensible.

Hoxbro:
> Notably Bokeh, which is maybe where you saw it in the first place?

Correct, I chose this because I saw it in Bokeh (and liked the approach).

My reason for exposing the warnings in `__init__.py`  is to have them easily accessible if you want to mute them, with that being in the `pytest` config file or with `warnings.filterwarnings`. This is especially relevant if you are not familiar with the library itself. Having to dig through the code and find the place with the warning is annoying (or maybe I'm doing it  wrong...)

My belief is even though the code itself is in a private module, the warning itself is public because we expose it to our users, and to a certain degree, they have to interact with it either by looking at the red warning or by muting it.

> ... as the warning name precedes the warning message (ush this temporary file path, yuk):

Yeah, I'm not a fan of that either, but that is how Jupyter works. It is much nicer if the warning is emitted for a Python file, which easily shows your the file and line where the warning occurs.

maximlt:
Thanks for your input!

I will keep things as is for two reasons:
- I think the main reason for us having to provide these specific warnings are to turn them into errors in our test suites, while what we should do is set up our test suites to fail on any warning and ignore those we deem are fine to ignore :) So I see that as an internal hack that we may want to remove one day. Not exposing it more doesn't prevent people from what they're usually doing when filtering a warning, using some regex and module location (as in https://github.com/holoviz/holoviews/blob/8df58d3b485a38236f979f21c2adc9b2a00fa6a0/pyproject.toml#L12-L28), since these warnings are subclasses of Python warnings.
- I hope that in Param 2.x we'll sit and figure out whether the logging system Param implements, that is mostly used to do warning, is actually needed and if instead we shouldn't rely on standard warnings. Because of that, I don't really want to set a too strong precedent with how Param handles warnings.

Issue #644, opened 2022-07-12
Setting default value of 'name' in a Parameterized class doesn't work


chbrandt:
#### ALL software version info
```
channels:
  - conda-forge
  - defaults
dependencies:
  - ipykernel=6.15.1
  - ipython=8.4.0
  - jupyter_client=7.3.4
  - jupyter_core=4.10.0
  - libcxx=14.0.6
  - param=1.12.2
  - python=3.10.5
  - traitlets=5.3.0
```
- MacOS 11.6.5

#### Description of expected behavior and the observed behavior

Setting a default (string) value to Parameterized class' `name` attribute doesn't work (as expected).
A given _default_ value is overwritten by an incremental "`Classname<SERIAL>`" value.

We can set the/a value later, though.

I was expecting the default value to be used as initial value.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
class Person(param.Parameterized):
    name = param.String('ze')
    birthyear = param.Integer(1991)

p = Person()
p
# Person(birthyear=1991, name='Person00003')
p.name = 'ze'
p
# Person(birthyear=1991, name='ze')
```

#### Screenshots or screencasts of the bug in action
- Notebook playing that: https://github.com/docdatpie/params/blob/c8b6efaab6146bf08385d0c6e8dfec1f1b05fbc3/oneOf.ipynb

Hoxbro:
You shouldn't use the variable `name` as it is reserved. A more in-depth description is given [here](https://discourse.holoviz.org/t/creating-an-optional-datetime-parameter-using-param/3147/10), which I have copied below.

> 4. name is the one Parameter that Parameterized objects always have, and it’s also the only one that has special semantics in Param: If the name is not set on an instance or in its constructor, it has a name autogenerated for it from the class name. The intention of this mechanism is for each object to have some relatively unique name to distinguish it when exploring collections of objects, though that’s not necessarily always useful. In any case, here, if we want it to have the value John Doe as in the Pydantic example, we have to provide that name explicitly on the instance as above. The original Pydantic example doesn’t make a whole lot of sense for name anyway, as it would be a recipe for errors for something like a name to inherit a concrete value like John Doe from the base class.

jbednar:
[...]

PR #740, opened 2023-04-13
Overriding the `name` Parameter is taken into account at the class and instance level

maximlt:
Fixes https://github.com/holoviz/param/issues/644

Overriding the `name` String Parameter was possible at the class level but it actually had no effect. This PR changes that so it is taken into account. This is the new behavior:

```python
import param

class P(param.Parameterized):
    name = param.String(default='other')

assert P.name == 'other'

p = P()

assert p.name == 'other'
```

Overriding is only allowed when `name` is set to be a `String` Parameter, as this is something that `Param` relies on internally for other things. A Parameterized class and its instances will have its `name` overriden if the `default` value of the `name` String Parameter is set to something that is not `None` or not an empty string.

I used this branch to run the test suites of Panel and HoloViews (both somewhat partially as I don't have all the dependencies installed locally), which helped me to refine it, as there are indeed a few cases of `name` being overriden by both Panel and HoloViews.

- [x] I still have a question to resolve. For the need of the implementation I have set a new internal attribute on `Parameterized`, whose value is set to `True` when that particular class overrides the value of `name`. I am not particularly happy with adding new attributes, even if internal, to `Parameterized` as there is still a chance they will interact with some user defined attribute. However, I played around with some different implementation and this is the only one I found that worked. I would like this variable name to be as "hidden" as possible, I have called it `__renamed` so that it's mangled, and it happens I can always retrieve it later on the class via the attribute name `'_ParameterizedMetaclass__renamed'`. I don't find that particularly simple to understand and clean, even if I think it does a pretty good job at making the attribute internal and pretty safe from a user perspective. I'm open to better ideas on naming this internal attribute!

To the reviewers. I suggest to have a close look at the tests that have been added and capture the new behavior. Notably a test I added just before opening this PR on *main* needed to be updated (`test_name_instance_generated_class_name_reset`).

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2894,14 +2894,15 @@ class ParameterizedMetaclass(type):
             setattr(param,'objtype',mcs)
             del slots['objtype']
 
+        supers = classlist(mcs)[::-1]
+
         # instantiate is handled specially
-        for superclass in classlist(mcs)[::-1]:
+        for superclass in supers:
             super_param = superclass.__dict__.get(param_name)
             if isinstance(super_param, Parameter) and super_param.instantiate is True:
                 param.instantiate=True
         del slots['instantiate']
 
-        supers = classlist(mcs)[::-1]
         callables = {}
         for slot in slots.keys():
             superclasses = iter(supers)
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2678,10 +2678,10 @@ class ParameterizedMetaclass(type):
         default values (see __param_inheritance()) and setting
         attrib_names (see _set_names()).
         """
+        # print('DEBUG', mcs, name, bases, dict_)
         type.__init__(mcs, name, bases, dict_)
 
-        # Give Parameterized classes a useful 'name' attribute.
-        mcs.name = name
+        mcs.__set_name(name, dict_)
 
         mcs._parameters_state = {
             "BATCH_WATCH": False, # If true, Event and watcher objects are queued.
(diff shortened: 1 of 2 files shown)


jbednar:
I'm very much in support of allowing changes to the name parameter, but I wonder if there is a simpler way now that we have Undefined available. Can we make the name lookup bottom out in something that references the auto-generated name, and otherwise allow the user-specified name to be used (as it won't bottom out in that case)? I may not be understanding all the details involved, of course.

maximlt:
> I'm very much in support of allowing changes to the name parameter, but I wonder if there is a simpler way now that we have Undefined available. Can we make the name lookup bottom out in something that references the auto-generated name, and otherwise allow the user-specified name to be used (as it won't bottom out in that case)? I may not be understanding all the details involved, of course.

I remember that when I implemented this I tried an approach in this vain, and had to give up (which doesn't mean it's not possible!). `name` isn't quite like any other Parameter on a Parameterized class, as I realized. The class is given a `name` that is the class name (what you get with the more regular `type(cls).__name__`.

https://github.com/holoviz/param/blob/aa484c4305c569a182289793307bd46ee078bf68/param/parameterized.py#L2751

And every instance of a Parameterized class gets by default a unique `name` value that is independent of any class hierarchy, while with `Undefined` what we're doing is respecting the class hierarchy.

maximlt:
> If anything, I would like the default `name` to be something that is optional (set by a flag somewhere) so that we it could be deprecated, or at the very least disabled by people who do not wish to have this attribute (`name` is itself a common name in code that people may have other, important uses for).

I would also like Parameterized classes not to have a default `name`. However this is definitely not going to happen in this PR. Let's please open a new issue (no one dared opening this issue already?) and discuss that more, for post Param 2.0.

philippjfr (approved):
Let's get this merged! The `__renamed` attribute is a little bit unfortunate but I don't see a better way.

Issue #703, opened 2023-03-10
Drop support of Python 2.7 and 3.6 for Param 2.0


maximlt:
I think we agreed on dropping support of Python 2.7. I'm not sure about which minimum Python 3 version we should support, I *think* we mentioned 3.7 and that would be my vote, which means dropping support of Python 3.6.

Mostly opening this issue so that we don't forget to do that for Param 2.0, as maintaining Python 2.7 in particular is quite a maintenance burden at this stage.

PR #741, opened 2023-04-16
Upgrade code base to Python 3

maximlt:
Fixes https://github.com/holoviz/param/issues/703

I ran `pyupgrade` on the code base to upgrade it to Python 3.7, and then manually made some other upgrades. Notable changes include:

- Added `watch` and `on_init` keywords to the signature of `depends`
- Deprecated `param.as_unicode`
- Deprecated `param.is_ordered_dict`

Edit:

- Added `warn` keyword to the signature of `param_union`


param/ipython.py
```diff
@@ -168,7 +160,7 @@ def _build_table(self, info, order, max_col_len=40, only_changed=False):
             readonly = 'RO' if p.readonly else 'RW'
             allow_None = ' AN' if hasattr(p, 'allow_None') and p.allow_None else ''
 
-            mode = '%s %s%s' % (constant, readonly, allow_None)
+            mode = '{} {}{}'.format(constant, readonly, allow_None)
```
philippjfr, line 163: f-strings seem nicer.

jlstevens: Agreed

jlstevens:
Looks mostly like I would expect though I would definitely prefer f-strings everywhere. Currently it is removing all the `%s` format strings and replacing it with a style that isn't that much better imho.

param/parameterized.py
```diff
@@ -855,7 +850,7 @@ def __getattribute__(mcs,name):
 
 
 @add_metaclass(ParameterMetaclass)
```
Hoxbro, line 852: `add_metaclass` is a Python 2 thing. 

https://github.com/holoviz/param/blob/2a974928537519895a814ef9c1027df7b4b6c00e/param/parameterized.py#L284-L292

maximlt: Indeed, I have addressed that in another PR actually https://github.com/holoviz/param/pull/734/files#diff-c8fc4fb68012b66fbf42ad9e0b4714fd96a3a72b83558bc543130d78bd622eb3R889

maximlt:
I've applied the suggested `ruff` command.

Issue #699, opened 2023-03-07
Remove `bounds` slot from `Boolean`?


maximlt:
The `Boolean` Parameter creates a unique slot `bounds` that is set to `(0, 1)` by default. The source code has this comment:

>     # Bounds are set for consistency and are arguably accurate, but have
>     # no effect since values are either False, True, or None (if allowed).

`Boolean` actually directly inherits from `Parameter`, I don't really understand what consistency means in this context. Maybe at some point `Boolean` was inheriting from a Number-ish Parameter? To me it looks like `bounds` could be entirely removed from the `Boolean`'s API.

PR #744, opened 2023-04-17
Remove `bounds` slot on `Boolean`

maximlt:
Fixes https://github.com/holoviz/param/issues/699

I could not see a good reason to keep `bounds` on the `Boolean` attribute.


jbednar (approved):
Thanks for tidying this up!

maximlt:
Thanks for the review, merging.

Issue #543, opened 2021-10-07
Param method API plan for 2.0


jbednar:
For Param 2.0, we have a chance to make API improvements that we've been building up to for some time. These were described in https://github.com/holoviz/param/issues/233 and issues linked to that, but this new issue will be focused on the changes we will actually make for 2.0, not just ones we think might be nice but won't get to. Here we'll focus specifically on what to do with the methods on Parameterized and on the `param` subobject of Parameterized, which dominate the namespace of a Parameterized object. Everything listed by `dir(param.Parameterized())` and `dir(param.Parameterized().param)` is included below, along with a disposition using the following key:

Key:
- **dunder**: Pre-existing (though sometimes overridden) magic Python internal methods, that API is generally fixed and not up for discussion.
- **private**: Private API that should stay private. Might not be great API, but also not part of this discussion.
- **delete**: Deprecated API that should be deleted for v2.0.
- **makeprivate**: Rename to a private method
- **makepublic**: Rename to a public method
- **rename**: Rename on current object
- **move**: Move to a different object
- **keep**: Plan to keep as-is.
- **add**: Add new functionality (not actually in dir() currently)
- Options for Done: N/A (no action needed), deprecated (marked deprecated but not yet removed), PR# (where changed)

| Done?	| Disposition | name                | Notes                                  |
|:-----:|:-----------:|:--------------------|:---------------------------------------|
| N/A	| dunder      | `.__class__`
| N/A	| dunder      | `.__delattr__`
| N/A	| dunder      | `.__dict__`
| N/A	| dunder      | `.__dir__`
| N/A	| dunder      | `.__doc__`
| N/A	| dunder      | `.__eq__`
| N/A	| dunder      | `.__format__`
| N/A	| dunder      | `.__ge__`
| N/A	| dunder      | `.__getattribute__`
| N/A	| dunder      | `.__getstate__`
| N/A	| dunder      | `.__gt__`
| N/A	| dunder      | `.__hash__`
| N/A	| dunder      | `.__init__`
| N/A	| dunder      | `.__init_subclass__`
| N/A	| dunder      | `.__le__`
| N/A	| dunder      | `.__lt__`
| N/A	| dunder      | `.__module__`
| N/A	| dunder      | `.__ne__`
| N/A	| dunder      | `.__new__`
| N/A	| dunder      | `.__reduce__`
| N/A	| dunder      | `.__reduce_ex__`
| N/A	| dunder      | `.__repr__`
[...]

PR #745, opened 2023-04-17
Make `Parameterized.initialized` private

maximlt:
I think this one slipped through the cracks until now, but it was noted in the Param 2.0 plan that it should be made private https://github.com/holoviz/param/issues/543

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1273,7 +1273,7 @@ class Parameter(object):
             elif obj is None:
                 _old = self.default
                 self.default = val
-            elif not obj.initialized:
+            elif not obj._initialized:
                 _old = obj.__dict__.get(self._internal_name, self.default)
                 obj.__dict__[self._internal_name] = val
             else:
@@ -1291,7 +1291,7 @@ class Parameter(object):
         self._post_setter(obj, val)
 
         if obj is not None:
-            if not getattr(obj, 'initialized', False):
+            if not getattr(obj, '_initialized', False):
                 return
             obj.param._update_deps(self.name)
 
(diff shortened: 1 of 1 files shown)


maximlt:
Hmm that's why I don't like so much internal attributes on a Parameterized class, they can easily clash with user code:

- https://github.com/holoviz/panel/blob/e93b963751cf3ca1219fcfcfc66461047a3f842e/panel/pane/holoviews.py#L108
- https://github.com/holoviz/lumen/blob/7f4d22a2b47ec6ea2209bc90a915398786a00e72/lumen/views/base.py#L135
- https://github.com/holoviz-topics/EarthSim/blob/5d6bde55feaa6e95e292a90575dfc5179f5b3beb/earthsim/grabcut.py#L190

I really wonder what's the best way to avoid this kind of situation. Mangling? Relying on a private namespace a la `.param`?

jlstevens:
Thanks for this! 

As you know, this particular issue recently bit me.

I think *both* `initialized` and `_initialized` are common and likely to clash. As you say, attributes on parameterized classes often cause clashing namespaces: for this reason I suggest replace `.initialized` with something even less likely to be problematic e.g. `_initialized_` (two underscores should do the trick!)

Hoxbro:
I agree with @jlstevens about the high likelihood of a name clash, though I would suggest `__initialized` instead. This is because I never _really_ look at the end of the variable but always at the start.

jbednar:
I agree with the goal of this PR, but I also agree with the observation that `_initialization` is also a common idiom, and so I'm not sure it's the right fix. `__initialized` seems fine if it can be made to work, or else putting such attributes under .param in in a very obscurely named attribute seems better.

jlstevens:
If sticking `_initialized` on the param namespace object works and is easier than dealing with name mangling, I think that would be a big improvement.

maximlt:
Superseded by https://github.com/holoviz/param/pull/766

Issue #723, opened 2023-03-26
pre-commit-config.yaml is not a file


MarcSkovMadsen:
To my surprise. Param does not support pre-commit! That must be a bug :-)

Please support

## Reproduce

```bash
$ pre-commit run --all
An error has occurred: InvalidConfigError: 
=====> .pre-commit-config.yaml is not a file
Check the log at C:\Users\...\.cache\pre-commit\pre-commit.log
```

maximlt:
Definitely not a bug.

Issue #678, opened 2023-01-14
Distributing tests in PyPI tarball


TheChymera:
It is often preferable for distributions to refer to PyPI tarballs rather than github ones. This is because PyPI does a better job at making sure that releases aren't retroactively changed, whereas on github this often happens, leading to checksum errors.

Gentoo distributes you package, but sadly we cannot provide testing support if we use PyPI, as the test files are not included:

```
[deco]/tmp/param ❱ tar xvf param-1.12.3.tar.gz
param-1.12.3/
param-1.12.3/LICENSE.txt
param-1.12.3/MANIFEST.in
param-1.12.3/PKG-INFO
param-1.12.3/README.md
param-1.12.3/numbergen/
param-1.12.3/numbergen/__init__.py
param-1.12.3/param/
param-1.12.3/param/.version
param-1.12.3/param/__init__.py
param-1.12.3/param/_async.py
param-1.12.3/param/ipython.py
param-1.12.3/param/parameterized.py
param-1.12.3/param/serializer.py
param-1.12.3/param/version.py
param-1.12.3/param.egg-info/
param-1.12.3/param.egg-info/PKG-INFO
param-1.12.3/param.egg-info/SOURCES.txt
param-1.12.3/param.egg-info/dependency_links.txt
param-1.12.3/param.egg-info/requires.txt
param-1.12.3/param.egg-info/top_level.txt
param-1.12.3/setup.cfg
param-1.12.3/setup.py
[deco]/tmp/param ❱ wget https://github.com/holoviz/param/archive/refs/tags/v1.12.3.tar.gz
--2023-01-14 08:18:41--  https://github.com/holoviz/param/archive/refs/tags/v1.12.3.tar.gz
Resolving github.com... 140.82.112.3
Connecting to github.com|140.82.112.3|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://codeload.github.com/holoviz/param/tar.gz/refs/tags/v1.12.3 [following]
--2023-01-14 08:18:41--  https://codeload.github.com/holoviz/param/tar.gz/refs/tags/v1.12.3
Resolving codeload.github.com... 140.82.112.10
Connecting to codeload.github.com|140.82.112.10|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: unspecified [application/x-gzip]
Saving to: ‘v1.12.3.tar.gz’

[...]

PR #679, opened 2023-01-16
Add tests to sdist

ianthomas23:
Fixes #678.

This PR adds the `tests` directory to the `sdist` so that those who build the library from the `sdist` rather than github repo have access to the tests. This includes Linux distribution packagers.

Contents of `sdist` are now:
```
$ tar tf param-1.12.3.post6+g00d89ba.tar.gz
param-1.12.3.post6+g00d89ba/
param-1.12.3.post6+g00d89ba/LICENSE.txt
param-1.12.3.post6+g00d89ba/MANIFEST.in
param-1.12.3.post6+g00d89ba/PKG-INFO
param-1.12.3.post6+g00d89ba/README.md
param-1.12.3.post6+g00d89ba/numbergen/
param-1.12.3.post6+g00d89ba/numbergen/__init__.py
param-1.12.3.post6+g00d89ba/param/
param-1.12.3.post6+g00d89ba/param/.version
param-1.12.3.post6+g00d89ba/param/__init__.py
param-1.12.3.post6+g00d89ba/param/_async.py
param-1.12.3.post6+g00d89ba/param/ipython.py
param-1.12.3.post6+g00d89ba/param/parameterized.py
param-1.12.3.post6+g00d89ba/param/serializer.py
param-1.12.3.post6+g00d89ba/param/version.py
param-1.12.3.post6+g00d89ba/param.egg-info/
param-1.12.3.post6+g00d89ba/param.egg-info/PKG-INFO
param-1.12.3.post6+g00d89ba/param.egg-info/SOURCES.txt
param-1.12.3.post6+g00d89ba/param.egg-info/dependency_links.txt
param-1.12.3.post6+g00d89ba/param.egg-info/requires.txt
param-1.12.3.post6+g00d89ba/param.egg-info/top_level.txt
param-1.12.3.post6+g00d89ba/setup.cfg
param-1.12.3.post6+g00d89ba/setup.py
param-1.12.3.post6+g00d89ba/tests/
param-1.12.3.post6+g00d89ba/tests/API0/
param-1.12.3.post6+g00d89ba/tests/API0/__init__.py
param-1.12.3.post6+g00d89ba/tests/API0/testcalendardateparam.py
param-1.12.3.post6+g00d89ba/tests/API0/testcalendardaterangeparam.py
param-1.12.3.post6+g00d89ba/tests/API0/testclassselector.py
param-1.12.3.post6+g00d89ba/tests/API0/testcolorparameter.py
[...]

PR #749, opened 2023-04-29
Upgrade the infrastructure to hatch, pyproject.toml and pre-commit

maximlt:
Fixes https://github.com/holoviz/param/issues/647
Fixes https://github.com/holoviz/param/issues/678 (supersedes https://github.com/holoviz/param/pull/679)
Fixes https://github.com/holoviz/param/issues/723


maximlt:
It seems to be working, merging.

.github/workflows/build.yaml
```diff
@@ -39,46 +39,40 @@ jobs:
...
+          python-version: "3.9"
       - name: env setup
         run: |
           python -m pip install --upgrade pip
-          python -m pip install setuptools wheel twine tox
+          python -m pip install build
       - name: pip build
         run: |
-          python setup.py sdist bdist_wheel
+          python -m build
```
Hoxbro, line 75: Did you try `hatch build`?

.github/workflows/docs.yaml
```diff
@@ -45,12 +43,9 @@ jobs:
       - name: env setup
         run: |
           python -m pip install --upgrade pip
-          python -m pip install nbsite==0.8.0rc2
-          python -m pip install -e .[doc]
+          python -m pip install hatch
```
Hoxbro, line 46: Personal preference but I would use `pipx` to install `hatch`.

.github/workflows/build.yaml
maximlt: Yes, both with the same result. I saw `hatch` is using `build` in its CI (https://github.com/pypa/hatch/blob/master/.github/workflows/build-hatch.yml#L23). Also, I'm not sure whether `hatch build` can build the package if another backend is used, while I believe `python -m build` will build whatever the backend.

.github/workflows/docs.yaml
Hoxbro: I would use `pipx` mostly to isolate `hatch`. 

And it is available in the CI without installing it. I use it [here](https://github.com/Hoxbro/clean_notebook/blob/main/.github/workflows/test.yaml#L32)

.github/workflows/build.yaml
Hoxbro: It could be a chicken and egg situation. 

I was mostly curious about it :+1:

.github/workflows/docs.yaml
maximlt: Ah interesting I didn't know it was available on the runners. But yeah as you said, that's personal preference at this stage, I think it's fine installing `hatch` with pip instead of with `pipx`. I'll change my mind when this breaks!

Issue #635, opened 2022-05-31
Callback triggered when depending on a subobject not declared


maximlt:
From https://github.com/holoviz/panel/issues/3580

I didn't expect that setting the `test_param` attribute in the example below would trigger the callback, but it does.

```python
import param

class Sub(param.Parameterized):
    s = param.String()

class P(param.Parameterized):
    test_param = param.Parameter()

    def __init__(self, **params):
        self._sub = Sub()
        super().__init__(**params)

    @param.depends('_sub.s', watch=True)
    def cb(self):
        print('cb called')

p = P()
p.test_param = 'should not trigger'
# 'cb called'
```

A simple fix consists in adding `_sub = param.Parameter()` to the Parameters of `P`.

maximlt:
@jlstevens do you know if this pattern, i.e. depending on a subobject that is not one of the Parameters of the class but is supported by Param? I think it is as I've seen some usage of this in Panel but I'd like a confirmation if possible.

https://github.com/holoviz/panel/blob/120019e4318ac51bc2b9d0a1b2eb2239c8a0c9ad/panel/widgets/input.py#L891-L904

jbednar:
> A simple fix consists in adding `_sub = param.Parameter()` to the Parameters of P.

In this case I don't see any reason not to declare _sub as a Parameter, because then you don't need to write an `__init__` method at all. Plus it's then clear to anyone that `_sub` is available as a Parameter to be depended upon.

That said, if we do continue to support depending on attributes that are not Parameters, then I agree that the behavior illustrated is surprising and not at all desirable. I think it would take some debugging to figure out, e.g. by raising an Exception in the callback and tracing back to figure out how it got executed. My guess is that _sub.s is somehow being initialized in some way that is triggered by setting `test_param`, but I don't know why that would be. Will probably involve some head scratching!

maximlt:
My intuition is that we get a mix of two things:

* even if `_sub` is not a Parameter the callback is correctly registered on `_sub.s` with `watch=True`
* methods in a Parameterized class automatically depend on all the Parameters if they're not decorated with `depends`, with the behavior of `watch=False`

We get a mix of these two things, so the callback is triggered on changes of `_sub.s` and any other declared Parameter value, with `watch=True`.

Total guess here, could be completely wrong.

jbednar:
> methods in a Parameterized class automatically depend on all the Parameters if they're not decorated with depends, 

That doesn't seem to apply in this case, which does have `depends`.

[...]

PR #753, opened 2023-05-23
Ensure non-function types are not resolved as empty function declarations

philippjfr:
Fixes https://github.com/holoviz/param/issues/635


maximlt:
Great thanks for the quick fix! I added a test. Merging!

PR #763, opened 2023-06-14
Fix stacklevel start

### LINKED ISSUE #756: Warnings not displaying the right frame
@Hoxbro reported internally some issues with how the warnings are set up. For instance this one with the warning not being emitted from the right spot while running the GeoViews test suite:

image

Hoxbro:
fixes #756

The decorator adds an extra stack level. 

``` python
import param

class Defaults(param.Parameterized):
    filters = param.List(doc="Defaults for Filter objects.", class_=dict)
```

Before: 
image

After:
image


philippjfr (approved):
Looks great!

maximlt:
Thanks!

PR #767, opened 2023-06-20
More Param 2.0 API cleanup

maximlt:
My priority with this PR was to further clean up the Parameterized namespace, following and even going further the dispositions suggested in https://github.com/holoviz/param/issues/543

- I looked through all the holoviz repos + examples for usage of `.pprint` or `._pprint` and found none, so I decided to remove them all together, using `.param.pprint()` was anyway the recommendation in the Version 1.12.0 release notes.
- The same search showed no usage of `.script_repr` that has been deleted from `Parameterized` and also from `ParameterizedFunction`.
- Instead of just deleting `.state_pop` in favor of `._state_pop`, I decided to move it (and `.state_push`) to `.param._state_pop`. This API isn't documented and was planned to be made private, this is why I chose to make it private on the `.param` namespace.

EDIT:
- Removed `Time.next` that was used for compatibility with Python 2

### DIFF
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -2,6 +2,7 @@ import inspect
 import functools
 import warnings
 
+from textwrap import dedent
 from threading import get_ident
 
 
@@ -38,7 +39,9 @@ def _deprecated(extra_msg="", warning_cat=ParamDeprecationWarning):
         def inner(*args, **kwargs):
             msg = f"{func.__name__!r} has been deprecated and will be removed in a future version."
             if extra_msg:
-                msg = msg + ' ' + extra_msg
+                em = dedent(extra_msg)
+                em = em.strip().replace('\n', ' ')
+                msg = msg + ' ' + em
             warnings.warn(msg, category=warning_cat, stacklevel=2)
             return func(*args, **kwargs)
         return inner
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -2,6 +2,8 @@ import inspect
 import functools
 import warnings
 
+from threading import get_ident
+
 
 class ParamDeprecationWarning(DeprecationWarning):
     """Param DeprecationWarning"""
@@ -64,3 +66,25 @@ def _deprecate_positional_args(func):
         return func(*args, **kwargs)
 
     return inner
+
+
+# Copy of Python 3.2 reprlib's recursive_repr but allowing extra arguments
+def _recursive_repr(fillvalue='...'):
+    'Decorator to make a repr function return fillvalue for a recursive call'
+
+    def decorating_function(user_function):
+        repr_running = set()
+
+        def wrapper(self, *args, **kwargs):
+            key = id(self), get_ident()
+            if key in repr_running:
+                return fillvalue
+            repr_running.add(key)
+            try:
+                result = user_function(self, *args, **kwargs)
+            finally:
+                repr_running.discard(key)
+            return result
+        return wrapper
+
+    return decorating_function
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -3382,8 +3382,8 @@ def pprint(val,imports=None, prefix="\n    ", settings=[],
     elif type(val) in script_repr_reg:
         rep = script_repr_reg[type(val)](val,imports,prefix,settings)
 
-    elif hasattr(val,'_pprint'):
-        rep=val._pprint(imports=imports, prefix=prefix+"    ",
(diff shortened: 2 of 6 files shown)


param/_utils.py
```diff
@@ -64,3 +66,25 @@ def inner(*args, **kwargs):
...
+
+        def wrapper(self, *args, **kwargs):
+            key = id(self), get_ident()
+            if key in repr_running:
+                return fillvalue
+            repr_running.add(key)
+            try:
+                result = user_function(self, *args, **kwargs)
+            finally:
+                repr_running.discard(key)
```
ahuang11, line 86: Oooh learned that this is like remove, but doesn't error if missing.

maximlt: Oh indeed, I also didn't know about it :)

param/parameterized.py
```diff
@@ -2826,13 +2831,135 @@ def log(self_, level, msg, *args, **kw):
...
+                           if parameters[k].precedence is None else
+                           parameters[k].precedence))
+
+        arglist, keywords, processed = [], [], []
+        for k in args + ordering:
+            if k in processed: continue
+
+            # Suppresses automatically generated names.
+            if k == 'name' and (values[k] is not None
+                                and re.match('^'+self.__class__.__name__+'[0-9]+$', values[k])):
```
Hoxbro, line 2929: I think Regex is overkill here. But I can't seem to get a better speedup with other methods for a simple example. Maybe it is caching...

It could be improved by using `re.compile` outside the for loop. Maybe also limit the numbers to 5.

param/parameterized.py
```diff
@@ -2826,13 +2831,135 @@ def log(self_, level, msg, *args, **kw):
...
+                continue
+
+            value = pprint(values[k], imports, prefix=prefix,settings=[],
+                           unknown_value=unknown_value,
+                           qualify=qualify) if k in values else None
+
+            if value is None:
+                if unknown_value is False:
+                    raise Exception(f"{self.name}: unknown value of {k!r}")
```
Hoxbro, line 2938: Maybe use ValueError here?

maximlt:
Just for context, Simon added some comments on code that wasn't moved around, thinking that was new code. I agree with the suggestions made, they'd be worth making in other PRs to keep this one focused on API cleanup.

tests/testdeprecations.py
```diff
@@ -19,40 +19,188 @@ def specific_filter():
...
+            class Base: pass
+
+    def test_deprecate_recursive_repr(self):
+        with pytest.raises(param._utils.ParamDeprecationWarning):
+            param.parameterized.recursive_repr(lambda: '')
+
+
+class TestDeprecateParameters:
+
+    def test_deprecate_print_param_defaults(self):
```
Hoxbro, line 104: I would maybe take advantage of using a class and only create the P instance once with `setup_method` https://docs.pytest.org/en/latest/how-to/xunit_setup.html.

maximlt: Yep true the tests I write for Param aren't usually very DRY. I think it's because I've struggled with the original test suites, where some test modules have a setup that defines a Parameterized class with lots of Parameters, and when working in these modules I end up spending way too much time scrolling 🙃 So I've come down to preferring repeating myself and having self-contained tests.

param/parameterized.py
```diff
@@ -3593,169 +3711,8 @@ def __str__(self):
         """Return a short representation of the name and class of this object."""
         return f"<{self.__class__.__name__} {self.name}>"
 
-    # PARAM3_DEPRECATION
-    @_deprecated(extra_msg="Use instead `.param.pprint()`")
-    def script_repr(self,imports=[],prefix="    "):
```
jlstevens, line 3598: @maximlt Now the methods are removed, I believe the `pprint` and `script_repr` functions won't work either. It looks like you might be able to use `pprint` on the `param` namespace instead...

jlstevens:
The removals in this PR all make sense to me, other than one comment, looks good to me!

jlstevens (approved):

philippjfr (approved):
Looks good to me!

param/parameterized.py
maximlt: Thanks for catching that! Addressed in https://github.com/holoviz/param/pull/767/commits/7f606a7e618aa9e89d2736102f8e4a038f427d51 after adding some simple tests that would have caught this regression in https://github.com/holoviz/param/pull/778.

Issue #707, opened 2023-03-10
Deprecate utility functions


maximlt:
I came across many functions in the `__init__` namespace that could potentially be cleaned up. I've looked for usages of these functions inside Param's code base and across all the HoloViz/PyViz repositories. Unless I mention it, these functions are not documented in the user guides. They're all documented in the API reference though. Here's a list of clean up suggestions:

- [x] Deprecate `normalize_path`, and remove

`normalize_path` is a ParameterizedFunction that isn't used anywhere I looked.

https://github.com/holoviz/param/blob/cd1b4ae50c01c14c0eb22721e41a9c8fb1e8f738/param/__init__.py#L1726-L1748

- [x] Deprecate `produce_value`, and turn into a private function `_produce_value`

General utility function only used internally.

https://github.com/holoviz/param/blob/cd1b4ae50c01c14c0eb22721e41a9c8fb1e8f738/param/__init__.py#L68-L77

- [x] Deprecate `as_unicode`, and remove to replace it with the built-in `str`

General utility function only used once internally.

https://github.com/holoviz/param/blob/cd1b4ae50c01c14c0eb22721e41a9c8fb1e8f738/param/__init__.py#L80-L87

- [x] Deprecate `is_ordered_dict`, and turn into a private function `_is_ordered_dict`

General utility function only used once internally.

https://github.com/holoviz/param/blob/cd1b4ae50c01c14c0eb22721e41a9c8fb1e8f738/param/__init__.py#L90-L97

- [x] Deprecate `hashable`, remove mention from the documentation and update the documentation, and turn into a private function `_hashable`

Specific utility function only used internally. Panel has its own identical and unused implementation (https://github.com/holoviz/panel/blob/8e6798cc0f1cde5232b9a808ebecf00b5683fd0c/panel/util/__init__.py#L45-L51).

The documentation on `Selector`s says that:

> Otherwise, the objects should be provided as a _name_:_value_ dictionary, where the string name will be stored for use in such a UI, but is not otherwise accessed by Param. The values from setting and getting the parameter are always the actual underlying object, not the string names. Because the string name will need to be looked up from the value if this parameter is used in a UI, all objects need to be hashable via the `param.hashable()` function, which accepts Python literals plus list and dictionary types (treating them like tuples).

However, I believe this is no longer true since https://github.com/holoviz/param/pull/632.

[...]

PR #768, opened 2023-06-20
Deprecate private or unused functions in `__init__.py`

maximlt:
Fixes https://github.com/holoviz/param/issues/707


maximlt:
This PR just added deprecation warnings for functions to be removed or to make private and includes no functional change. Given that that we agreed on the affected functions in the issue, I'm going to go ahead and merge it.

Issue #747, opened 2023-04-21
watcher doesn't work with __len__


rishikanthc:
#### ALL software version info
param 1.13.10
python 3.10.11

#### Description of expected behavior and the observed behavior
In the code below the watcher doesn't get invoked when the parameter is set. This happens only when I define a `__len__` function in my class. Adding this seems to break the dependency watcher. Without `__len__` the code works as intended

#### Complete, minimal, self-contained example code that reproduces the issue
```python
import param as pm

class WindowDS(pm.Parameterized):
    trials = pm.ListSelector([1], objects=[0, 1, 2, 3, 4, 5], doc="Experiment to load")
    window_size = pm.Integer(100, bounds=(16, 100), doc="window size for samples")

    def __init__(self, **params):
        super().__init__(**params)

    @pm.depends("trials", watch=True)
    def _get_df(self):
        print(f"_get_df invoked")
        self.create_data()

    @pm.depends("window_size", watch=True)
    def create_data(self):
        print(f"creating dataset")

    def __len__(self):
        return 0


train_ds = WindowDS(trials=[1], window_size=20)
train_ds.window_size = 100
```

For the code above The program should print "creating dataset" but nothing gets printed

PR #769, opened 2023-06-21
Fix watchers support when the Parameterized instance is falsey

maximlt:
Fixes https://github.com/holoviz/param/issues/747

`inst or cls` was used to select `cls` when `inst` was None, except that this fails when `inst` is not None and `inst` is falsey evaluates, as in the example reported in the issue where `__len__` is defined on a Parameterized class and returns `0`. This PR fixes a few occurrences of this bug.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -619,7 +619,7 @@ def _params_depended_on(minfo, dynamic=True, intermediate=True):
     deps, dynamic_deps = [], []
     dinfo = getattr(minfo.method, "_dinfo", {})
     for d in dinfo.get('dependencies', list(minfo.cls.param)):
-        ddeps, ddynamic_deps = (minfo.inst or minfo.cls).param._spec_to_obj(d, dynamic, intermediate)
+        ddeps, ddynamic_deps = (minfo.inst if minfo.inst is not None else minfo.cls).param._spec_to_obj(d, dynamic, intermediate)
         dynamic_deps += ddynamic_deps
         for dep in ddeps:
             if isinstance(dep, PInfo):
@@ -1859,7 +1859,7 @@ class Parameters:
                     init_methods.append(m)
             elif dynamic:
                 for w in obj._dynamic_watchers.pop(method, []):
-                    (w.inst or w.cls).param.unwatch(w)
+                    (w.inst if w.inst is not None else w.cls).param.unwatch(w)
             else:
                 continue
 
@@ -1894,7 +1894,7 @@ class Parameters:
             subobj = getattr(subobj, subpath.split(':')[0], None)
             subobjs.append(subobj)
 
-        dep_obj = (param_dep.inst or param_dep.cls)
+        dep_obj = param_dep.inst if param_dep.inst is not None else param_dep.cls
         if dep_obj not in subobjs[:-1]:
             return None, None, param_dep.what
 
@@ -1930,7 +1930,7 @@ class Parameters:
         on the old subobject and create watchers on the new subobject.
         """
         dynamic_dep, param_dep = group[0]
-        dep_obj = (param_dep.inst or param_dep.cls)
+        dep_obj = param_dep.inst if param_dep.inst is not None else param_dep.cls
         params = []
         for _, g in group:
             if g.name not in params:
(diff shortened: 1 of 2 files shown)


tests/testparamdepends.py
```diff
@@ -820,6 +820,27 @@ def method1(self):
...
+                nonlocal count
+                count += 1
+
+            # bool(P()) evaluates to False
+            def __len__(self):
+                return 0
+
+        p = P()
+        p.x = 1
+        assert count == 1
```
Hoxbro, line 842: I would rewrite this to something like this:
``` python
    import param

    class P(param.Parameterized):
        x = param.Parameter()
        count = param.Number(0)

        @param.depends('x', watch=True)
        def debug(self):
            self.count += 1

        # bool(P()) evaluates to False
        def __len__(self):
            return 0

    p = P()
    p.x = 1
    assert p.count == 1
```

Hoxbro: No distinct advantages.

I think this aligns more with a real-world example and as such it is easier (at least for me) to read and understand.

maximlt: I think I still prefer the approach that separates concerns more.

>  it is easier (at least for me) to read and understand.

Oh I'm 100% sure you have no problem understanding that test! :)

Hoxbro: It just looks like you have gone out of your way to do it this way over the parametrized class approach. This could lead to confusion in the future.  

It is very nit, and as I said it was something I would have done.

param/parameterized.py
```diff
@@ -1930,7 +1930,7 @@ def _watch_group(self_, obj, name, queued, group, attribute=None):
         on the old subobject and create watchers on the new subobject.
         """
         dynamic_dep, param_dep = group[0]
-        dep_obj = (param_dep.inst or param_dep.cls)
+        dep_obj = param_dep.inst if param_dep.inst is not None else param_dep.cls
```
philippjfr, line 1933:
```suggestion
        dep_obj = param_dep.cls if param_dep.inst is None else param_dep.inst 
```

param/parameterized.py
```diff
@@ -1894,7 +1894,7 @@ def _resolve_dynamic_deps(self, obj, dynamic_dep, param_dep, attribute):
             subobj = getattr(subobj, subpath.split(':')[0], None)
             subobjs.append(subobj)
 
-        dep_obj = (param_dep.inst or param_dep.cls)
+        dep_obj = param_dep.inst if param_dep.inst is not None else param_dep.cls
```
philippjfr, line 1897:
```suggestion
        dep_obj = param_dep.cls if param_dep.inst is None else param_dep.inst
```

param/parameterized.py
```diff
@@ -1859,7 +1859,7 @@ def _update_deps(self_, attribute=None, init=False):
                     init_methods.append(m)
             elif dynamic:
                 for w in obj._dynamic_watchers.pop(method, []):
-                    (w.inst or w.cls).param.unwatch(w)
+                    (w.inst if w.inst is not None else w.cls).param.unwatch(w)
```
philippjfr, line 1862:
```suggestion
                    (w.cls if w.inst is None else w.inst).param.unwatch(w)
```

param/parameterized.py
```diff
@@ -619,7 +619,7 @@ def _params_depended_on(minfo, dynamic=True, intermediate=True):
     deps, dynamic_deps = [], []
     dinfo = getattr(minfo.method, "_dinfo", {})
     for d in dinfo.get('dependencies', list(minfo.cls.param)):
-        ddeps, ddynamic_deps = (minfo.inst or minfo.cls).param._spec_to_obj(d, dynamic, intermediate)
+        ddeps, ddynamic_deps = (minfo.inst if minfo.inst is not None else minfo.cls).param._spec_to_obj(d, dynamic, intermediate)
```
philippjfr, line 622:
```suggestion
        ddeps, ddynamic_deps = (minfo.cls if minfo.inst is None else minfo.inst).param._spec_to_obj(d, dynamic, intermediate)
```

philippjfr:
Minor, nitpicky comments.

PR #771, opened 2023-06-22
Regression: Re-set instantiate to True when constant is True

maximlt:
Another evidence that it's quite easy to break Param despite its own test suite and the test suite of other packages heavily relying on it. This PR fixes a regression introduced in https://github.com/holoviz/param/pull/605, whereby setting constant to `True` was no longer automatically setting `instantiate` to True. I've added tests that would have passed on Param `1.13.0`.


jbednar:
The change looks good, thanks for catching it.

Issue #287, opened 2018-10-02
Why force instantiate for constant=True?


jbednar:
I'm confused by this behavior:

```
import param

class A(param.Parameterized):
    a = param.Number(2)

aa = A(name="AAA")

class B(param.Parameterized):
    b = param.ObjectSelector(aa)
    c = param.ObjectSelector(aa, readonly=True)
    c = param.ObjectSelector(aa, constant=True)

bb = B(name="BBB")
print(aa, B.b, B.c, bb.b, bb.c)
```

&lt;A AAA&gt; &lt;A AAA&gt; &lt;A AAA&gt; &lt;A AAA&gt; &lt;A A00004&gt;

I don't see why bb.c differs from any of the other objects shown -- why is it not the specific object that I supplied for parameter `b`, and instead is a new copy of that object? This was just the source of a very confusing behavior when using ObjectSelector.  

The [code involved](https://github.com/ioam/param/blob/master/param/parameterized.py#L485
) states that constant parameters must be instantiated, but I don't see why that should be true for ObjectSelector, which when instantiate=False (the default) is meant to select between *existing* objects. Should we start overriding this general `constant=True` behavior to avoid it for ObjectSelector?

philippjfr:
>Should we start overriding this general constant=True behavior to avoid it for ObjectSelector?

I definitely think so, as you say it should be selecting between existing objects not making new copies. In fact I would have inspected the validation to fail because the copy is not one of the existing objects.

philippjfr:
>As faras I remember, the default is not validated unless objects are supplied (no objects supplied in example above). Not sure how well tested, though.

Yes, thanks, didn't look at the example closely enough.

ceball:
xref #28

jbednar:
This still seems like it needs addressing.

maximlt:
> instantiate happens for constant to avoid the normal behavior where setting on the class would change the value on an existing instance

Just some comments to make this issue a little easier to understand.  The normal Param behavior when `constant=False` is that setting on the class changes the value of an existing instance if the value hasn't been set:

```python
import param

class P(param.Parameterized):
    x = param.Parameter(1)

p = P()
assert p.x  == 1

# Changing the class value (Parameter default)
P.x = 10
# affects the instance if the value hasn't yet been set (instantiation or setattr)
assert p.x  == 10
```

[...]

PR #776, opened 2023-06-29
Don't automatically set instantiate to True when constant is True

maximlt:
Fixes https://github.com/holoviz/param/issues/287

Pushing this WIP PR to get some feedback on the approach. The idea is that when:
1. `instantiate` is `True` -> deepcopy `default` and assign it to the local cache (pseudo: `parameterized_obj._X_param_value = deepcopy(parameter.default)`)
2. `constant` is `True` -> don't deepcopy, just assign a reference to `default` to the local cache (pseudo: `parameterized_obj._X_param_value = parameter.default`)

With 1. taking precedence over 2.

The test added in https://github.com/holoviz/param/commit/7635d55e265b432417bad4ed5571fe4c84371037 and that is based on the one given in the issue passes with this change.

Of course this means that if the `default` object is mutable, any change made from a reference will be seen by all the other references, all pointing to the same updated object.

```python
import param

class A(param.Parameterized):
    x = param.Number(0)

a = A(name="AAA")

class B(param.Parameterized):
    y = param.ObjectSelector(a, constant=True)

b = B(name="BBB")

B.y.x = 1
print(b.y.x)
# 0 before this PR
# 1 with this PR

b.y.x = 2
print(B.y.x)
# 1 before this PR
# 2 with this PR
```

If reviewers are OK with this approach, I'll need to clean up the code and do some refactoring, it's a bit hacky right now.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1263,8 +1263,6 @@ class Parameter(_ParameterBase):
         # having this code avoids needless instantiation.
         if self.readonly:
             self.instantiate = False
-        # elif self.constant is True:
-        #     self.instantiate = True
         elif instantiate is not Undefined:
             self.instantiate = instantiate
         else:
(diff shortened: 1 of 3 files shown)


philippjfr (approved):
This looks good to me. It's always very difficult to consider all the implications of these changes but let's just make sure we run all test suites against this and then merge.

maximlt:
The unit test suites of Panel, HoloViews and Lumen passed with this branch checked out. Merging.

Issue #357, opened 2019-09-26
param.depends does not warn about unresolved parameters


philippjfr:
As the title says the depends decorator simply silently ignores attributes which are not parameters leading to unexpected behavior as in this example: https://stackoverflow.com/questions/58114082/why-is-my-plot-updated-by-panel-twice-when-i-change-a-button-setting-that-shou

We should make sure that it error or warns when a referenced parameter is not found.

maximlt:
To be more explicit about this issue, in the example below we depend on `y` that doesn't exist and there's no warning/error that lets the user know they've done something wrong, either at the class or instance creating time

```python
import param

class A(param.Parameterized):
    x = param.Number(0)

    @param.depends('y')
    def debug(self): pass

a = A()
```

PR #777, opened 2023-06-29
Resolve non watched references too to catch bad references

maximlt:
Fixes https://github.com/holoviz/param/issues/357

@philippjfr I'm very far from being confident with this change.

This example:

```python
import param

class A(param.Parameterized):
    x = param.Number(0)

    @param.depends('y', watch=True)
    def debug(self): pass
```

Would immediately fail at the class creating time with:
```
  File ".mltmess/issue_params.py", line 47, in <module>
    class A(param.Parameterized):
  File "/Users/mliquet/dev/param/param/parameterized.py", line 2903, in __init__
    deps, dynamic_deps = _params_depended_on(minfo, dynamic=False)
  File "/Users/mliquet/dev/param/param/parameterized.py", line 640, in _params_depended_on
    ddeps, ddynamic_deps = (minfo.cls if minfo.inst is None else minfo.inst).param._spec_to_obj(d, dynamic, intermediate)
  File "/Users/mliquet/dev/param/param/parameterized.py", line 2592, in _spec_to_obj
    raise AttributeError(f"Attribute {attr!r} could not be resolved on {src}.")
AttributeError: Attribute 'y' could not be resolved on <class '__main__.A'>.
```

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2898,12 +2898,11 @@ class ParameterizedMetaclass(type):
         for name, method, dinfo in dependers:
             watch = dinfo.get('watch', False)
             on_init = dinfo.get('on_init', False)
-            if not watch:
-                continue
             minfo = MInfo(cls=mcs, inst=None, name=name,
                           method=method)
             deps, dynamic_deps = _params_depended_on(minfo, dynamic=False)
-            _watch.append((name, watch == 'queued', on_init, deps, dynamic_deps))
+            if watch:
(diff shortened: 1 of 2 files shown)


philippjfr:
My guess is that this breaks cases like this, which we didn't intend to support but did accidentally?

```
class A(param.Parameterized):

    def __init__(self):
        self.x = pn.widgets.FloatSlider()
        super().__init__()

    @param.depends('x.value')
    def update_on_x(self):
        return self.x.value * 2
```

philippjfr:
Nvm, still seems to work. Seems okay.

Issue #685, opened 2023-02-10
Add context manager to temporarily set parameter(s)


philippjfr:
In my code I very frequently find myself wanting to temporarily set some parameter and then reset it to its original value. For this purpose it would be super helpful to have a context manager that handles this for me without me having to take care of storing the old value, adding error handling and then resetting the value.

This is particularly useful in cases where you want to configure something while some longer running task is occurring. Prime examples include setting `loading` parameter in Panel:

```python
with pane.param.set(loading=True):
    ... # Do some expensive computation
    pane.object = new_pane
```

or temporarily overriding the styling of some component:

```python
with pane.param.set(styles={'border': '1px solid yellow'}):
    ...
```

Additionally it can be useful when temporarily overriding some configuration option(s):

```python
with pn.config.param.set(resource_mode='inline'):
    ...

with pn.config.param.set(design='bootstrap'):
    ...
```

I personally like `.param.set` as the API as we are temporarily setting parameter values as opposed to `.param.update` which permanently updates the values. I am open to other suggestions though.

jlstevens:
I think this is useful and don't see why it wouldn't be included in param. The only argument against that I can think of is that we want to minimize the API surface on the `param` namespace object (though I think this would be more useful than many things currently on there!)

jlstevens:
Would there be any special handling for read-only/constant parameters? E.g. should it operate the same way or warn?

jbednar:
Frankly I thought we already had this. :-)

jbednar:
Ah, what I was remembering was `with param.edit_constant(pane.param.loading):`. So maybe this should be a standalone function `with param.edit(pane.param.loading)` to avoid clashes with parameters named `set`.

philippjfr:
I like not having to import it though :/

PR #779, opened 2023-06-30
Add Parameterized.param.set context manager

philippjfr:
Implements https://github.com/holoviz/param/issues/685

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2117,6 +2117,19 @@ class Parameters:
                                  (self_or_cls.name))
         return self_.update(kwargs)
 
+    @contextmanager
+    def set(self_, **kwargs):
+        """
+        Context manager that temporarily sets parameter values and then
+        restores the original values on exit.
+        """
+        values = self_.values()
+        restore = {k: values[k] for k, v in kwargs.items() if k in values}
+        try:
+            self_.update(kwargs)
+            yield
+        finally:
+            self_.update(restore)
 
     def objects(self_, instance=True):
         """
--- a/tests/testparameterizedobject.py
+++ b/tests/testparameterizedobject.py
@@ -535,6 +535,29 @@ class TestParameterized(unittest.TestCase):
         with pytest.raises(TypeError, match=re.escape(msg)):
             TestPO(not_a_param=2)
 
+    def test_set_context_single_parameter(self):
+        t = TestPO(inst='foo')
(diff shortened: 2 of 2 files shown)


maximlt:
I should have chimed in https://github.com/holoviz/param/issues/685 back in February, or maybe at the time it was not so obvious to me, but I think that `.param.set` and `.param.update` aren't distinguishable enough naming-wise, that will just cause confusion (constant/readonly kind of confusion comes to mind, since we talked about it recently with Jean-Luc).

How about modifying `.param.update` so that it can also be used as a context manager? Here's a prototyped implementation where calling `.update` returns an instance of `_Update` while previously (as in Param now on the _main_ branch) it returns `None`. I think that's okay?

```python
class _Update:
    
    def __init__(self, instance):
        self.instance = instance
    
    def __call__(self, **kwargs):
        self.restore = self.instance.kwargs
        self.instance._update(**kwargs)
        return self
    
    def __enter__(self):
        print('ENTER')
    
    def __exit__(self, *exc_details):
        print('EXIT')
        self.instance._update(**self.restore)
        self.restore.clear()


class Parameters:
    
    def __init__(self):
        self.kwargs = dict(a=1)
    
    def update(self, **kwargs):
        return _Update(instance=self)(**kwargs)

    def _update(self, **kwargs):
        self.kwargs = kwargs
```

image

If we don't go down that route, we should try to align `.set` and `.update` more (i.e. same signatures, work on both classes and instances).

maximlt:
What do you think about that @philippjfr ?

philippjfr:
I'll play around with it.

param/parameterized.py
```diff
@@ -1640,6 +1640,34 @@ def compare_mapping(cls, obj1, obj2):
...
+        return self
+
+    def __enter__(self):
+        return self._restore
+
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        try:
+            self._parameters._update(self._restore)
+        except Exception as e:
+            raise e
```
Hoxbro, line 1666: I don't think this is needed.

philippjfr:
Made the changes you requested, ready to review.

param/parameterized.py
philippjfr: Indeed.

maximlt (approved):
This PR looked good to me overall, I completed it with:
- more tests
- updated docs
- making ParametersRestore private

Thanks!

Issue #759, opened 2023-06-05
Cloudpickle compatibility 


samimia-swks:
params 1.13.0
cloudpickle  2.2.1
python 3.11
Ubuntu 22.04

The docs recommend Cloudpickle as a way to pickle Parameterized classes with lambda functions, but even the most basic Parameterized object (with or without lambdas) cannot be pickled with cloudpickle :

```python 
import param
import cloudpickle

class MyParams(param.Parameterized):
    a = param.Integer()
    b = param.Integer()
    # sum_pull = Integer(lambda : p.a + p.b)

p = MyParams()
pkl = cloudpickle.dumps(p)
p2 = cloudpickle.loads(pkl)

```

Output
```
File [/param/parameterized.py:1685), in Parameters.__setstate__(self, state)
   1682 def __setstate__(self, state):
   1683     # Set old parameters state on Parameterized._parameters_state
   1684     self_or_cls = state.get('self', state.get('cls'))
-> 1685     for k in self_or_cls._parameters_state:
   1686         key = '_'+k
   1687         if key in state:

AttributeError: 'NoneType' object has no attribute '_parameters_state'
```

maximlt:
Hi @samimia-swks,

`cloudpickle` is indeed mentioned in the docs, albeit in a pretty light way 🙃 

> lambdas cannot easily be pickled for saving and restoring state (though see [cloudpickle](https://github.com/cloudpipe/cloudpickle) for an alternative to pickle that does support lambdas)

Unfortunately it seems that Param doesn't support it out of the box, even for very simple cases like the one you tried above. It'd be nice to add support for `cloudpickle` (and fix the original support for `pickle` if it's been broken in some ways), contributions in the area of serializing/deserializing Param objects are welcomed :) !

samimia-swks:
Hi @maximlt,

I was able to figure out that cloudpickle (or dill) support was broken by https://github.com/holoviz/param/pull/386 which is between version 1.9.3 and 1.10.0. 

If I remove the offending lines of code, which are the definition __setstate__() method for the Parameters class, I can get cloudpickle to work even on the latest 1.13.0 version :
```python 
    def __setstate__(self, state):
        # Set old parameters state on Parameterized._parameters_state
        self_or_cls = state.get('self', state.get('cls'))
        for k in self_or_cls._parameters_state:
            key = '_'+k
            if key in state:
                self_or_cls._parameters_state[k] = state.pop(key)
        for k, v in state.items():
            setattr(self, k, v)
[...]

Issue #757, opened 2023-05-31
Pickle Serialization doesn't work for param.Parameterized with @param.depends function decorators


steenha:
VSCode, Jupyter Notebook Extension
params==1.13.0
python==3.11.3
OS=Ubuntu 22.04 (Windows Subsystem for Linux)

#### Description of expected behavior and the observed behavior
Normal param.Parameterized classes can be pickled just fine, but when adding the @param.depends function decorator to a method, the class can no longer be serialized with pickle

#### Complete, minimal, self-contained example code that reproduces the issue
```
import param
from param import Integer
import pickle

class package_nodepends(param.Parameterized):
    a = Integer(1)
    b = Integer(1)

class package_depends(param.Parameterized):
    a = Integer(1)
    b = Integer(1)
    c = Integer(1, doc="DERIVED: c=a+b")

    @param.depends("a", "b", watch=True, on_init=True) 
    def update_c(self): self.c = self.a+self.b

p_nodep = package_nodepends()
p_dep = package_depends()

with open("p_nodep.pkl", "wb") as pkl:
    pickle.dump(p_nodep, pkl)  # works fine

with open("p_dep.pkl", "wb") as pkl:
    pickle.dump(p_dep, pkl)  # fails here
```

#### Stack traceback and/or browser JavaScript console output
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[5], line 24
     21     pickle.dump(p_nodep, pkl)
     23 with open("p_dep.pkl", "wb") as pkl:
---> 24     pickle.dump(p_dep, pkl)

AttributeError: Can't pickle local object '_m_caller..caller'
```

samimia-swks:
What is interesting is that the current implementation of the depends() decorator, shown below, does use functools.wrap from the standard library; and that is supposed to solve the problem of pickeling decorated functions per [this article](https://towardsdatascience.com/why-you-should-wrap-decorators-in-python-5ac3676835f9)... 

```python
@accept_arguments
def depends(func, *dependencies, watch=False, on_init=False, **kw):
    if iscoroutinefunction(func):
        @wraps(func)
        async def _depends(*args, **kw):
            return await func(*args, **kw)
    else:
        @wraps(func)
        def _depends(*args, **kw):
            return func(*args, **kw)
```

maximlt:
Hi @steenha and @samimia-swks,

[...]

PR #783, opened 2023-07-03
Fix basic pickling

maximlt:
Fixes https://github.com/holoviz/param/issues/759
Fixes https://github.com/holoviz/param/issues/757

This PR also adds some basic pickling tests, some that we failing before the changes made. I wouldn't be surprised if there was more pickling functionality broken, adding more pickling tests is out of scope for this PR.


maximlt:
Merging since the tests pass and I want to merge this into the private namespace PR for it to benefit from more pickling tests.

Issue #785, opened 2023-07-04
param.pprint breaks if HoloViews is imported


Hoxbro:
Running the following: 
``` python
import holoviews as hv
import param


class T(param.Parameterized):
    a = param.Number(1)


t = T()
t.param.pprint()
```

Will give the following traceback:
``` python-traceback
Traceback (most recent call last):
  File "/home/shh/Development/holoviz/development/tmp.py", line 10, in <module>
    t.param.pprint()
  File "/home/shh/Repos/holoviz/param/param/parameterized.py", line 2894, in pprint
    return partial(self_._pprint, self, imports=imports, prefix=prefix,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/shh/Repos/holoviz/param/param/_utils.py", line 105, in wrapper
    result = user_function(self, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/shh/Repos/holoviz/param/param/parameterized.py", line 2914, in _pprint
    args = spec.args[1:] if spec.args[0] == 'self' else spec.args
                            ~~~~~~~~~^^^
IndexError: list index out of range
```

If I add `__init__` to the class, it works or remove the holoviews import it works again.

``` python
class T(param.Parameterized):
    a = param.Number(1)
    
    def __init__(self):
        super().__init__()
```

maximlt:
It can be reproduced with:

```python
import inspect
import param


class A(param.Parameterized): pass

A.__init__.__signature__ = inspect.Signature([inspect.Parameter('test', inspect.Parameter.KEYWORD_ONLY)])


class T(param.Parameterized):
    a = param.Number(1)


t = T()
print(t.param.pprint())
```

This is indeed what HoloViews is doing:
https://github.com/holoviz/holoviews/blob/a64cfc544acc94b3806d427912047ca660253587/holoviews/util/__init__.py#L626

But I don't understand yet why updating the docstring of a class (`A`) affects another one (`T`).

Hoxbro:
Isn't it because A's `__init__` is `param.parameterized.Parameterized.__init__`
![image](https://github.com/holoviz/param/assets/19758978/1068c8d1-af99-4c54-b17b-4289f572b8de)

maximlt:
Yep indeed I also just figured this out. So I think there's two workaround?

[...]

PR #786, opened 2023-07-04
Validate that `'self'` is present in the init signature of a Parameterized class

maximlt:
Fixes https://github.com/holoviz/param/issues/785

### DIFF
--- a/tests/testparameterizedrepr.py
+++ b/tests/testparameterizedrepr.py
@@ -230,8 +230,8 @@ def test_pprint_signature_overriden():
 
     t = T()
 
-    # This is actually setting the signature of param.Parameterized.__init__
-    # as P doesn't define __init__
+    # This is actually setting the signature of P.__init__
+    # as T doesn't define __init__
 
     # bad
     T.__init__.__signature__ = inspect.Signature(
--- a/tests/testparameterizedrepr.py
+++ b/tests/testparameterizedrepr.py
@@ -1,6 +1,7 @@
 """
 Unit test for the repr and pprint of parameterized objects, and for pprint/script_repr.
 """
+import inspect
 import unittest
 
 import param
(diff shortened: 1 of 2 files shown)


tests/testparameterizedrepr.py
```diff
@@ -217,3 +218,38 @@ def test_script_repr_parameterized_instance(P):
...
+        # as P doesn't define __init__
+
+        # bad
+        P.__init__.__signature__ = inspect.Signature(
+            [
+                inspect.Parameter('test', inspect.Parameter.KEYWORD_ONLY),
+            ]
+        )
+
+        with pytest.raises(KeyError, match="'T.__init__.__signature__' must contain a 'self' Parameter."):
```
Hoxbro, line 241: I would escape `.` with `\.`

Hoxbro (commented):
I would maybe make a subclass with a custom `__init__` and use that as a base to not mess around with `param.Parameterized` directly.

Maybe also have an example with a parameter in it with a non-default (If it does not already exist.)

This still needs a change in Holoviews, correct?

maximlt:
I addressed your comments.

> Maybe also have an example with a parameter in it with a non-default (If it does not already exist.)

`testparameterizedrepr` has a number of `.param.pprint` tests, including testing classes with an `__init__` that has positional only arguments.

tests/testparameterizedrepr.py
```diff
@@ -217,3 +218,39 @@ def test_script_repr_parameterized_instance(P):
...
+
+    class P(param.Parameterized):
+        def __init__(self, **params): pass
+
+    class T(P): pass
+
+    t = T()
+
+    # This is actually setting the signature of param.Parameterized.__init__
+    # as P doesn't define __init__
```
Hoxbro, line 234:
```suggestion
    # This is actually setting the signature of P.__init__
    # as T doesn't define __init__
```

Hoxbro (approved):

PR #787, opened 2023-07-11
fix minor errors in getting started page

droumis:
- the class in a code block should be `B` not `ParamClass`
- link to the Param guide on Panel docs was fixed


maximlt:
LGTM!

PR #788, opened 2023-07-11
Add basic benchmark suite

maximlt:
Fixes https://github.com/holoviz/param/issues/165

This PR adds a limited benchmark suite that leverages `asv`. I ran it on my machine with Python 3.11, from commits dating back to v1.11.1 (released in July 2021) to now.

image

Two PRs have recently degraded the performance (of that limited benchmark):
- https://github.com/holoviz/param/pull/605
- https://github.com/holoviz/param/pull/742

I think before releasing Param 2.0 it would be good to check whether this is acceptable.

---

Of course, more benchmarks could be added in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,7 @@ coverage.xml
 
 # Versionning
 param/_version.py
+
+# asv benchmark
+benchmarks/.asv
+benchmarks/param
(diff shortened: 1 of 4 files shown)


maximlt:
Merging as I'd like to use it to test another PR.

PR #790, opened 2023-07-12
Fix issues due to the Parameterized private namespace

maximlt:
https://github.com/holoviz/param/pull/766 was merged too quickly, running the HoloViews test suite locally revealed a bunch of new errors:

- setting an attribute in the `__init__` **before** calling `super().__init__` wasn't exerted in Param's test suite but leveraged in HoloViews (https://github.com/holoviz/holoviews/blob/a64cfc544acc94b3806d427912047ca660253587/holoviews/plotting/plot.py#L1168). This is now supported again and tested.
- the private namespace broke pickling when the protocol `0` is used which is apparently used by HoloViews in some cases. This PR fixes the support for that protocol, at least in the sense that the HoloViews unit test suite should now pass. To complete that fix, I have had to remove the compatibility code on `Watcher` that allowed to exclude `precedence` while unpacking (added in https://github.com/holoviz/param/pull/557). I believe that compatibility code is no longer required for Panel. The code in Param was released in 1.12.0 and Panel 0.13.0, which depended on Param >= 1.12.0, was released with updated code that no longer required the compatibility code (https://github.com/holoviz/panel/pull/2809). I had to remove that code as, I think, `__iter__` was called both when pickling and unpickling, resulting in the last two attributes being filtered out (error I got: `TypeError: __new__() missing 1 required positional argument: 'queued'`). I would hope no other Param dependents relied on that compatibility code like Panel used to do, and if they did I don't feel too bad breaking that for Param 2.0.
---

Working on these fixes I have stumbled upon this PR https://github.com/holoviz/param/pull/400 which had no tests, I have added some (thinking about another WIP PR that touches constant/instantiate too)

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -845,15 +845,6 @@ class Watcher(_Watcher):
             values['precedence'] = 0
         return super().__new__(cls_, **values)
 
-    def __iter__(self):
-        """
-        Backward compatibility layer to allow tuple unpacking without
-        the precedence value. Important for Panel which creates a
-        custom Watcher and uses tuple unpacking. Will be dropped in
-        Param 3.x.
-        """
-        return iter(self[:-1])
-
     def __str__(self):
         cls = type(self)
         attrs = ', '.join([f'{f}={getattr(self, f)!r}' for f in cls._fields])
@@ -3649,6 +3640,13 @@ class _ClassPrivate:
         self.params = {} if params is None else params
         self.values = {} if values is None else params
 
+    def __getstate__(self):
+        return {slot: getattr(self, slot) for slot in self.__slots__}
+
+    def __setstate__(self, state):
+        for k, v in state.items():
+            setattr(self, k, v)
+
 
 class _InstancePrivate:
     """
@@ -3699,6 +3697,13 @@ class _InstancePrivate:
         self.params = {} if params is None else params
         # self.watchers = {} if watchers is None else watchers
         self.values = {} if values is None else values
+
+    def __getstate__(self):
+        return {slot: getattr(self, slot) for slot in self.__slots__}
+
+    def __setstate__(self, state):
+        for k, v in state.items():
+            setattr(self, k, v)
 
 
 class Parameterized(metaclass=ParameterizedMetaclass):
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -3615,6 +3615,9 @@ class _ClassPrivate:
         Whethe the class has been renamed by a super class
     params: dict
         Dict of parameter_name:parameter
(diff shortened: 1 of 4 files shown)


param/parameterized.py
```diff
@@ -845,15 +845,6 @@ def __new__(cls_, *args, **kwargs):
             values['precedence'] = 0
         return super().__new__(cls_, **values)
 
-    def __iter__(self):
```
maximlt, line 848: @philippjfr are you fine with this removal? I needed to do that to fix pickling as explained in the post, or let's say that was the easier way to fix pickling I found.

philippjfr: Hmm, did you try running the Panel test suite with this?

philippjfr: Here's the Panel code that I think will break:

```python
        link_args = tuple(cb)
        # Compatibility with Param versions where precedence is dropped
        # from iterator for backward compatibility with older Panel versions
        if 'precedence' in Watcher._fields and len(link_args) < len(Watcher._fields):
            link_args += (cb.precedence,)
        link = LinkWatcher(*(link_args+(target, links, callbacks is not None, bidirectional_watcher)))
        self._links.append(link)
```

param/parameterized.py
```diff
@@ -3744,7 +3755,14 @@ class Foo(Parameterized):
     def __init__(self, **params):
         global object_count
 
-        self._param__private = _InstancePrivate()
+        # Setting a Parameter in an __init__ block before calling super().__init__
```
philippjfr, line 3758: Seems scary, what happens for a admittedly contrived case like this:

```python
class A(param.Parameterized):

    value = param.String(default="A")

    def __init__(self, depth=0):
       self.value = 'B'
        if depth < 2:
            self.sub = A()
        super().__init__()
```

param/parameterized.py
maximlt: > Hmm, did you try running the Panel test suite with this?

Yes the tests ran fine with this branch checked out.

That code seemed to deal with backward compatibility with older Panel versions. Given that they pin `param < 2.0`, aren't we safe? With the change made in this PR, `len(link_args) < len(Watcher._fields)` is `False`.

param/parameterized.py
maximlt: Ok indeed that's plain wrong!

param/parameterized.py
philippjfr: Ah, missed that.

maximlt:
This is again ready for review @philippjfr

philippjfr (approved):
Makes good sense and thanks for all the tests!

Issue #693, opened 2023-03-06
Default value not in `.objects` when `check_on_set=False`


maximlt:
The [docs on Selectors](https://param.holoviz.org/user_guide/Parameter_Types.html?highlight=check_on_set#selectors) say:

> If the list of available objects is not meant be exhaustive, you can specify `check_on_set=False` (which automatically applies if the initial list is empty). Objects will then be added to the objects list whenever they are set, **including as the initial default.**

However it's not quite true, I think, as I'd expect `p.param.s.objects` to contain the default value?

```python
import param

class P(param.Parameterized):
    s = param.Selector(default=3, objects=[1, 2], check_on_set=False)

p = P()

print(p.s, p.param.s.objects)
# 3 [1, 2]
```

jbednar:
Good catch; I think 3 should be put into the objects list in the above case.

PR #794, opened 2023-07-20
Populate `objects` when `check_on_set` is False and `default` is not in `objects`

maximlt:
Fixes https://github.com/holoviz/param/issues/693

It's been pretty tricky to implement as `Selector` isn't the easiest Parameter to deal with (its setup is quite dynamic) and handling slot values inheritance for it adds an extra level of complexity. The approach I've opted for is to add a method to `Parameter` that subclasses can implement and that is called only after slot values inheritance has completed, to update the Parameter state. In that case, it's used to add the `default` value to `objects` when `check_on_set` is False and `default` is not in there.

 I haven't managed to get that to work together with the unbound case so I marked that test as xfailed https://github.com/holoviz/param/compare/fix_checkonset_default?expand=1#diff-df60be25aab735646b74b0a737306d420bec4718b0dadd8323fe8e5c31e67995R259-R267

Overall, I'm not quite sure how I feel about this PR. `Selector` has quite a number of issues and it's always easier to add some new behavior rather than removing it. Also, it's easy to miss edge cases with the various combinations of Selector attributes when in an inheritance context. I tend to think that the status-quo is probably a better option, I'm interested in feedback anyway!

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1695,6 +1695,15 @@ class Selector(SelectorBase, _SignatureSelector):
             self.allow_None = allow_None
         if self.default is not None and self.check_on_set is True:
             self._validate(self.default)
+        self._update_state()
+
+    def _update_state(self):
+        if (
+            self.check_on_set is False
+            and self.default is not None
+            and self.default not in self.objects
+        ):
+            self.objects.append(self.default)
 
     @property
     def objects(self):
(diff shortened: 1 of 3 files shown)


jbednar (approved):
Looks right to me. I was just playing around with code related to a different PR and noticed this incorrect behavior, so I do think this should be merged.

Issue #606, opened 2022-02-13
Parameterized.param.pprint() hangs when param.Array parameter is present


peterroelants:
#### Description of expected behavior and the observed behavior

Having a [`param.Array`](https://param.holoviz.org/reference.html?highlight=array#param.__init__.Array) in  a `Parameterized` class results in [`Parameterized.param.pprint()`](https://param.holoviz.org/user_guide/Parameters.html#displaying-parameterized-objects) hanging.

#### Complete, minimal, self-contained example code that reproduces the issue

Running:
```python
import param
import panel as pn
import numpy as np
pn.extension()


class MatParam(param.Parameterized):
    mat = param.Array(np.zeros((2, 2)))

              
mp = MatParam()
p = pn.Column(mp.param.mat)
display(p)

print("Before pprint.")
print(mp.param.pprint())
print("After pprint (Does not print this!)")
```
hangs, and does not print the `"After pprint (Does not print this!)"` line.

Screenshot of the full output:
<img src="https://user-images.githubusercontent.com/951093/153769225-f4911830-ead1-48a4-9cd7-822e458cc154.png" width="400"/>

#### ALL software version info

- Panel 0.12.6
- Param 1.12.0
- Python 3.9.10
- jupyterlab : 3.2.8

jbednar:
Thanks! I can reproduce this without panel, where it gives an error message rather than hanging:

```python
import param, numpy as np

class MatParam(param.Parameterized):
    mat = param.Array(np.zeros((2, 2)))

mp = MatParam()

print("Before pprint.")
print(mp.param.pprint())
print("After pprint (Does not print this!)")
```
```python
Before pprint.
 ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all() 
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-1-1ab328ac4e59> in <module>
      8 
      9 print("Before pprint.")
---> 10 print(mp.param.pprint())
     11 print("After pprint (Does not print this!)")

~/param/param/parameterized.py in pprint(self_, imports, prefix, unknown_value, qualify, separator)
   2597         """See Parameterized.pprint"""
   2598         self = self_.self
-> 2599         return self._pprint(imports, prefix, unknown_value, qualify, separator)
   2600 
   2601 

~/param/param/parameterized.py in wrapper(self, *args, **kwargs)
   3089                 repr_running.add(key)
   3090                 try:
[...]

PR #795, opened 2023-07-21
Fix `param.pprint` for `Array` parameters by replacing `all_equal` with `Comparator.is_equal` in `values()`

maximlt:
Fixes https://github.com/holoviz/param/issues/606
Fixes https://github.com/holoviz/param/issues/179
Supersedes https://github.com/holoviz/param/pull/437

I chose to replace `all_equals` in `.param.values` by `Comparator.is_equal`, which then needed to be extended to deal with objects that have the `_infinitely_iterable` attribute (`Time`). I've had to change the code in `is_equal` a little to fix the support for `predicate: equality`.

As a reminder (https://github.com/holoviz/param/issues/179#issuecomment-362818251), `all_equal` has this weird behavior that `all_equal('aaaa', 'aa')` is True. That behavior is relied upon (intentionally I guess) by `.param.values(onlychanged=True)` so that the Parameterized instance auto-generated name is filtered out of the returned values. I copied code from https://github.com/holoviz/param/pull/437 to make sure that this behavior is preserved. Because it's no longer used, I have deprecated `all_equals`.

https://github.com/holoviz/param/pull/437 was discussing a bit more some API to decide whether you want to auto-suppress `name` or not. I have decided not to deal with that in this PR that is focused on fixing the linked issues.

https://github.com/holoviz/param/pull/437 also discussed how to deal with comparing numpy arrays. As indeed if `.param.pprint()` now works with `Array` Parameters it will always include the default array, as the comparator doesn't handle this object type.  I actually started to work on that and later realized this is a can of worms :)

### DIFF
--- a/tests/testparameterizedobject.py
+++ b/tests/testparameterizedobject.py
@@ -391,7 +391,7 @@ class TestParameterized(unittest.TestCase):
         assert 'inst' in TestPO.param.values()
         assert 'notinst' in TestPO.param.values()
 
-    def test_values_name_ignored_for_intances_and_onlychanged(self):
+    def test_values_name_ignored_for_instances_and_onlychanged(self):
         default_inst = param.Parameterized()
         assert 'Parameterized' in default_inst.name
         # name ignored when automatically computed (behavior inherited from all_equal)
--- a/tests/testparameterizedobject.py
+++ b/tests/testparameterizedobject.py
@@ -391,6 +391,14 @@ class TestParameterized(unittest.TestCase):
         assert 'inst' in TestPO.param.values()
         assert 'notinst' in TestPO.param.values()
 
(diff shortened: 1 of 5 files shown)


tests/testparameterizedobject.py
```diff
@@ -391,6 +391,14 @@ def test_values(self):
         assert 'inst' in TestPO.param.values()
         assert 'notinst' in TestPO.param.values()
 
+    def test_values_name_ignored_for_intances_and_onlychanged(self):
```
jbednar, line 394:
```suggestion
    def test_values_name_ignored_for_instances_and_onlychanged(self):
```

jbednar (approved):
Looks good to me, thanks!

philippjfr (approved):

Issue #780, opened 2023-06-30
Add API to access watchers


maximlt:
Following @philippjfr's comment in https://github.com/holoviz/param/pull/766#issuecomment-1614500690, we have identified the need to provide an API to access watchers.

More precisely, it seems like we do not have an API to access watchers on an instance, which are currently available through the private attribute `_param_watchers`:
```python
import param

class P(param.Parameterized):
    x = param.Parameter()
    
    @param.depends('x', watch=True)
    def d(self): pass

p = P()
print(p._param_watchers)
# {'x': {'value': [Watcher(inst=P(name='P00005', x=None), cls=<class '__main__.P'>, fn=<function _m_caller.<locals>.caller at 0x118fa5c10>, mode='args', onlychanged=True, parameter_names=('x',), what='value', queued=False, precedence=-1)]}}
print(p.param['x'].watchers)
# {}
```

There is a sort of public API to access the watchers set on class Parameters:
```python
import param

class P(param.Parameterized):
    x = param.Parameter()

P.param.watch(lambda e: print(e), 'x')
print(P.param['x'].watchers)
# {'value': [Watcher(inst=None, cls=<class '__main__.P'>, fn=<function <lambda> at 0x103f3b1f0>, mode='args', onlychanged=True, parameter_names=('x',), what='value', queued=False, precedence=0)]}
```

There is the public `.param.watchers` that under the hood returns the watchers queue defined on `_parameters_state`, that holds transient data:
```python
import param

class P(param.Parameterized):
    x = param.Parameter()
    
    @param.depends('x', watch=True)
    def d(self): pass

p = P()
p._param_watchers

print('start', p.param.watchers)
with param.parameterized.batch_call_watchers(p):
    print('   start cm', p.param.watchers)
    p.x = 1
    p.x = 2
    p.x = 3
    print('   end cm', p.param.watchers)
print('end', p.param.watchers)
```
Output:
```
start []
   start cm []
   end cm [Watcher(inst=P(name='P00005', x=3), cls=<class '__main__.P'>, fn=<function _m_caller.<locals>.caller at 0x118fa5c10>, mode='args', onlychanged=True, parameter_names=('x',), what='value', queued=False, precedence=-1)]
end []
```

---

[...]

PR #797, opened 2023-07-25
Deprecate `._param_watchers` in favor of `.param.watchers`

maximlt:
Closes https://github.com/holoviz/param/issues/780

- `.param.watchers` no longer return the transient dict of watchers (breaking change)
- instead it returns what used to be returned by `_param_watchers`
- `_param_watchers` has been turned into a deprecated property

I just ran the unit test suite of Panel with this branch checked out without any problem.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2556,6 +2556,8 @@ class Parameters:
         outputs = {}
         for cls in classlist(self_.cls):
             for name in dir(cls):
+                if name == '_param_watchers':
+                    continue
                 method = getattr(self_.self_or_cls, name)
                 dinfo = getattr(method, '_dinfo', {})
                 if 'outputs' not in dinfo:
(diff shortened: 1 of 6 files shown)


param/parameterized.py
```diff
@@ -1720,17 +1718,25 @@ def _events(self_, value):
...
 
     @property
-    def watchers(self):
-        """Read-only list of watchers on this Parameterized"""
-        return self._watchers
+    def watchers(self_):
+        """Dictionary of instance watchers."""
+        if self_.self is None:
+            raise TypeError()
+        return self_.self._param__private.watchers
```
philippjfr, line 1733: Briefly considered whether this should be a `FrozenDict` copy of the real dict but I think being able to modify the watchers is kind of the point.

philippjfr (approved):
Very happy to see this merged. Thank you!

param/parameterized.py
maximlt: Yep indeed I thought you told me that you were manipulating the watchers in Panel. Also if I'm right there's no frozendict implementation in the standard library.

Good anyway that you commented right there, the `TypeError` raised in line 1732 is a little dry :) I'm going to merge after fixing that.

Issue #760, opened 2023-06-11
Subclassing a Parameterized class and changing the class parameters of list / dict does not work


Hoxbro:
#### Description of expected behavior and the observed behavior

When subclassing a Parameterized class and changing the class parameter of the subclass, it does not behave well when the value is dict or a list.

A git bisect shows #424 seems to be the culprit. 

First discovered here: https://discord.com/channels/1075331058024861767/1116349791505494106

#### Complete, minimal, self-contained example code that reproduces the issue

``` python
import param

class MyParam(param.Parameterized):
    d = param.Dict(default={})
    i = param.Integer(default=0)

class MySubParam(MyParam):
    pass

MySubParam.d = {'a': 1}
MySubParam.i = 2

p = MySubParam()
print(p.d, p.i)

```
![image](https://github.com/holoviz/param/assets/19758978/225e913a-f6d6-4e8c-8f84-b0978b54bb5d)

PR #798, opened 2023-07-25
Materialize default values based on all the parameters available

maximlt:
Fixes https://github.com/holoviz/param/issues/760

https://github.com/holoviz/param/issues/760 identified that this regression happened in https://github.com/holoviz/param/pull/424. More precisely I've found out it was due to this change https://github.com/holoviz/param/pull/424/files#diff-c8fc4fb68012b66fbf42ad9e0b4714fd96a3a72b83558bc543130d78bd622eb3R1275-R1279 by which Param would  "instantiate" (i.e. materialize a default Parameter value into an instance) the wrong Parameter default. That happened if a Parameter is inherited and isn't re-declared by the subclass, in which case the "instantiated" value would be the one from the first class above in the hierarchy that declares the Parameter.

Instead of the custom class hierarchy traversal I've opted for leveraging `.param.objects(instance=False)`.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1837,19 +1837,13 @@ class Parameters:
         """
         self = self_.param.self
         ## Deepcopy all 'instantiate=True' parameters
-        # (building a set of names first to avoid redundantly
-        # instantiating a later-overridden parent class's parameter)
         params_to_deepcopy = {}
         params_to_ref = {}
-        for class_ in classlist(type(self)):
-            if not issubclass(class_, Parameterized):
-                continue
-            for (k, v) in class_.param._parameters.items():
-                # (avoid replacing name with the default of None)
-                if v.instantiate and k != "name":
-                    params_to_deepcopy[k] = v
-                elif v.constant and k != 'name':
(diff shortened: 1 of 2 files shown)


param/parameterized.py
```diff
@@ -3064,8 +3058,6 @@ def __init__(mcs, name, bases, dict_):
         parameters = [(n, o) for (n, o) in dict_.items()
                       if isinstance(o, Parameter)]
 
-        mcs._param__parameters._parameters = dict(parameters)
```
philippjfr, line 3067: Can yo explain why this was needed and is no longer needed?

maximlt: Sure I'll try. It was only used here https://github.com/holoviz/param/pull/798/files#diff-c8fc4fb68012b66fbf42ad9e0b4714fd96a3a72b83558bc543130d78bd622eb3L1847 and as far as I can see it was just used for performance reasons (added in https://github.com/holoviz/param/pull/424) to avoid looping through the whole class `__dict__` and just through the class Parameters.

philippjfr (approved):
Looks good!

PR #801, opened 2023-07-27
`FileSelector` defaults to the first globbed path

maximlt:
Addresses the first issue raised https://github.com/holoviz/param/issues/545, i.e. that the `FileSelector` parameter is not updating its default value to the first path found in `objects`, which is inconsistent with how `Selector` works.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -2462,20 +2462,23 @@ class FileSelector(Selector):
     def __init__(self, default=Undefined, *, path="", **kwargs):
         self.default = default
         self.path = path
-        self.update()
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good to me!

Right now it looks like the glob is run when the parameter is declared which is probably what I would expect. Unlike `Selector`, the set of 'objects' is implicit so I think it might be worth stating in the docstring that the glob is run at the point where the parameter is declared.

maximlt:
Thanks for the review! I'll update the docstring in another PR.

Issue #371, opened 2020-01-18
Use __signature__ to generate Parameterized signatures


philippjfr:
Currently we use a docstring trick to generate signatures for Parameterized classes. This approach is no longer supported by IPython in recent versions and is also pretty hacky (and ugly to look at in a docstring since all the keywords appear there). Instead we should use the ``__signature__`` object (see [PEP362](https://www.python.org/dev/peps/pep-0362/)).

![Screen Shot 2020-01-18 at 9 51 33 PM](https://user-images.githubusercontent.com/1550771/72670351-b8236e80-3a3c-11ea-87a9-1077571a698a.png)

philippjfr:
Since this may not be desirable for all Parameterized classes another suggestion might be that you can turn this one for certain Parameter classes by setting some flag on the class, e.g. `Parameterized._expand_signature = True`.

jlstevens:
The above screenshot is based on a holoviews example (``opts``) but there is an issue when generalized to all the things param supports. The problem is that ``__signature__`` is used for inspection (as described in the PEP above) so it will be relied on by anything doing inspection (as opposed to just being helpful for tab-completion).

The issue then is that we would like to support tab-completion for parameterized classes (i.e constructors) as well as for parameterized functions. The ```__signature__``` should then reflect reality i.e including any positional arguments, `*args` and ``**kwargs`` that don't correspond to parameters.

The problem with say expanding ``**kwargs`` into a list of explicit parameter keywords followed by ``**kwargs`` is that this makes assumptions about how param is being used:

* The ``**kwargs`` in the constructor may be filtered/processed before mapping them to the parameters. For instance, nothing stops you from declaring a class with parameter `foo` that does not accept `foo` as a keyword in the constructor.

* You would always need to allow for ``**kwargs`` even when the entire set of keywords *are* parameters - you have no way of knowing whether kwargs that aren't parameters are being used or not.

* Things can get even messier with parameterized functions (i.e many instance of ``__call__`` won't use ``ParamOverrides``).

While I do think the docstring approach was ugly and hacky, it had the nice property of only really being a hint for tab-completion. Setting ``__signature__`` is a more serious affair (assuming you don't want to mess around with the results of inspection) so while I prefer this approach, I think it is important to specify signatures that don't deviate too much from what is actually accepted by the constructors/``__call__`` methods.

jlstevens:
[...]

PR #802, opened 2023-07-29
Remove inferred docstring signature and add basic __signature__ support

maximlt:
Fixes https://github.com/holoviz/param/issues/371 by:

1. removing the automatically inferred *docstring signature* (e.g. `params(x=Number, name=String)`) as IPython would no longer leverage it for parameter autocompletion (as noted in the issue) and because it was wrong when the class was having an `__init__` method with a signature different than `__init__(self, **params)`
2. implementing `Parameterized.__signature__` as a class property that returns a `Signature` built from the class Parameters, **only** when the class (or its superclasses) doesn't implement an `__init__` method that is different than the default one `__init__(self, **params)`.

image

Somewhat surprisingly 2) seemed to have broken only a few tests in Panel and none in HoloViews. The two unit tests that broke in Panel are https://github.com/holoviz/panel/blob/df0b277dca2fd54dd2ae016159adc5c33b61e358/panel/tests/widgets/test_base.py#L25 and https://github.com/holoviz/panel/blob/df0b277dca2fd54dd2ae016159adc5c33b61e358/panel/tests/test_viewable.py#L40, I believe they test that the `__signature__` of some Panel objects is actually *not* modified, as there is some code in Panel that modifies the signature of components. That code in Panel by the way doesn't seem to be tested, given https://github.com/holoviz/panel/blob/df0b277dca2fd54dd2ae016159adc5c33b61e358/panel/tests/conftest.py#L32. Anyway, it seems to me the change done in this PR would not affect Panel, let's double check that however if it's not tested.

I have chosen not to even try to handle custom `__init__` in 2) as I'm not even sure it's possible (@jlstevens raised some good points in the issue) and I believe that a better approach for Param in the future would be to add Parameter attributes to control how the Parameter appears in the signature (e.g. positional, hidden) and provide a post init hook so that users don't have to create an `__init__` in most cases.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -3171,6 +3171,42 @@ class ParameterizedMetaclass(type):
         except AttributeError:
             return False
 
+    def __get_signature(mcs):
+        """
+        When a class doesn't implement an __init__ method with a signature
+        different than Parameterized.__init__, calling its `__signature__`
+        property will return a custom signature built by this method, None
+        otherwise.
+        """
+        # allowed_signature must be the signature of Parameterized.__init__
+        allowed_signature = inspect.Signature([
+            inspect.Parameter('self', inspect.Parameter.POSITIONAL_OR_KEYWORD),
+            inspect.Parameter('params', inspect.Parameter.VAR_KEYWORD),
+        ])
+        # Inspecting `mcs.__init__` instead of `mcs` to avoid a recursion error
+        if inspect.signature(mcs.__init__).parameters != allowed_signature.parameters:
+            return
+        processed_kws, keyword_groups = set(), []
+        for cls in reversed(mcs.mro()):
+            keyword_group = []
+            for k, v in sorted(cls.__dict__.items()):
+                if isinstance(v, Parameter) and k not in processed_kws and not v.readonly:
+                    keyword_group.append(k)
+                    processed_kws.add(k)
+            keyword_groups.append(keyword_group)
+
+        keywords = [el for grp in reversed(keyword_groups) for el in grp]
+        # if mcs.name == 'C':
+        #     import pdb; pdb.set_trace()
+        # keywords = [el for grp in keyword_groups for el in grp]
+        signature = inspect.Signature([
+            inspect.Parameter(k, inspect.Parameter.KEYWORD_ONLY)
+            for k in keywords
+        ])
+        return signature
+
+    __signature__ = property(__get_signature)
+
     abstract = property(__is_abstract)
 
     def _get_param(mcs):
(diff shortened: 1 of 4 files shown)


param/parameterized.py
```diff
@@ -3129,30 +3129,16 @@ def __set_name(mcs, name, dict_):
...
-
-        keywords = [el for grp in reversed(keyword_groups) for el in grp]
-        class_docstr = "\n"+mcs.__doc__ if mcs.__doc__ else ''
-        signature = "params(%s)" % (", ".join(keywords))
-        description = param_pager(mcs) if (docstring_describe_params and param_pager) else ''
-        mcs.__doc__ = signature + class_docstr + '\n' + description
+        if not docstring_describe_params or not param_pager:
+            return
+        class_docstr = mcs.__doc__ if mcs.__doc__ else ''
+        description = param_pager(mcs)
```
philippjfr, line 3140: I actually really dislike having the pager in the docstring too, but that's a separate discussion I guess.

param/parameterized.py
```diff
@@ -3185,6 +3171,42 @@ def __is_abstract(mcs):
...
+            for k, v in sorted(cls.__dict__.items()):
+                if isinstance(v, Parameter) and k not in processed_kws and not v.readonly:
+                    keyword_group.append(k)
+                    processed_kws.add(k)
+            keyword_groups.append(keyword_group)
+
+        keywords = [el for grp in reversed(keyword_groups) for el in grp]
+        # if mcs.name == 'C':
+        #     import pdb; pdb.set_trace()
+        # keywords = [el for grp in keyword_groups for el in grp]
```
maximlt, line 3201:
```suggestion
```

philippjfr:
Made some minor changes:

- Pull out the default signature into a global variable
- Cache the generated signature on `._param__private`

tests/testparameterizedobject.py
```diff
@@ -1411,3 +1412,143 @@ def __init__(self, **params):
             super().__init__(**params)
 
     P()
+
+
+def test_parameterized_docstring():
+    # TODO
+    pass
+
```
philippjfr, line 1420:
```suggestion
def test_parameterized_docstring():
    # TODO
    pass

```

philippjfr (approved):
Looks good and I'm happy with this conservative approach to auto-generating signatures!

param/parameterized.py
jbednar: Same -- I want to control the pager myself, using Emacs in my case, and am annoyed when any single program tries to handle paging.

Issue #566, opened 2021-10-29
No error raised when a method depends on Parameterized instances set after calling super()


maximlt:
See this example taken from Panel's user guide. The two objects must be declared in the `__init__` before calling `super()`.

```python
class P2(param.Parameterized):
    value = param.String()

class P1(param.Parameterized):
    
    def __init__(self, **params):
        # Works as expected when declared before
        # self.p2 = P2()
        super().__init__(**params)
        # Does not work when declared after super()
        # Raises an error in param<1.12, nothing for param 1.12
        self.p2 = P2()

    @param.depends('p2.value', watch=True)
    def cb(self):
        print('fired')
```

When they're declared after as in the example above, the following error used to be raised (this is param 1.11.1):

```python
--------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-36-7e8d061f6e7f> in <module>
----> 1 p1 = P1()

<ipython-input-35-d76ac1c5846c> in __init__(self, **params)
      6     def __init__(self, **params):
      7 #         self.p2 = P2()
----> 8         super().__init__(**params)
      9         self.p2 = P2()
     10 

~/miniconda3/envs/holoviz37/lib/python3.7/site-packages/param/parameterized.py in __init__(self, **params)
   2610                 # 'dependers'.
   2611                 grouped = defaultdict(list)
-> 2612                 for dep in self.param.params_depended_on(n):
   2613                     grouped[(id(dep.inst),id(dep.cls),dep.what)].append(dep)
   2614                 for group in grouped.values():

~/miniconda3/envs/holoviz37/lib/python3.7/site-packages/param/parameterized.py in params_depended_on(self_, name)
   1858 
   1859     def params_depended_on(self_,name):
-> 1860         return _params_depended_on(MInfo(cls=self_.cls,inst=self_.self,name=name,method=getattr(self_.self_or_cls,name)))
   1861 
   1862 

~/miniconda3/envs/holoviz37/lib/python3.7/site-packages/param/parameterized.py in _params_depended_on(minfo)
    500     dinfo = getattr(minfo.method,"_dinfo", {})
    501     for d in dinfo.get('dependencies', list(minfo.cls.param)):
--> 502         things = (minfo.inst or minfo.cls).param._spec_to_obj(d)
    503         for thing in things:
    504             if isinstance(thing,PInfo):

[...]

PR #813, opened 2023-08-02
Validate dependencies in constructor

philippjfr:
The dynamic resolution of dependencies loosened the validation for dependencies. However this was only intended to allow dynamic dependency resolution of the leafs or branches of nested dependencies, i.e. the root of a nested dependency has to be declared. This adds validation to ensure that the root dependency can be resolved and adds an error otherwise.

- [x] Fixes https://github.com/holoviz/param/issues/566
- [x] Add test(s)

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2607,6 +2607,14 @@ class Parameters:
         elif not dynamic:
             return [], [DInfo(spec=spec)]
         else:
+            if not hasattr(self_.self_or_cls, obj.split('.')[1]):
+                raise AttributeError(
+                    f'Dependency {obj[1:]!r} could not be resolved, {self_.self_or_cls} '
+                    f'has no parameter or attribute {obj.split(".")[1]!r}. Ensure '
+                    'the object being depended on is declared before calling the '
+                    'Parameterized constructor.'
+                )
+
             src = _getattrr(self_.self_or_cls, obj[1::], None)
             if src is None:
                 path = obj[1:].split('.')
--- a/tests/testparamdepends.py
+++ b/tests/testparamdepends.py
@@ -1300,3 +1300,22 @@ def test_misspelled_parameter_in_depends_watch():
             @param.depends("tlim", watch=True)  # <- Misspelled xlim
             def test(self):
                 return True
+
+def test_param_depends_on_undefined_attribute():
+    class P2(param.Parameterized):
+        value = param.String()
+
+    class P1(param.Parameterized):
+
+        def __init__(self, **params):
+            super().__init__(**params)
+            self.p2 = P2()
+
+        @param.depends('p2.value', watch=True)
+        def cb(self):
+            print('fired')
+
+    with pytest.raises(AttributeError) as excinfo:
+        P1()
+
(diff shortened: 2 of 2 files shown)


maximlt (approved):
I see that the error is not raised at class creation time but on instantiation and only when `watch=True`, it's a big improvement anyway.

philippjfr:
It cannot be done at class creation time because the class cannot know what happens in the constructor before super is called.

maximlt:
Is it because we allow people depending on objects that aren't declared as `Parameter` but are set in the `__init__`? If so, I'd be happy to see a warning in cases like that, it's a bit of an anti-pattern isn't it?

A recent change made this fail when `watch=False`:
```python
import param

class P(param.Parameterized):
    x = param.Parameter()
    
    @param.depends('y')
    def debug(self): pass
```

Traceback:
```
AttributeError                            Traceback (most recent call last)
Cell In[42], line 3
      1 import param
----> 3 class P(param.Parameterized):
      4     x = param.Parameter()
      6     @param.depends('y')

File ~/dev/param/param/parameterized.py:3097, in ParameterizedMetaclass.__init__(mcs, name, bases, dict_)
   3094 on_init = dinfo.get('on_init', False)
   3095 minfo = MInfo(cls=mcs, inst=None, name=name,
   3096               method=method)
-> 3097 deps, dynamic_deps = _params_depended_on(minfo, dynamic=False)
   3098 if watch:
   3099     _watch.append((name, watch == 'queued', on_init, deps, dynamic_deps))

File ~/dev/param/param/parameterized.py:649, in _params_depended_on(minfo, dynamic, intermediate)
    647 dinfo = getattr(minfo.method, "_dinfo", {})
    648 for d in dinfo.get('dependencies', list(minfo.cls.param)):
--> 649     ddeps, ddynamic_deps = (minfo.cls if minfo.inst is None else minfo.inst).param._spec_to_obj(d, dynamic, intermediate)
    650     dynamic_deps += ddynamic_deps
    651     for dep in ddeps:

File ~/dev/param/param/parameterized.py:2661, in Parameters._spec_to_obj(self_, spec, dynamic, intermediate)
   2659     return [], [] if intermediate == 'only' else [DInfo(spec=spec)]
   2660 else:
-> 2661     raise AttributeError(f"Attribute {attr!r} could not be resolved on {src}.")
   2663 if obj is None or not intermediate:
   2664     return [info], []

AttributeError: Attribute 'y' could not be resolved on <class '__main__.P'>.
```

philippjfr:
I'm a little on the fence on it and likely have to revert your change. I've seen many people use the pattern and I have also used it on occasion because declaring parameters is sometimes awkward, e.g. often I use parameters for public API declarations but internally I want to be able to set up dependencies on static attributes and not have to set up parameter declarations.

maximlt:
> I use parameters for public API declarations but internally I want to be able to set up dependencies on static attributes and not have to set up parameter declarations.

Yes that's a weakness of Param that has no way to tell whether a parameter is an input or some state you compute later on. Param has `@output` but I haven't seen that used much. `dataclasses.field` has an `init` that is `True` by default and that you can use to indicate whether the generated `init` should contain this field, I think Param should have something similar.

PR #814, opened 2023-08-03
Updating `path` on a `File/MultiFileSelector` Parameter updates `objects`

maximlt:
Fixes https://github.com/holoviz/param/issues/545
Supersedes https://github.com/holoviz/param/pull/546

The issue stated about `FileSelector`:

> FileSelector behaves like Selector and should as such set its value to the first item of objects, objects being the result of a glob on path.

This has been addressed in https://github.com/holoviz/param/pull/801.

About `MultiFileSelector`:

> Also the docs currently state that the default value is all of the matched files, but in practice it's None.

While that's what the docs say, the behavior of `ListSelector` from which `MultiFileSelector` inherits is for `default` to be `None` so I'm not sure we should make this change without also updating `ListSelector`. Also, Panel maps the `MultiFileSelector` parameter to the [FileSelector](https://panel.holoviz.org/reference/widgets/FileSelector.html) widget (the `FileSelector` **Parameter** being mapped to a simple `Select` or `TextInput` widget) and I don't think we'd want all the files to be selected by default.

```python
import param

class P(param.Parameterized):
    ls = param.ListSelector(objects=[1, 2, 3])

a = A()

assert a.ls is None
```

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -2486,6 +2486,10 @@ class FileSelector(Selector):
     """
     __slots__ = ['path']
 
+    _slot_defaults = _dict_update(
+        Selector._slot_defaults, path="",
+    )
+
     @typing.overload
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


philippjfr (approved):
Looks good to me!

PR #817, opened 2023-08-17
Listselectordefault

### LINKED ISSUE #807: `ListSelector`: regression when `check_on_set=False`
I think this was caused by https://github.com/holoviz/param/pull/794:

```python
import param

ls = param.ListSelector([1, 2], objects=[1, 2, 3], check_on_set=False)
ls.objects
# [1, 2, 3, [1, 2]]  :(
```

jbednar:
Fixes #807. Makes _update_state from #794 aware of list-type defaults so that it adds the items on the list to the objects, not the entire default value as a sublist. This change does seem to be successful, avoiding `objects=1, 2, 3, [1, 2]]`:

image

Regarding the discussion in #807:

> Actually I'm even thinking that https://github.com/holoviz/param/issues/693 is a silly usage of Param and I'm no longer sure we should "fix" it. That's the example I shared in https://github.com/holoviz/param/issues/693, why would you not include 3 in the objects?? On top of that check_on_set is pretty clearly worded with on_set and not on_init.

After looking into it I agree. We definitely shouldn't do any extra work to make this code legal:

```python
import param

class P(param.Parameterized):
    s = param.Selector(default=3, objects=[1, 2], check_on_set=False)

p = P()
```

If people wanted 3 on the objects list, they should have put it there, since they supplied an objects list.  This code is currently accepted by this PR, yielding `p.params.objects=[1, 2, 3]`, but if it weren't, I'd be equally happy.

Less clear is what should happen when no objects list is supplied:

image

In this case, it does seem appropriate to be adding the default value to the objects list, and supporting this case automatically makes the above less-important case work.

The way it works still depends on Maxime's _update_state approach, which I can now appreciate. I had previously suggested trying to eliminate _update_state, but the problem (of which I'm sure @maximlt is aware) is that using the Sentinel approach for slot values means that the _objects list doesn't get a concrete value until after the constructor completes and returns control to the metaclass. By that point it's too late for the constructor to add items to the list based on the default, hence needing a new _update_state method to fill it in later. 

So if we want to eliminate _update_state (which is a worthwhile goal), I think we'd have to require the objects list to include the default value, which seems like a big disruption to previous usage. If Panel developers are ok with that, sure, but otherwise I guess we are stuck with _update_state.

### DIFF
--- a/tests/testlistselector.py
+++ b/tests/testlistselector.py
@@ -18,7 +18,7 @@ class TestListSelectorParameters(unittest.TestCase):
         super().setUp()
         class P(param.Parameterized):
             e = param.ListSelector(default=[5],objects=[5,6,7])
-            f = param.ListSelector(default=10)
+            f = param.ListSelector(default=[10])
             h = param.ListSelector(default=None)
             g = param.ListSelector(default=None,objects=[7,8])
             i = param.ListSelector(default=[7],objects=[9],check_on_set=False)
@@ -173,16 +173,17 @@ class TestListSelectorParameters(unittest.TestCase):
             r = param.ListSelector(default=[6])
 
     ##########################
-    # CEBALERT: not sure it makes sense for ListSelector to be set to
-    # a non-iterable value (except None). I.e. I think this first test
-    # should fail.
     def test_default_not_checked_to_be_iterable(self):
-        class Q(param.Parameterized):
-            r = param.ListSelector(default=6)
+        with pytest.raises(
+            ValueError,
+            match=re.escape("ListSelector parameter 'r' only takes list types, not 6."),
+        ):
+            class Q(param.Parameterized):
+                r = param.ListSelector(default=6)
 
     def test_set_checked_to_be_iterable(self):
         class Q(param.Parameterized):
-            r = param.ListSelector(default=6,check_on_set=False)
+            r = param.ListSelector(default=[6],check_on_set=False)
 
         with self.assertRaises(ValueError):
             Q.r = 6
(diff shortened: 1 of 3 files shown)


tests/testlistselector.py
```diff
@@ -18,7 +18,7 @@ def setUp(self):
         super().setUp()
         class P(param.Parameterized):
             e = param.ListSelector(default=[5],objects=[5,6,7])
-            f = param.ListSelector(default=10)
+            f = param.ListSelector(default=[10])
```
jbednar, line 21: As best I can tell this was a typo in the test, previously not caught?

tests/testlistselector.py
```diff
@@ -173,16 +173,17 @@ class Q(param.Parameterized):
...
+            ValueError,
+            match=re.escape("ListSelector parameter 'r' only takes list types, not 6."),
+        ):
+            class Q(param.Parameterized):
+                r = param.ListSelector(default=6)
 
     def test_set_checked_to_be_iterable(self):
         class Q(param.Parameterized):
-            r = param.ListSelector(default=6,check_on_set=False)
+            r = param.ListSelector(default=[6],check_on_set=False)
```
jbednar, line 186: Looks like previously it was catching a non-iterable value being set, but not when it was the default. Already verified that the error was caught above, so just updated the value here.

tests/testlistselector.py
philippjfr: Definitely looks like it.

philippjfr (approved):
Looks good!

Issue #714, opened 2023-03-14
Inherited `default` is not validated


maximlt:
The value of a Parameter *attribute* - like `default`, `doc`, etc. - is inherited in a hierarchy of Parameterized classes. In practice, the inheritance only applies when the default attribute value is `None`.

In the example below, `B` is a subclass of `A` and the user intent seems to be to override the definitions of `p1` and `p2`, from `String` and `String` to `Action` to `Number`, respectively. The result of that operation is that:
- Because the default value of the `default` attribute of `Action` is `None`, `B.p1.param.default` will inherit from `A.p1.param.default` which is `'1'`. The value of `p1` is then set based on that new `default`, leading to `b.p1 == '1'`. This is an invalid value for a `Number`.
- Because the default value of the `default` attribute of `Action` is not `None` but `0.0`, it is used directly to set `B.p1.param.default`, leading to `b.p1 == 0.0`.

```python
import param

class A(param.Parameterized):
    p1 = param.String(default='1')
    p2 = param.String(default='1')

class B(A):
    p1 = param.Action()
    p2 = param.Number()

b = B()

assert b.p1 == '1'  # passes
assert b.p2 == 0.0  # passes
```

The focus of this issue is on the `default` attribute as it can be used to set the actual Parameter value, if the user does not provide a default when instantiating the class. But this issue is more general, the Parameters of a subclass can inherit other invalid attributes without any error/warning.

This discussion is linked to the proposal of enforcing an "is-a" relationship (https://github.com/holoviz/param/issues/97#issuecomment-783734974) between Parameters in a class hierarchy.

A pragmatic approach to this issue might be to run the Parameter validation code (`Parameter._validate`) after the whole inheritance procedure is executed, and emit a warning if that fails. As to avoid the issue above with `p1`, the author of `B` should define a default value for the overriding `p1`.

jbednar:
I agree that enforcing `is-a` would address this, but it also seems appropriate to run `_validate` on whatever the result of inheritance is.

jbednar:
Note that the behavior of main is now more reasonable (no longer type dependent), though still not validating as it should:

<img width="312" alt="image" src="https://github.com/holoviz/param/assets/1695496/ec9bf00d-e004-450b-b511-f021ed02c82a">

jbednar:
Fixed by #812.

maximlt:
[...]

PR #820, opened 2023-08-21
Ensure Parameter re-validates default if slot changes

philippjfr:
This PR adds validation when overriding slot values on a Parameter in a subclass. It does this by checking if a superclass defines a slot value and if that slot value is overridden by the subclass. It attempts to minimize the number of actual lookups by aborting early if we already know that re-validation has to happen and by skipping checks on various slot values which do not affect validation.

Fixes https://github.com/holoviz/param/issues/714

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -3388,7 +3388,7 @@ class ParameterizedMetaclass(type):
 
         # If the type has changed to a more specific or different type
         # or a slot value has been changed validate the default again.
-        if type_change or slot_overridden:
+        if type_change or slot_overridden and param.default is not None:
             param._validate(param.default)
 
     def get_param_descriptor(mcs,param_name):
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1114,6 +1114,13 @@ class Parameter(_ParameterBase):
         per_instance=True
     )
 
+    # During parameter instantiation we have to record which slots
+    # require the default value to be re-validated. Any slots in this
+    # list do not have to trigger such re-validation.
+    _non_validated_slots = ['_label', 'doc', 'name', 'precedence',
+                            'constant', 'pickle_default_value',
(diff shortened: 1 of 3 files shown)


philippjfr:
I had to disable re-validation of None values because the automatic appending of an unknown value on `Selector` opens a whole rabbit hole in regard to the validation.

param/parameterized.py
```diff
@@ -3325,27 +3332,35 @@ def __param_inheritance(mcs, param_name, param):
...
+                new_value = getattr(new_param, slot)
+                old_value = slot_values.get(slot, Undefined)
+                if new_value is Undefined or new_value is old_value:
+                    continue
+                elif old_value is Undefined:
+                    slot_values[slot] = new_value
+                    # If we already know we have to re-validate abort
+                    # early to avoid costly lookups
+                    if slot_overridden or type_change:
+                        continue
```
maximlt, line 3357: As far as I can see this `continue` is not skipping any code path, was it supposed to be a `break`?

jbednar (commented):
I'm having a good bit of trouble following the logic here, so hopefully we can work through it together. I think I probably wrote some of the confusing bits (e.g. what's "new" about what I think I was the one to call "new_param"?), but as the code has gotten more complex I'm now lost. The basic idea seems sound, i.e. to see if any slots have changed and thus whether we need to re-validate the default value.

param/parameterized.py
philippjfr, line 3357:
```suggestion
                        break
```

philippjfr:
@maximlt Can you either merge or ask for further clarification? You are release manager :)

maximlt:
I was waiting for https://github.com/holoviz/param/pull/826 to be merged! Now it's done I'll rebase this and merge if it goes well.

Issue #660, opened 2022-11-16
param.Range does not check if values are numbers


Hoxbro:
Param doesn't complain when the bounds of a `Range` Parameter are set with weird values.

```python
import param

class X(param.Parameterized):
    xlim = param.Range(bounds=('a', 'b'))

x = X()
```

This is because `_valudate_bounds` are not checking if the values are numbers, e.g., this is a valid comparison `"a" < "b"`.

https://github.com/holoviz/param/blob/ddcd5f36bcff1712f2c7fb984268a0ad6dc9649f/param/__init__.py#L2086-L2096

jbednar:
I can replicate this behavior, and also consider it an error that we should catch.

jbednar:
I think the code to fix this would be something like

```python
     def _validate_bounds(self, val, bounds, inclusive_bounds):
-        if bounds is None or (val is None and self.allow_None):
-            return
+        if bounds is None:
+            return
+
+        for b in bounds:
+            if b is not None and not _is_number(b):
+                raise ValueError(f"Range bound '{b}' can only be None or a "
+                                 f"numeric value, not type '{type(b).__name__}'.")
+
+        if val is None and self.allow_None:
+            return
+
```

but there will be merge conflicts if #808 is merged first, so we can come back to it after merging that.

maximlt:
Yep it will look like something like that. You'll probably want to make sure that it's a change valid for the `DateRange` and `CalendarDateRange` subclasses.

PR #824, opened 2023-08-28
Validate the type of `bounds` and `softbounds` attributes for Range parameter types

maximlt:
Fixes https://github.com/holoviz/param/issues/660

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -2988,8 +2988,8 @@ class Range(NumericTuple):
             too_high = (vmax is not None) and (v > vmax if incmax else v >= vmax)
             if too_low or too_high:
                 raise ValueError(
-                    f"{_validate_error_prefix(self, 'bound')} must be in "
-                    f"range {self.rangestr()!r}."
+                    f"{_validate_error_prefix(self)} {bound} bound must be in "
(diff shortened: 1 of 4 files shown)


philippjfr (approved):
Looks good, thanks for all the tests!

param/__init__.py
```diff
@@ -2978,7 +2978,25 @@ def _validate_order(self, val, step, allow_None):
...
+            for pos, v in zip(['lower', 'upper'], bounds):
+                if v is None:
+                    continue
+                self._validate_bound_type(v, pos, 'bound')
+        if softbounds is not None:
+            for spos, sv in zip(['lower', 'upper'], softbounds):
+                if sv is None:
+                    continue
+                self._validate_bound_type(sv, spos, 'softbound')
+
```
jbednar, line 2999: If you're changing the signature of this already, why not just change it to accept a bounds obj and a description, so that this code isn't duplicated? That way some future change to one of the code paths won't be missed on the other one.

maximlt: Refactored in https://github.com/holoviz/param/pull/824/commits/4a810acd2d41cd3835411214ad4c47b96a7b879b, I'll let you merge if you're okay with the change.

Issue #793, opened 2023-07-20
Selector regression in an inheritance context?


maximlt:
There's maybe a regression on the _main_ branch brought by the sentinel PR, that happens when setting a Selector value on an instance of a subclass and that leads to the `objects` of the super class being updated.

Here's the most basic example:

```python
class A(param.Parameterized):
    p = param.Selector()

class B(A):
    p = param.Selector()


b = B()
b.p = 2

# Before [], now [2]
print(A.param.p.objects)
```

And maybe a more realistic example:
```python
import param

class A(param.Parameterized):
    p = param.Selector(objects=[1, 2], check_on_set=False)

class B(A):
    p = param.Selector(default=2)


b = B()
b.p = 3

# Before [1, 2], now [1, 2, 3]
print(A.param.p.objects)
```

That is probably again a problem of `objects` being a mutable object, this time being shared across a class and all of its subclasses and their instances. I come to wonder whether `objects` accepting a mutable object (list/dict) is actually a good idea, and if instead it wouldn't be best if it'd accept a factory function. cc @jbednar to gauge how blocking this is for the release of Param 2.0

jbednar:
When instantiate=True leads to copying the Parameter object, can we make it do a copy (not a deepcopy) of any mutable slot values (or all slot values, which should have the same result)? That seems cheap enough and preserves object identity for the items in the Selector while avoiding this problem.

maximlt:
> When instantiate=True leads to copying the Parameter object, can we make it do a copy (not a deepcopy) of any mutable slot values (or all slot values, which should have the same result)?

Yes there's nothing that prevents us from doing that when `instantiate=True`. However that doesn't resolve neither this issue nor https://github.com/holoviz/param/issues/746 as they deal with the `Selector` Parameter which is set by default with `instantiate=False`, right?

jbednar:
I've corrected my message above to say `per_instance` where I had written `instantiate`; I always get those two confused. `per_instance` is the relevant one here, which controls instantiation of the Parameter object per instance, not instantiation of the default value into a current value (`instantiate=True`).

Issue #746, opened 2023-04-17
Selector `objects` shared across instances and class


maximlt:
When `check_on_set` is set to `False`, `objects` is updated in place and the change is seen at both the instance and class level.

![image](https://user-images.githubusercontent.com/35924738/232618202-f41cea65-dbca-40fd-8a07-e0736fb832ce.png)

```python
import param

class P(param.Parameterized):
    s = param.Selector(objects=[1, 2], check_on_set=False)

p = P()

p.s = 3
p.param.s.objects

P.param.s.objects

P.s = 4
P.param.s.objects

p.param.s.objects
```

maximlt:
While the original post focuses on `check_on_set`, it's just an outcome of the `_objects` attribute (`objects` being now a property using `_objects) being shared across instances and classes:

```python
import param

class P(param.Parameterized):
    s = param.Selector(objects=[1, 2])

p1 = P()
p2 = P()

assert p1.param.s._objects is p2.param.s._objects is P.param.s._objects
```

I was a bit surprised by that when I started to think about `per_instance` being `True` by default for Selectors. Except that the Parameter copy each instance holds is merely a shallow copy of the class Parameter, this is by design (https://github.com/holoviz/param/pull/306), and it explains they all just hold a reference to the same list/dict.

---

I suggest we push this to post Param 2.0 given the breadth of implications "fixing" that could have. cc @jbednar

jbednar:
Seems like you're suggesting the same approach as I just proposed over in #793, i.e. shallow copying Parameter slots, which I think can be done in general (not just for Selectors) unless Selectors truly are the only Parameters with mutable slot values.  I think we should at least try that to see if it raise any unexpected problems, because this is a subtle change to the semantics that I _think_ will be a bugfix, but which is one to make in 2.0 if possible if it has implications.

maximlt:
I started to look into that and had a hackish implementation I ran across Panel's test suite. Ignoring some horrible failures that might be due to the hackish implementation, I found one failure that is relevant to that issue as it's caused by a [test](https://github.com/holoviz/panel/blob/df0b277dca2fd54dd2ae016159adc5c33b61e358/panel/tests/test_param.py#L951) that assumed `objects` could be updated on a class Parameter in order to affect already created instances (comment is mine):

```python
def test_switch_param_subobject(document, comm):
    class Test(param.Parameterized):
        a = param.ObjectSelector()

    o1 = Test(name='Subobject 1')
[...]

PR #826, opened 2023-08-30
Shallow-copy mutable containers in slot values 

jbednar:
Fixes https://github.com/holoviz/param/issues/793.
Fixes https://github.com/holoviz/param/issues/746.

Adds shallow copying of mutable slot values when instantiating Parameter objects or inheriting values for slots. Shallow copying is only done when `per_instance` is `True` (the default); if you don't want such copying you can share both Parameter instances and mutable slot values by setting `per_instance=False`.

```python
import param

class A(param.Parameterized):
    p = param.Selector(objects=[1, 2], check_on_set=False)

class B(A):
    p = param.Selector(default=2)


b = B()
b.p = 3

print(A.param.p.objects, B.param.p.objects, b.param.p.objects)
# [1, 2] [1, 2, 3] [1, 2, 3]
```
(previously A's objects list was also [1, 2, 3]; now fixed to be [1, 2])

There are various issues still, as listed below.

## Tests

Needs a test capturing the example above, but I've got to run, so if anyone wants to add that I'd be grateful, or I can do that tomorrow.

## Previous handling of watchers

PR https://github.com/holoviz/param/pull/306 added this logic to Parameters.__getitem__() for shallow-copying watchers:

```python
try:
    # Do not copy watchers on class parameter
    watchers = p.watchers
    p.watchers = {}
    p = copy.copy(p)
except:
    raise
finally:
    p.watchers = {k: list(v) for k, v in watchers.items()}
```

This PR generalizes such copying to apply to all the slot values besides default, but I was unable to determine why the existing code was so complex. First, why `p.watchers = {k: list(v) for k, v in watchers.items()}` instead of just `p.watchers = copy.copy(watchers.items())`? copy.copy is used just a couple of lines before, and when I use it here it seems to work just the same, as I'd expect, so I can't tell why it was a list comprehension that appears to achieve the same as copy.copy.

Second, what's with the try/except/finally? Deleting all that seems to work just fine, and I can't quite imagine a situation when the try block could fail but the watchers are still valid to reconstruct finally. Maybe something in Panel? For now I deleted it as it does not _seem_ to achieve anything.

Third, I've kept the convoluted mechanism for zeroing-out the watchers in the original Parameter object when creating the new one. Simply shallow-copying the watchers after copying the Parameter object leaves an extra set of watchers that causes tests to fail, but I'm not at all sure why it's appropriate to move the watchers to the new Parameter object rather than where they were originally watching. For now I chose to be conservative, but I'm not confident why it was done this way, so if there's a reason, it should have a clear comment added.

## What's mutable and when is it copied?

For now, mutable sequences, mappings, and sets are considered mutable, which covers the list and dict cases that I'm aware of us using in slots. Such values are shallow-copied for all slots _except_ `default`, to ensure that any independent Parameter copies also have independent objects lists, etc. Parameterized objects are also mutable, but I don't know if they are ever used in slots. If they are, we should consider whether they should be shallow-copied, but my guess is that they should not, since it's largely containers that we want shallow-copied, not arbitrary instantiated objects. Maybe `is_mutable` should be renamed `is_mutable_container` to convey that?

The reason for not copying the `default` value is to allow a specific list or dict to be shared across Parameters, which has always been supported. E.g. one could have a few global lists like `search_paths`, `debug_search_paths`, etc., whose values are curated and maintained independently of which particular Parameters have been set to those values. We can revisit whether mutable containers should be copied for `default` too, but if so, it should be a separate PR with its own justification.

I currently check for `default` only when copying the Parameter object, not during inheritance. It might also be appropriate to skip shallow-copying for for inheritance as well, but I haven't found an example where that would come up. Probably worth doing?

## _update_state

After merging this PR, I believe we should be able to remove _update_state as discussed in https://github.com/holoviz/param/issues/807 and https://github.com/holoviz/param/pull/817 . _update_state was needed because we had to wait for param inheritance before adding to the objects lists for Selectors, but now that values get shallow copied, it should be feasible to populate the objects lists (e.g. from the default value) immediately, during the constructor. Doing so should simplify the Selector logic, but I haven't yet tested that this PR's approach will achieve that goal.

My guess is that eliminating _update_state is required, for fixing this remaining bad behavior:

0.

```python
import param

class A(param.Parameterized):
    p = param.Selector(objects=[1, 2], check_on_set=False, per_instance=True)
    
class B(A):
    i = param.Integer(8)

b = B()
b.p = 3

print(A.param.p.objects, B.param.p.objects, b.param.p.objects)
#[1, 2, 3] [1, 2, 3] [1, 2, 3]
```
Here the Parameter object isn't copied into B, and the new value 3 ends up in A's object list, which it should not. I _think_ that may be from _update_state being called on the class's copy of the objects list, not the instance's, since it's called in the metaclass, but I haven't tracked that down.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -42,6 +42,7 @@ from ._utils import (
     _recursive_repr,
     _validate_error_prefix,
     ParamDeprecationWarning as _ParamDeprecationWarning,
+    _is_mutable_container,
 )
 
 try:
@@ -963,10 +964,6 @@ class _ParameterBase(metaclass=ParameterMetaclass):
         cls.__signature__ = new_sig
 
 
-def is_mutable(value):
-    """True if the given object is known to be mutable"""
-    return issubclass(type(value), (abc.MutableSequence, abc.MutableSet, abc.MutableMapping))
-
 class Parameter(_ParameterBase):
     """
     An attribute descriptor for declaring parameters.
@@ -1708,7 +1705,7 @@ def _instantiate_param_obj(paramobj, owner=None):
     # shallow-copy any mutable slot values other than the actual default
     for s in p.__class__.__slots__:
         v = getattr(p, s)
-        if is_mutable(v) and s != "default":
+        if _is_mutable_container(v) and s != "default":
             setattr(p, s, copy.copy(v))
     return p
 
@@ -3377,7 +3374,7 @@ class ParameterizedMetaclass(type):
 
             # Avoid crosstalk between mutable slot values in different Parameter objects
             v = getattr(param, slot)
-            if (not getattr(param, "constant", False) and is_mutable(v)):
+            if (not getattr(param, "constant", False) and _is_mutable_container(v)):
                 setattr(param, slot, copy.copy(v))
 
         # Once all the static slots have been filled in, fill in the dynamic ones
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1708,11 +1708,16 @@ def _instantiate_param_obj(paramobj, owner=None):
     # Shallow-copy Parameter object, with special handling for watchers
     # (from try/except/finally in Parameters.__getitem__ in https://github.com/holoviz/param/pull/306)
     p = paramobj
-    watchers = p.watchers
-    p.watchers = {}
-    p = copy.copy(p)
+    try:
+        # Do not copy watchers on class parameter
+        watchers = p.watchers
+        p.watchers = {}
+        p = copy.copy(p)
+    except:
+        raise
+    finally:
+        p.watchers = {k: list(v) for k, v in watchers.items()}
 
-    p.watchers = {k: list(v) for k, v in watchers.items()}
     p.owner = owner
 
     # shallow-copy any mutable slot values other than the actual default
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -24,7 +24,7 @@ try:
 except ImportError:
     serializer = None
 
-from collections import defaultdict, namedtuple, OrderedDict, abc
+from collections import defaultdict, namedtuple, OrderedDict
 from functools import partial, wraps, reduce
 from html import escape
 from operator import itemgetter, attrgetter
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1699,7 +1699,7 @@ def _instantiate_param_obj(paramobj, owner=None):
     p.watchers = {}
     p = copy.copy(p)
 
-    p.watchers = copy.copy(watchers)
+    p.watchers = {k: list(v) for k, v in watchers.items()}
     p.owner = owner
 
     # shallow-copy any mutable slot values other than the actual default
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -400,17 +400,35 @@ def iscoroutinefunction(function):
         return False
 
 
+def _instantiated_parameter(parameterized, param):
+    """
+    Given a Parameterized object and one of its class Parameter objects,
+    return the appropriate Parameter object for this instance, instantiating
+    it if need be.
+    """
+    if (getattr(parameterized._param__private, 'initialized', False) and param.per_instance and
+        not getattr(type(parameterized)._param__private, 'disable_instance_params', False)):
+        key = param.name
+
+        if key not in parameterized._param__private.params:
+            parameterized._param__private.params[key] = _instantiate_param_obj(param, parameterized)
+
+        param = parameterized._param__private.params[key]
+
+    return param
+
+
 def instance_descriptor(f):
     # If parameter has an instance Parameter, delegate setting
     def _f(self, obj, val):
-        if obj is None:
-            # obj is None when the metaclass is setting
-            return f(self, obj, val)
-        params = obj._param__private.params
-        instance_param = None if params is None else params.get(self.name)
-        if instance_param is not None and self is not instance_param:
-            instance_param.__set__(obj, val)
-            return
+        # obj is None when the metaclass is setting
+        if obj is not None:
+            params = obj._param__private.params
+            instance_param = None if params is None else params.get(self.name)
+
+            if instance_param is not None and self is not instance_param:
+                instance_param.__set__(obj, val)
+                return
         return f(self, obj, val)
     return _f
 
@@ -1785,17 +1803,10 @@ class Parameters:
         Returns the class or instance parameter
         """
         inst = self_.self
-        parameters = self_.objects(False) if inst is None else inst.param.objects(False)
-        p = parameters[key]
-        if (inst is not None and getattr(inst._param__private, 'initialized', False) and p.per_instance and
-            not getattr(self_.cls._param__private, 'disable_instance_params', False)):
+        params = self_ if inst is None else inst.param
 
-            if key not in inst._param__private.params:
-                inst._param__private.params[key] = _instantiate_param_obj(p, inst)
-
-            p = inst._param__private.params[key]
-
-        return p
+        p = params.objects(False)[key]
+        return p if inst is None else _instantiated_parameter(inst,p)
 
 
     def __dir__(self_):
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -400,6 +400,32 @@ def iscoroutinefunction(function):
         return False
 
 
+def _instantiate_param_obj(paramobj, owner=None):
+    """Return a Parameter object suitable for instantiation given the class's Parameter object"""
+
+    # Shallow-copy Parameter object, with special handling for watchers
+    # (from try/except/finally in Parameters.__getitem__ in https://github.com/holoviz/param/pull/306)
+    p = paramobj
+    try:
+        # Do not copy watchers on class parameter
+        watchers = p.watchers
+        p.watchers = {}
+        p = copy.copy(p)
+    except:
+        raise
+    finally:
+        p.watchers = {k: list(v) for k, v in watchers.items()}
+
+    p.owner = owner
+
+    # shallow-copy any mutable slot values other than the actual default
+    for s in p.__class__.__slots__:
+        v = getattr(p, s)
+        if _is_mutable_container(v) and s != "default":
+            setattr(p, s, copy.copy(v))
+    return p
+
+
 def _instantiated_parameter(parameterized, param):
     """
     Given a Parameterized object and one of its class Parameter objects,
@@ -1700,32 +1726,6 @@ class _ParametersRestorer:
             self._parameters._update(self._restore)
         finally:
             self._restore = {}
-
-
-def _instantiate_param_obj(paramobj, owner=None):
-    """Return a Parameter object suitable for instantiation given the class's Parameter object"""
-
-    # Shallow-copy Parameter object, with special handling for watchers
-    # (from try/except/finally in Parameters.__getitem__ in https://github.com/holoviz/param/pull/306)
-    p = paramobj
-    try:
-        # Do not copy watchers on class parameter
-        watchers = p.watchers
-        p.watchers = {}
-        p = copy.copy(p)
-    except:
-        raise
-    finally:
-        p.watchers = {k: list(v) for k, v in watchers.items()}
-
-    p.owner = owner
-
-    # shallow-copy any mutable slot values other than the actual default
-    for s in p.__class__.__slots__:
-        v = getattr(p, s)
-        if _is_mutable_container(v) and s != "default":
-            setattr(p, s, copy.copy(v))
-    return p
 
 
 class Parameters:
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -426,6 +426,9 @@ def instance_descriptor(f):
             params = obj._param__private.params
             instance_param = None if params is None else params.get(self.name)
 
+            if (params is not None and instance_param is None):
+                instance_param = _instantiated_parameter(obj, self)
+
             if instance_param is not None and self is not instance_param:
                 instance_param.__set__(obj, val)
                 return
@@ -1404,7 +1407,7 @@ class Parameter(_ParameterBase):
                 warnings.warn(
                     'Number.set_hook has been deprecated.',
                     category=_ParamDeprecationWarning,
-                    stacklevel=5,
+                    stacklevel=6,
                 )
 
         self._validate(val)
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -3373,9 +3373,10 @@ class ParameterizedMetaclass(type):
                     setattr(param, slot, default_val)
 
             # Avoid crosstalk between mutable slot values in different Parameter objects
-            v = getattr(param, slot)
-            if (not getattr(param, "constant", False) and _is_mutable_container(v)):
-                setattr(param, slot, copy.copy(v))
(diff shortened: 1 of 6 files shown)


param/parameterized.py
```diff
@@ -3358,6 +3375,11 @@ def __param_inheritance(mcs, param_name, param):
                 else:
                     setattr(param, slot, default_val)
 
+            # Avoid crosstalk between mutable slot values in different Parameter objects
+            v = getattr(param, slot)
+            if (not getattr(param, "constant", False) and is_mutable(v)):
+                setattr(param, slot, copy.copy(v))
```
maximlt, line 3381: In a simple case like below do we need to make a shallow copy of the objects? Currently it does.

```python
class P(param.Parameterized):
    s = param.Selector(objects=[1, 2]))

File: param/parameterized.py [hunk 2]
```diff
@@ -3358,6 +3375,11 @@ def __param_inheritance(mcs, param_name, param):
                 else:
                     setattr(param, slot, default_val)
 
+            # Avoid crosstalk between mutable slot values in different Parameter objects
+            v = getattr(param, slot)
+            if (not getattr(param, "constant", False) and is_mutable(v)):
```
maximlt, line 3380: Why dealing with `constant` here?

tests/testparameterizedobject.py
```diff
@@ -276,7 +276,7 @@ def test_constant_parameter_modify_class_after_init(self):
         even when the instance value hasn't yet been set"""
         oobj = []
         class P(param.Parameterized):
-            x = param.Parameter(default=oobj, constant=True)
+            x = param.Parameter(default=oobj, constant=True, per_instance=False)
```
maximlt, line 279: This test passes with `per_instance=True`, this probably needs to be reverted.

param/parameterized.py
maximlt: Oh ok I see, that had to do with the failing test. Why not just doing like in https://github.com/holoviz/param/pull/826/files#diff-c8fc4fb68012b66fbf42ad9e0b4714fd96a3a72b83558bc543130d78bd622eb3R1711 and not setting the parameter slot when the slot is `'default'`?

param/parameterized.py
```diff
@@ -963,6 +963,10 @@ def _sorter(p):
         cls.__signature__ = new_sig
 
 
+def is_mutable(value):
```
maximlt, line 966: Param has now a `_utils.py` module in which this utility could live. Also I've been trying to make any new function/class private not to pollute the namespace. So including your suggestion I'd rename it `_is_mutable_container` and move it to `_utils.py`.

param/parameterized.py
```diff
@@ -1688,6 +1692,27 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+    """Return a Parameter object suitable for instantiation given the class's Parameter object"""
+
+    # Shallow-copy Parameter object, with special handling for watchers
+    # (from try/except/finally in Parameters.__getitem__ in https://github.com/holoviz/param/pull/306)
+    p = paramobj
+    watchers = p.watchers
+    p.watchers = {}
+    p = copy.copy(p)
+
+    p.watchers = copy.copy(watchers)
```
maximlt, line 1705: For context here is the [PR](https://github.com/holoviz/param/pull/419) that changed `p.watchers = watchers` to `p.watchers = {k: list(v) for k, v in watchers.items()}`. @philippjfr if by any chance you remember how to reproduce the error mentioned in the PR we could write a test for it.

@jbednar I'm also unsure why a dict comprehension was used instead of simpler approaches like `watchers.copy()`.

param/parameterized.py
jbednar: Because I didn't think of that. :-) I think that's the right approach, at least for now; copying `default` is a separate issue.

maximlt: And I see now your related comment:

> I currently check for default only when copying the Parameter object, not during inheritance. It might also be appropriate to skip shallow-copying for for inheritance as well, but I haven't found an example where that would come up. Probably worth doing?

I'd be in favor not to touch `default` in this PR.

maximlt:
- on *Previous handling of watchers*, I agree we need to better understand what was going on there, hopefully Philipp can remember some of that otherwise we'll have to try to dig more into it
- on `_update_state`, I wish we could get rid of it but I also know that the Selector(s) Parameters are quite dynamic (some slot default being callable that depend on the state of other slots) and are difficult to setup correctly when the slots aren't populated in the expected order, which happens in an inheritance context.

---

What do you think about the examples below? In the first snippet the code accesses the Parameter **before** setting `p.s` to 3, to run `__getitem__` and shallow-copy the objects. In the second case this is commented out, leading to the objects on the class Parameter being updated as the Parameter doesn't yet exist on the instance. This is similar to the last example you shared @jbednar, without any inheritance though.

1)
```python
class P(param.Parameterized):
    s = param.Selector(objects=[1, 2], check_on_set=False)

p = P()
p.param.s.objects  # runs `__getitem__`
p.s = 3
assert P.param.s.objects == [1, 2]
```

2)
```python
class P(param.Parameterized):
    s = param.Selector(objects=[1, 2], check_on_set=False)

p = P()
# p.param.s.objects
p.s = 3
assert P.param.s.objects == [1, 2, 3]
```

---

param/parameterized.py
jbednar: Yes, if `per_instance` is True, I think we should be making a copy of the objects list per instance.  I think we should not  make any copies if `per_instance=False`.

param/parameterized.py
jbednar: Sounds good, thanks!

param/parameterized.py
jbednar: Ok, done and pushed.

jbednar:
@maximlt , that's a good point:

image

Maybe we need to populate the Parameter objects more proactively. Any suggestions for where to make that change?

param/parameterized.py
philippjfr: The problem is that both the dictionary itself and the `list` values need to be copied but we do not want to copy the `watcher` objects that are already there.

philippjfr: What was wrong with the dictionary comprehension?

jbednar: copy.copy is shallow, and therefore doesn't copy the underlying objects, right?

maximlt:
> Any suggestions for where to make that change?

Maybe at the start of `Parameters._setup_params` which is called from `Parameterized.__init__`?

param/parameterized.py
philippjfr: I can try, watcher's were originally Chris' thing so I frequently have re-familiarize myself as well.

maximlt:
Actually the behavior I reported in my previous comment is not tied to mutable containers.

image

<details>
 <summary>Code</summary>

```python
import param

class P(param.Parameterized):
    x = param.Number()

p = P()

P.param.x.bounds = (100, 200)

assert p.param.x.bounds == (100, 200)

class P(param.Parameterized):
    x = param.Number()

p = P()

p.param.x  # side-effects there

P.param.x.bounds = (100, 200)

assert p.param.x.bounds is None
```

</details>

Accessing a Parameter on the `.param` namespace creates a Parameter on the instance. I'm not sure it's really a bug, certainly it's confusing, but really the way to get a Parameter you can control at the class level should be with `per_instance=False`.

philippjfr:
My reading of the watchers code above is that it's not fully correct. The problem being that we can't distinguish between watchers that are meant to be on the class parameter and watchers that are meant to be on the instance parameter since there is no difference until the instance parameter is created, i.e.:

```python
param.Parameterized.param.watch(print, 'name', what='allow_None')
```

looks exactly the same as:

```python
p = param.Parameterized()
p.param.watch(print, 'name', what='allow_None')
```

In general it seems very rare to have watchers for attributes on class parameters so this hasn't come up, but it's definitely not correct. I think this is another issue we can't easily fix without making instance parameter creation non-lazy.

>Second, what's with the try/except/finally? Deleting all that seems to work just fine, and I can't quite imagine a situation when the try block could fail but the watchers are still valid to reconstruct finally. Maybe something in Panel? For now I deleted it as it does not seem to achieve anything.

I do think it's important, at least if we decide not to fix the underlying issue yet. If there is somehow an error while creating the instance parameter (e.g. because some value is not safe to copy) then the watchers end up being lost entirely.

jbednar:
> If there is somehow an error while creating the instance parameter (e.g. because some value is not safe to copy) then the watchers end up being lost entirely.

How would it help to have watchers copied correctly but not whatever slot value was about to be copied after the failing one? This try/except/finally seems like it would just cover up problems, not avoid them -- why would we want watchers copied correctly when other slots aren't?  If we wanted to copy all the slots that we could, and put a warning for any that didn't work, wouldn't we do that in a loop where the try/except was per slot, not for the entire copy operation?

philippjfr:
It's not about the object being copied to but about restoring the class that's being copied from.

jbednar:
Hmm; that's weird. My understanding of that code is that if copy.copy fails, an exception will be raised, with `p` remaining the class Parameter object, except that `p.watchers` has been replaced with a copy of the original dict. Seems like a lot of hoops to be jumping back and forth through to achieve that outcome, i.e. emptying out the watchers and then replacing it with something just like the original.

And then if copy.copy succeeds, no exception will be raised, and `p` will become the instance's copy of the Parameter, where the instance `p.watchers` will have a copy of the original dict. But won't the original class Parameter now have an empty watchers dict? How is this working at all when there are multiple instances that need to have the same Parameter instantiated into it? Only the first one gets the watchers? 

Unless I'm very confused, it seems like we have two defensible options here: (1) Leave watchers on class Parameter objects, and don't copy them to instances. Instances would need their own watchers, which would be specific to that instance. Watchers on class Parameter objects would be watching the class Parameter object, not instances. (2) Leave watchers on Class Parameter objects, and copy them to all instances. Thus anything that watches the class Parameter will also be watching all the instances.

I am not seeing any argument for what it seems like the current code is doing, i.e. emptying out the watchers from the class Parameter object after instantiation. When is that ever appropriate?

jbednar:
@maximlt , I've pushed a few more commits to address the issues above:

- In 9ba4aa1 I've restored the watchers logic to its original state using try/except/finally; if it's to be addressed, that should be done in a separate PR. It should now be unchanged from before this PR, just refactored differently.
- In 7fcf8d4 I refactored the new `Parameter.__getitem__` implementation to pull out  the logic for copying Parameter objects `per_instance` into a new function `_instantiated_parameter()`. This commit should not change _anything_ about the behavior; it just makes that logic reusable elsewhere.
- In 6ce1ef3 I make `Parameter.__set__` use `_instantiated_parameter()` to ensure that the Parameter object has been copied `per_instance` before any setting is done, so that e.g. `_ensure_value_is_in_objects` will affect the `per_instance` Parameter object instead of the class's Parameter object. In this commit I had to update one warning message because it's being generated another level deep in the stack trace, plus I deleted one newly-added line from the tests that was essentially enforcing that the `objects` on a Selector were stored only in the class, which this commit deliberately changes.

6ce1ef3 is the key one for addressing the problems above. After that commit, I believe these behaviors are now correct:

0.

```python
import param

class A(param.Parameterized):
    p = param.Selector(objects=[1, 2], check_on_set=False, per_instance=True)

class B(A):
    i = param.Integer(8)

b = B()
b.p = 3

print(A.param.p.objects, B.param.p.objects, b.param.p.objects)
# [1, 2] [1, 2] [1, 2, 3]
```

(`b` now gets its own `per_instance` copy of the objects list)

1.

```python
class P(param.Parameterized):
    s = param.Selector(objects=[1, 2], check_on_set=False)

p = P()
p.param.s.objects
p.s = 3
print(P.param.s.objects, p.param.s.objects)
#[1, 2] [1, 2, 3]
```

2.

```python
class P(param.Parameterized):
    s = param.Selector(objects=[1, 2], check_on_set=False)

p = P()
p.s = 3
print(P.param.s.objects, p.param.s.objects)
#[1, 2] [1, 2, 3]
```

(invoking `__getitem__` is no longer required for `p` to get its own `per_instance` copy of the `s` Parameter object)


I'm ok with keeping the behavior you show in 3. above, i.e. that accessing a Parameter on the `.param` namespace creates a Parameter on the instance. I agree that if you want to control a Parameter at the class level, you can _either_ change the class values _before_ creating any instances, _or_ you can set `per_instance=False`. If you change the Parameter at the class level after having instantiated some classes, there's no guarantee that the class-level changes will propagate down. As you show, sometimes they do, but really if you want such changes you have to make them on the instance or you have to set `per_instance` to False!

So far, I haven't messed with _update_state, because changing that wasn't necessary for anything to do with instantiating slots. 

So this PR should be ready to review and merge now; it now does what it says: adds shallow-copying for Parameter slot values that happen to be mutable containers, whenever a Parameter object is copied to an instance or inherits its slot values. Parameter objects are still copied only lazily, as before, but now there is one more case where it gets copied, namely a when a parameter value is set on the instance (to handle `_ensure_value_is_in_objects`).

maximlt (commented):
I left a question in my review. In addition to that, I would like to get your opinion @jbednar on the following examples?

1.
```python
class A(param.Parameterized):
    p = param.Selector(objects=[1, 2])

class B(A):
    p = param.Selector(default=2)

B.param.p.objects.append(3)
print(A.param.p.objects)  # [1, 2]
```

2.
```python
class A(param.Parameterized):
    p = param.Selector(objects=[1, 2])

class B(A):
    pass

B.param.p.objects.append(3)
print(A.param.p.objects)  # [1, 2, 3]
```

param/parameterized.py
```diff
@@ -3358,6 +3383,12 @@ def __param_inheritance(mcs, param_name, param):
                 else:
                     setattr(param, slot, default_val)
 
+            # Avoid crosstalk between mutable slot values in different Parameter objects
+            if slot != "default":
```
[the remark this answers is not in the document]
maximlt: I should add, I ran the Panel and HoloViews unit tests suite against this PR, no error!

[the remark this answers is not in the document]
jbednar: (And in any case would be a separate PR, not this one, so that it can be explained and justified very clearly.)

jbednar:
I assume everyone is happy with the behavior in case 1 (two separate Parameter objects, and two separate objects lists)? 

Case 2 is trickier. On the one hand, the current behavior in 2 seems like the only way to define a Parameter object whose settings can easily be inherited across a hierarchy. I.e., not copying the Parameter object between A and B is what lets people mess with the one in A and set things up as they like, while knowing that it will all propagate nicely. Plus, if they do want separate Parameter objects in subclasses, it's easy enough to just put them there, as in 1. 

On the other, I do agree that there's some argument for duplicating the Parameter automatically to make case 2 act like case 1, so that they behave independently. On balance I think it's ok that we aren't doing that; seems like we lose more than we gain.

param/parameterized.py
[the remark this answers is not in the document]
jbednar: > I should add, I ran the Panel and HoloViews unit tests suite against this PR, no error!

I guess that's good! (Though really it means we need more tests!)

philippjfr:
I'll be addressing watcher related issues in a separate PR. I also think that on balance the behavior in the example above is defensible. So I'm going to consider this ready and merge.

Issue #829, opened 2023-09-04
Class-level watchers and copying watchers behaviors


maximlt:
In https://github.com/holoviz/param/pull/826 a discussion started on how the watchers are specially handled when a Parameter is copied from class to an instance. I've come up with some examples that exhibit some bugs and interesting behavior (e.g. side effects of calling `p.param.x`).

1. Class-level watcher on `default`

```python
store = []
def cb(event): store.append(event)

class P(param.Parameterized):
    x = param.Parameter()

P.param.watch(cb, 'x')

P.x = 10

assert len(store) == 1

print(P.param.x.watchers)
assert 'value' in P.param.x.watchers

p = P()

print(P.param.x.watchers)
assert 'value' in P.param.x.watchers

p.param.x  # side-effect

print(P.param.x.watchers)
assert P.param.x.watchers == {}  # :( Watchers no longer saved on the class Parameter

print(p.param.x.watchers)
assert 'value' in p.param.x.watchers  # :( Now saved on the instance Parameter

P.x = 20
p.x = 30

assert len(store) == 1  # Changes at the class or instance level no longer re-trigger the callback.
```

2. Class-level watcher on an Parameter attribute (`constant`)

```python
store = []
def cb(event): store.append(event)

class P(param.Parameterized):
    x = param.Parameter()

P.param.watch(cb, 'x', 'constant')

assert 'constant' in P.param.x.watchers

P.param.x.constant = True

assert len(store) == 1

p = P()

assert 'constant' in P.param.x.watchers
print(P.param.x.watchers)

p.param.x  # side-effect

assert P.param.x.watchers == {}  # :( Watchers no longer saved on the class Parameter
print(P.param.x.watchers)

print(p.param.x.watchers)
assert 'constant' in p.param.x.watchers  # :( Now saved on the instance Parameter

p.param.x.constant = False

assert len(store) == 2  # Changing constant on the instance triggers the callback

P.param.x.constant is True  # just checking the value

P.param.x.constant = False

assert len(store) == 2  # Changes at the class level no longer re-trigger the callback.
```

3. Instance-level watcher on `value`

```python
class P(param.Parameterized):
    x = param.Parameter()
    l = param.Parameter([])
    
    @param.depends('x', watch=True)
    def cb(self):
        self.l.append(self.x)

print(P.param.x.watchers)
assert P.param.x.watchers == {} 

print(P.param._depends)
assert P.param._depends['watch'][0][3][0].what == 'value'

p = P()

[...]

PR #833, opened 2023-09-15
Ensure class watchers are not inherited by instance parameter

philippjfr:
In https://github.com/holoviz/param/pull/826 we discovered some weird logic that was copying over watchers from class to instance parameters. After discussing this we all agreed this does not make sense, this is because slots are not shared between class and instance parameters and therefore it does not make sense that an instance parameter would inherit the watchers from a completely different `Parameter` instance. Worse yet, once an instance was created the watcher would be removed from the class which means that modifying the class `Parameter` slot value would no longer trigger the watcher.

The correct behavior is actually quite simple, watchers should not be copied over, instead the `Parameter` slot watchers should be reset when the instance parameter is created. This ensure that while the instance still shares the `Parameter` object with the class it correctly fires if a slot on the class Parameter is changed and when an instance parameter is created only newly created watchers on the instance will be watched. Since calling `inst.param.watch` will automatically create instance parameters for any parameters that are watched this means that the behavior will be correct in all cases both instances and classes will always correctly trigger the appropriate watchers.

Fixes https://github.com/holoviz/param/issues/829

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2693,17 +2693,15 @@ class Parameters:
                 raise ValueError("{} parameter was not found in list of "
                                  "parameters of class {}".format(parameter_name, self_.cls.__name__))
 
-            inst = self_.self
-            if inst is not None and what == "value":
-                watchers = inst._param__private.watchers
+            if self_.self is not None and what == "value":
+                watchers = self_.self._param__private.watchers
                 if parameter_name not in watchers:
                     watchers[parameter_name] = {}
                 if what not in watchers[parameter_name]:
                     watchers[parameter_name][what] = []
                 getattr(watchers[parameter_name][what], action)(watcher)
             else:
-                param_obj = self_[parameter_name]
-                watchers = param_obj.watchers
+                watchers = self_[parameter_name].watchers
                 if what not in watchers:
                     watchers[what] = []
                 getattr(watchers[what], action)(watcher)
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -419,13 +419,13 @@ def _instantiate_param_obj(paramobj, owner=None):
     return p
 
 
-def _instantiated_parameter(parameterized, param):
+def _instantiated_parameter(parameterized, param, force=False):
     """
     Given a Parameterized object and one of its class Parameter objects,
     return the appropriate Parameter object for this instance, instantiating
     it if need be.
     """
-    if (getattr(parameterized._param__private, 'initialized', False) and param.per_instance and
+    if ((getattr(parameterized._param__private, 'initialized', False) or force) and param.per_instance and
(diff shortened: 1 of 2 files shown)


param/parameterized.py
```diff
@@ -403,21 +403,14 @@ def iscoroutinefunction(function):
...
-        p.watchers = {k: list(v) for k, v in watchers.items()}
-
+    # Shallow-copy Parameter object
+    p = copy.copy(paramobj)
     p.owner = owner
 
+    # Reset watchers since class parameter watcher should not execute
+    # on instance parameters
+    p.watchers = {}
+
```
jbednar, line 413: That's a very pleasing simplification! Hope it will work well in practice.

philippjfr:
So it turns out the copying was effectively a hack to allow watchers that are set up during instance creation (i.e. primarily ones set up using `@param.depends` on a method) to be copied over after the instance is fully set up. I'm going to revert my latest changes which were forcing the creation of an instance parameter even before the class was fully initialized since it can potentially cause bad side-effects should someone try to set up a watcher before they call `super(Parameterized self).__init__`. Instead I will declare the instance initialized before we call `_update_deps` which is what is responsible for setting up the dependencies.

param/parameterized.py
```diff
@@ -2700,15 +2693,17 @@ def _register_watcher(self_, action, watcher, what='value'):
                 raise ValueError("{} parameter was not found in list of "
                                  "parameters of class {}".format(parameter_name, self_.cls.__name__))
 
-            if self_.self is not None and what == "value":
-                watchers = self_.self._param__private.watchers
+            inst = self_.self
```
jbednar, line 2696: Any reason to create `inst` now?

philippjfr: Not really can revert all of this I guess. Still reads a little clearer but not much.

param/parameterized.py
```diff
@@ -2700,15 +2693,17 @@ def _register_watcher(self_, action, watcher, what='value'):
...
+                watchers = inst._param__private.watchers
                 if parameter_name not in watchers:
                     watchers[parameter_name] = {}
                 if what not in watchers[parameter_name]:
                     watchers[parameter_name][what] = []
                 getattr(watchers[parameter_name][what], action)(watcher)
             else:
-                watchers = self_[parameter_name].watchers
+                param_obj = self_[parameter_name]
+                watchers = param_obj.watchers
```
jbednar: Does this change achieve anything?

maximlt:
I ran Panel and HoloViews unit test suites successfully against this branch. Merging, thanks!

maximlt:
I ran Panel and HoloViews unit test suites successfully against this branch. Merging, thanks!

PR #835, opened 2023-09-18
Make `check_on_set` static and `True` by default

maximlt:
This was possible with only making a few changes to the test suite. However, removing `_update_state` proves to be less straightforward, I'll keep looking tomorrow.

### DIFF
--- a/param/__init__.py
+++ b/param/__init__.py
@@ -1652,25 +1652,11 @@ class __compute_selector_default:
 _compute_selector_default = __compute_selector_default()
 
 
-class __compute_selector_checking_default:
-    def __call__(self, p):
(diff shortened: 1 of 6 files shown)


jbednar (commented):
Looks good so far!

maximlt:
Running this branch against Panel's unit tests suite shows a number of errors, which are a hint that this change cannot be made without going through a normal deprecation period.

jbednar:
Ok, let's include a statement in the Param 2.0 release notes that setting `check_on_set=False` explicitly is now required in Param 2.0 in any case where the user wants it to be False, and mention that later releases will warn and eventually disallow it.

PR #838, opened 2023-09-22
Warn instead or raise an error when doing an action before Parameterized initialization

maximlt:
https://github.com/holoviz/param/pull/834 added a few checks for preventing users from calling unsafe methods before a Parameterized instance has fully been initialized. This broke the hvPlot Explorer and is likely to break quite a lot of code, in particular for users who aren't so aware of the importance and meaning of calling `super().__init__()`. I propose with this PR to downgrade these errors to warnings, to FutureWarnings in particular that will always be surfaced to users, as they're likely to be shooting themselves in the foot and we want to let them know about it.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -43,6 +43,7 @@ from ._utils import (
     _recursive_repr,
     _validate_error_prefix,
     ParamDeprecationWarning as _ParamDeprecationWarning,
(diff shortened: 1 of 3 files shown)


philippjfr (approved):
Agree with this, thank you!

PR #839, opened 2023-09-22
Mitigating performance regressions in 2.0

philippjfr:
An attempt at mitigating some of the performance regressions associated with Param 2.0

# Comparing 2.0.0a3 vs this branch

Screen Shot 2023-09-22 at 15 35 39

# Comparing 1.13.0 vs this branch

Screen Shot 2023-09-22 at 15 36 19

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -908,10 +908,10 @@ class ParameterMetaclass(type):
         # when asking for help on Parameter *object*, return the doc slot
         classdict['__doc__'] = property(attrgetter('doc'))
 
-        # Compute all slots
-        all_slots = set()
-        for bcls in set(chain(*(base.__mro__ for base in bases))):
-            all_slots |= set(getattr(bcls, '__slots__', set()))
+        # Compute all slots in order
+        all_slots = {}
+        for bcls in set(chain(*(base.__mro__[::-1] for base in bases))):
+            all_slots.update(dict.fromkeys(getattr(bcls, '__slots__', [])))
 
         # To get the benefit of slots, subclasses must themselves define
         # __slots__, whether or not they define attributes not present in
@@ -920,7 +920,7 @@ class ParameterMetaclass(type):
         if '__slots__' not in classdict:
             classdict['__slots__'] = []
         else:
-            all_slots |= set(classdict['__slots__'])
+            all_slots.update(dict.fromkeys(classdict['__slots__']))
 
         classdict['_all_slots_'] = list(all_slots)
 
@@ -3347,10 +3347,8 @@ class ParameterizedMetaclass(type):
         """
         # get all relevant slots (i.e. slots defined in all
         # superclasses of this parameter)
-        slots = {}
         p_type = type(param)
-        for p_class in classlist(p_type)[1::]:
-            slots.update(dict.fromkeys(p_class.__slots__))
+        slots = dict.fromkeys(p_type._all_slots_)
 
         # note for some eventual future: python 3.6+ descriptors grew
         # __set_name__, which could replace this and _set_names
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1138,7 +1138,7 @@ class Parameter(_ParameterBase):
     __slots__ = ['name', 'default', 'doc',
                  'precedence', 'instantiate', 'constant', 'readonly',
                  'pickle_default_value', 'allow_None', 'per_instance',
-                 'watchers', 'owner', '_label', '_unbound']
+                 'watchers', 'owner', '_label']
 
     # Note: When initially created, a Parameter does not know which
     # Parameterized class owns it, nor does it know its names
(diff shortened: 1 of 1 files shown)


param/parameterized.py
```diff
@@ -403,19 +403,14 @@ def iscoroutinefunction(function):
...
+    finally:
+        paramobj.watchers = watchers
     p.owner = owner
-
-    # Reset watchers since class parameter watcher should not execute
-    # on instance parameters
-    p.watchers = {}
-
-    # shallow-copy any mutable slot values other than the actual default
-    for s in p.__class__.__slots__:
```
philippjfr, line 415: It seemed a little silly to set slot values twice (once in `__setstate__` and once here), but I think I should revert this.

param/parameterized.py
```diff
@@ -897,25 +898,35 @@ class ParameterMetaclass(type):
...
 
         # when asking for help on Parameter *object*, return the doc slot
-        classdict['__doc__']=property(attrgetter('doc'))
+        classdict['__doc__'] = property(attrgetter('doc'))
+
+        # Compute all slots
+        all_slots = set()
+        for base in bases:
+            for bcls in base.__mro__:
+                for slot in getattr(bcls, '__slots__', []):
```
Hoxbro, line 914: I think it would be faster if this `for` statement is skipped by converting directly to a set and updating `all_slots`.

philippjfr:
Added a few more micro-optimizations and dropped backward compatibility for unpickling objects from before watchers were introduced (i.e. v1.8.0), which is 5 years ago now.

param/parameterized.py
```diff
@@ -403,14 +404,15 @@ def iscoroutinefunction(function):
...
 
-    # Shallow-copy Parameter object
-    p = copy.copy(paramobj)
+    # Shallow-copy Parameter object without the watchers
+    try:
+        watchers = paramobj.watchers
+        paramobj.watchers = {}
+        p = copy.copy(paramobj)
+    finally:
+        paramobj.watchers = watchers
```
jbednar, line 413: This obscure code coming back deserves a clear comment explaining why it's done this way...

param/parameterized.py
```diff
@@ -1318,29 +1317,29 @@ def _set_instantiate(self,instantiate):
...
-            if slot_attribute:
+        if attribute == 'name':
+            name = getattr(self, 'name', None)
+            if name is not None and value != name:
+                raise AttributeError("Parameter name cannot be modified after "
+                                     "it has been bound to a Parameterized.")
+
+        is_slot = attribute in self.__class__._all_slots_
+        has_watcher = attribute != "default" and attribute in getattr(self, 'watchers', [])
+        if not (is_slot or has_watcher):
```
Hoxbro, line 1328: We could use the walrus operator here so we exit earlier and do not need the calculation of `has_watcher`. 

This is only possible if min. version is Python 3.8.

jbednar: py38 being four years old at this point, I'm happy with that as a version floor for Param 2.0, if that's ok for Panel purposes.

maximlt: @Hoxbro how would you use the walrus operator here?

I believe this could be rewritten as follows. But  I'm happy with how it is right now and haven't run any benchmark to see whether it'd make any difference.

```python
if not is_slot:
    has_watcher = attribute != "default" and attribute in getattr(self, 'watchers', [])
    if not has_watcher:
        # Return early if attribute is not a slot
        return super().__setattr__(attribute, value)
```

Oh and the minimum Python version for Param is 3.8 already :)

maximlt (commented):
LGTM! The numbers look great and some of the refactoring makes the code even more understandable 👍 I just have a question in a comment.

param/parameterized.py
```diff
@@ -897,25 +898,33 @@ class ParameterMetaclass(type):
...
         # the base Parameter class.  That's because a subclass will have
         # a __dict__ unless it also defines __slots__.
         if '__slots__' not in classdict:
-            classdict['__slots__']=[]
+            classdict['__slots__'] = []
+        else:
+            all_slots.update(dict.fromkeys(classdict['__slots__']))
 
-        # No special handling for a __dict__ slot; should there be?
+        classdict['_all_slots_'] = list(all_slots)
```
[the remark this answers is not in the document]
philippjfr: Fastest way to get an ordered set in python.

param/parameterized.py
Hoxbro: I didn't put that many thoughts into it. 

Looking at this again. I don't think it is possible to do as both variables are needed if we don't exit early.

Your example will give an `UnboundLocalError` and similar with the Walrus operator code change.

maximlt: Ah indeed `has_watcher` is re-used below, my bad!

param/parameterized.py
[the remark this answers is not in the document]
maximlt: Thanks that's what I thought but I wanted to be sure, I've added a comment in the code.

PR #840, opened 2023-09-23
Fix multi-level indirection in Parameters access

philippjfr:
Started looking at `.param` and noticed this really bizarre (and pointless) double indirection when accessing variables on the `Parameters` object. There's a bunch of cases where instead of accessing methods directly on the `Parameters` object we go through this two-level indirect attribute access that just to get back to the same object, i.e. `self_.self_or_cls.param._TRIGGER` simplifies to just `self_._TRIGGER`.

In fact this is so silly that fixing it gives a 10% performance boost for a bunch of common operations:

Screen Shot 2023-09-23 at 22 20 22

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1529,7 +1529,7 @@ def as_uninitialized(fn):
         parameterized_instance = self_.self
         original_initialized = parameterized_instance._param__private.initialized
         parameterized_instance._param__private.initialized = False
-        fn(parameterized_instance, *args, **kw)
+        fn(self_, *args, **kw)
         parameterized_instance._param__private.initialized = original_initialized
     return override_initialization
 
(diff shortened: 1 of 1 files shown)


jbednar (commented):
Lgtm. I'm not sure if these are the same bits where I was noticing a similar issue, but in that case I couldn't tell whether the indirection was because of trying to write code that works both with a class and with an instance, or whether it was just left over from originally moving code into the param object originally and not updating it. In any case the solution is the same, i.e. make sure we have tests covering both cases and then it's fine. Thanks!

philippjfr:
> I'm not sure if these are the same bits where I was noticing a similar issue, but in that case I couldn't tell whether the indirection was because of trying to write code that works both with a class and with an instance, or whether it was just left over from originally moving code into the param object originally and not updating it.

I'm pretty sure the indirection was always pointless and was introduced in an unnecessary attempt to make it work for both classes and instances, i.e. `.self_or_cls.param` by definition always takes you back to the same `Parameters` object you're already on.

PR #841, opened 2023-09-24
Updated reactive docs

jbednar:
I think this will take several iterations, but I've tried to tease out a better motivating introductory section without making too many changes. @philippjfr , please review and see if it's going in a direction you're happy with; feel free to close if not.

I suspect that it's still trying to be too ambitious, and that we shouldn't immediately dive in trying to change the dataframe itself, because that's an advanced topic using `bind`, while if we simply make the dataframe and some values reactive, I think the user will be able to see how it works much better. If I get a chance I'll try that out, but wanted you to have this before I went to bed in case you work on it tomorrow.

I also think the example still isn't that great without widgets. I do like what it does for the loop I put in there now, which substitutes for having a slider; I think there's something we can do with that. E.g. maybe some sort of animation? 

Plus the material after the "Why?" section is still starting with Parameters, while I think it would be better to focus on what a reactive expression _is_ and how it  behaves, and then get into Parameters only as one way of getting such an expression, as well as an implementation detail behind them. Anyway, not urgent, just handing this over in case it's useful!

### DIFF
--- a/doc/user_guide/Reactive_Expressions.ipynb
+++ b/doc/user_guide/Reactive_Expressions.ipynb
@@ -7,7 +7,7 @@
    "source": [
     "## Reactive Functions & Expressions\n",
     "\n",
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks, I honestly just wanted to get something in before the RC release. This may well still be too ambitious but it's definitely an improvement.

PR #843, opened 2023-09-24
Implement support for allowing Parameter references

philippjfr:
This PR brings an exceptionally useful feature that we have had in Panel since 1.2 to Param. Specifically it allows passing Parameter references (i.e. Parameter objects, reactive functions and expressions) as a parameter value and then automatically mirrors the value of the reference, e.g.:

```python
class Parameters(param.Parameterized):

    string = param.String(default="string", allow_refs=True)
 
p = Parameters()
p2 = Parameters(string=p.param.string)

assert p.string == p2.string
p.string = 'new_string'
assert p2.string == 'new_string'
```

This is much, much easier than setting up callbacks and together with the addition of reactive functions and expressions makes it super easy to set up parameters that depend on some other dynamically computed values. The implementation here is fully backward compatible, since you must enable the support for resolving references like this explicitly with the new `Parameter.allow_refs` slot.

Note that references can only be used on instances, classes will not attempt to resolve references, however you may use a class parameter as a reference.

This implementation also implements updating of references, which could not reasonably be implemented in Panel and is a good reason to move this support to Param.

In addition to allowing simple references to be resolved support can also be enabled for nested references using the `Parameter.nested_refs` slot. This allows constructs such as this to work:

```python
class Parameters(param.Parameterized):

    string = param.String(default="string", allow_refs=True)

    dictionary = param.Dict(default={}, allow_refs=True, nested_refs=True)

p = Parameters()
p2 = Parameters(dictionary={'key': p.param.string})
p3 = Parameters(string='foo')

assert p2.dictionary == {'key': 'string'}
p2.dictionary = {'new key': p3.param.string}
assert p2.dictionary == {'new key': 'foo'}
```

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -8,16 +8,18 @@ either alone (providing basic Parameter support) or with param's
 __init__.py (providing specialized Parameter types).
 """
 
+import asyncio
 import copy
 import datetime as dt
-import re
+import html
 import inspect
-import random
 import numbers
 import operator
+import random
+import re
+import types
 import typing
 import warnings
-import html
 
 # Allow this file to be used standalone if desired, albeit without JSON serialization
 try:
@@ -31,6 +33,7 @@ from html import escape
 from itertools import chain
 from operator import itemgetter, attrgetter
 from types import FunctionType, MethodType
+from typing import Generator
 
 import logging
 from contextlib import contextmanager
@@ -642,6 +645,33 @@ def _skip_event(*events, **kwargs):
     return True
 
 
+def extract_dependencies(function):
+    """
+    Extract references from a method or function that declares the references.
+    """
+    subparameters = list(function._dinfo['dependencies'])+list(function._dinfo['kw'].values())
+    params = []
+    for p in subparameters:
+        if isinstance(p, str):
+            owner = get_method_owner(function)
+            *subps, p = p.split('.')
+            for subp in subps:
+                owner = getattr(owner, subp, None)
+                if owner is None:
+                    raise ValueError('Cannot depend on undefined sub-parameter {p!r}.')
+            if p in owner.param:
+                pobj = owner.param[p]
+                if pobj not in params:
+                    params.append(pobj)
+            else:
+                for sp in extract_dependencies(getattr(owner, p)):
+                    if sp not in params:
+                        params.append(sp)
+        elif p not in params:
+            params.append(p)
+    return params
+
+
 # Two callers at the module top level to support pickling.
 async def _async_caller(*events, what='value', changed=None, callback=None, function=None):
     if callback:
@@ -1032,7 +1062,7 @@ class Parameter(_ParameterBase):
     __slots__ = ['name', 'default', 'doc',
                  'precedence', 'instantiate', 'constant', 'readonly',
                  'pickle_default_value', 'allow_None', 'per_instance',
-                 'watchers', 'owner', '_label']
+                 'watchers', 'owner', 'allow_refs', 'nested_refs', '_label']
 
     # Note: When initially created, a Parameter does not know which
     # Parameterized class owns it, nor does it know its names
@@ -1045,7 +1075,7 @@ class Parameter(_ParameterBase):
     _slot_defaults = dict(
         default=None, precedence=None, doc=None, _label=None, instantiate=False,
         constant=False, readonly=False, pickle_default_value=True, allow_None=False,
-        per_instance=True
+        per_instance=True, allow_refs=False, nested_refs=False
     )
 
     # Parameters can be updated during Parameterized class creation when they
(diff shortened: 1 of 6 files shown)


param/parameterized.py
```diff
@@ -1779,15 +1830,112 @@ def _setup_params(self_, **params):
...
+            if ref is not None:
+                refs[name] = ref
+                deps[name] = ref_deps
             setattr(self, name, val)
+        return refs, deps
+
+    def _setup_refs(self_, refs):
+        groups = defaultdict(list)
+        for pname, subrefs in refs.items():
+            print(pname, subrefs)
```
philippjfr, line 1860:
```suggestion
```

jbednar (approved):
I haven't used references in Panel code, so I'm not the best person to review this, but it looks reasonable to me. Having to declare nested references seems ok to me, since it's difficult to reason about all the implications involved, but having to declare a simple reference to be allowed seems like it will greatly limit how useful this is, because most people won't add that keyword and those that do will make their codebase get polluted with a lot of `allow_refs=True`.  My inclination is `allow_refs=True` and `nested_refs=False`, but I don't know how likely that is to cause problems with current code.

philippjfr:
I'm a little wary about changing that this close to release. How would you feel about the idea that in 2.0 we begin warning people that they should explicitly set `allow_refs=False` when we detect something that would be treated as a reference if we explicitly enabled it. Then we can give people some time to update their code and then flip the switch by default in 2.1.

jbednar:
Ok, though I'd probably not announce 2.0 widely until 2.1 is in place, leaving the 2.0 as a helpful interim step for people to move their code into compatibility, rather than one we'd expect people to adopt widely. 2.0 already has compatibility implications and I'd like to guide people from where they are into the new way, rather than make them adapt twice.

param/parameterized.py
```diff
@@ -1529,8 +1677,9 @@ def override_initialization(self_,*args,**kw):
         parameterized_instance = self_.self
         original_initialized = parameterized_instance._param__private.initialized
         parameterized_instance._param__private.initialized = False
-        fn(self_, *args, **kw)
+        ret = fn(self_, *args, **kw)
```
Hoxbro, line 1680: What if fn errors here? Should we try/except?

param/parameterized.py
```diff
@@ -1779,15 +1928,139 @@ def _setup_params(self_, **params):
...
+                # Until Parameter.allow_refs=True by default we have to
+                # speculatively evaluate a values to check whether they
+                # contain a reference and warn the user that the
+                # behavior may change in future.
+                if name not in self_.cls._param__private.explicit_no_refs:
+                    try:
+                        ref, _, resolved, _ = self_._resolve_ref(pobj, val)
+                    except Exception:
+                        ref = None
+                    if ref:
```
Hoxbro, line 1952: Just to be sure ref from `self._._resolve_ref` can never be falsey?

param/parameterized.py
```diff
@@ -1779,15 +1928,139 @@ def _setup_params(self_, **params):
...
+                        warnings.warn(
+                            f"Parameter {name!r} on {pobj.owner} is being given a valid parameter "
+                            f"reference {val} but is implicitly allow_refs=False. "
+                            "In future allow_refs will be enabled by default and "
+                            f"the reference {val} will be resolved to its underlying "
+                            f"value {resolved}. Please explicitly set allow_ref on the "
+                            "Parameter definition to declare whether references "
+                            "should be resolved or not.",
+                            category=_ParamFutureWarning,
+                            stacklevel=2,
```
Hoxbro, line 1962: Have you checked if the stack level is correct?

param/parameterized.py
```diff
@@ -1779,15 +1928,139 @@ def _setup_params(self_, **params):
...
+            if pobj is None or not pobj.allow_refs:
+                # Until Parameter.allow_refs=True by default we have to
+                # speculatively evaluate a values to check whether they
+                # contain a reference and warn the user that the
+                # behavior may change in future.
+                if name not in self_.cls._param__private.explicit_no_refs:
+                    try:
+                        ref, _, resolved, _ = self_._resolve_ref(pobj, val)
+                    except Exception:
+                        ref = None
```
Hoxbro, line 1951: `resolved` is also needed here.

param/parameterized.py
philippjfr: No, if it errors that should be raised, could try/except to remove one level from the stack trace but that seems hairy.

param/parameterized.py
philippjfr: Good catch, I haven't yet.

philippjfr: Fixed.

philippjfr:
I'm going to go ahead and merge. As I need this in a dev release asap the timelines here are tight and this was written with backward compatibility at the forefront. If you're coming at this later please still give it a thorough review.

PR #844, opened 2023-09-24
Rename param.reactive to param.rx

philippjfr:
In playing around with `reactive` @jbednar and I both found that it would actually be nicer if it was just called `rx`. It's shorter and once you're familiar it's quite readable. The other benefit is that it won't clash with `panel.reactive`, which means it can be re-exported on the `pn` namespace.


doc/user_guide/Reactive_Expressions.ipynb
```diff
@@ -515,7 +515,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "string_template = param.reactive('Hello {name}!')\n",
```
jbednar, line 518: I'm in favor of the simpler and more readable "rx()", not "param.rx()", using `from param import rx` as already being done above.

param/reactive.py
```diff
@@ -1,30 +1,30 @@
 """
 reactive API
 
-`reactive` is a wrapper around a Python object that lets users create
+`rx` is a wrapper around a Python object that lets users create
 reactive pipelines by calling existing APIs on an object with dynamic
 parameters or widgets.
```
jbednar, line 6:
```suggestion
`rx` is a wrapper around a Python object that lets users create
reactive expression pipelines by calling existing APIs on an object with dynamic
parameters or widgets.
```

(to underscore what the `x` stands for)

param/reactive.py
```diff
@@ -1,30 +1,30 @@
...
 
-A `reactive` instance watches what operations are applied to the object
+A `rx` instance watches what operations are applied to the object
```
jbednar, line 8:
```suggestion
An `rx` instance watches what operations are applied to the object
```

param/reactive.py
```diff
@@ -1,30 +1,30 @@
...
 and records these on each instance, which are then strung together
 into a chain.
 
-The original input to a `reactive` is stored in a mutable list and can be
+The original input to a `rx` object is stored in a mutable list and can be
```
jbednar, line 12:
```suggestion
The original input to an `rx` object is stored in a mutable list and can be
```

param/reactive.py
```diff
@@ -1,30 +1,30 @@
...
 accessed via the `_obj` property. The shared mutable data structure
-ensures that all `reactive` instances created from the same object can
+ensures that all `rx` instances created from the same object can
 hold a shared reference that can be updated, e.g. via the `.set`
 method or because the input was itself a reference to some object that
 can potentially be updated.
 
-When an operation is applied to a `reactive` instance, it will
+When an operation is applied to a `rx` instance, it will
```
jbednar, line 19:
```suggestion
When an operation is applied to an `rx` instance, it will
```

param/reactive.py
```diff
@@ -1,30 +1,30 @@
...
 record the operation and create a new instance using `_clone` method,
 e.g. `dfi.head()` first records that the `'head'` attribute is
 accessed, which is achieved by overriding `__getattribute__`. A new
 reactive object is returned, which will then record that it is
 being called, and that new object will be itself called as
```
jbednar, line 24:
```suggestion
being called, and that new object will be itself called, as
```

jbednar (approved):
Yay!

PR #846, opened 2023-09-24
Fix implementation of rx.__iter__

philippjfr:
The previous implementation was trying to be too clever. This implementation is simpler but achieves the same thing.

### DIFF
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -941,20 +941,9 @@ class rx:
             return
         elif not isinstance(self._current, Iterable):
             raise TypeError(f'cannot unpack non-iterable {type(self._current).__name__} object.')
-        iterator = []
-        def iterate(value):
-            if iterator:
-                iterate = iterator[0]
-            else:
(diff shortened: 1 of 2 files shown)


param/reactive.py
```diff
@@ -941,20 +941,9 @@ def __iter__(self):
...
-            try:
-                yield next(iterate)
-            except StopIteration as e:
-                iterator.clear()
-                raise e
-        for item in self._apply_operator(iterate):
-            yield item
+        items = self._apply_operator(list)
+        for i in range(len(self._current)):
+            yield items[i]
```
Hoxbro, line 946: Why not?

```suggestion
        for item in self._apply_operator(list):
            yield item
```

or even 

```suggestion
        for item in self._apply_operator(iter):
            yield item
```

(I have no idea how to test this, so it is mainly a curious question)

philippjfr: Good question but apply_operator returns another reactive expression so this would just result in recursion.

PR #848, opened 2023-09-25
Remove warnings related to explicitly setting allow_refs=False

philippjfr:
The speculative evaluation of references was causing issues because it evaluates functions when it shouldn't have.


maximlt:
@philippjfr should this be merged in Param 2.0?

philippjfr:
I think I've decided against merging it.

Issue #648, opened 2022-08-06
Refactor __init__.py according to modern Python and best practices


MarcSkovMadsen:
I see some issues with the main `__init__.py` file that makes using Param and contributing to it harder than it has to be.

- The`__init__.py` has grown to +2000 lines. What I normally see as best practice is to say maximum 1000 lines to make the file understandable. And preferably max around 250 lines.
- A `__init__.py` is not meant to contain a lot of the implementation details. A *best practice* `__init__.py` should
  - expose a module docstring which is useful for users
  - import the *public* classes, functions and other objects.
  - define an `__all__` method which guides *wild card imports* (if anyone uses that) and editors like VS Code based on static analysis to the "public" functionality.

My request is a refactoring guided by the best practices above to help users and future contributors.

jbednar:
Defining `__all__` would be useful. One could move the entire contents of `__init__.py` to its own file `parameters.py`, but that would add an additional file, and I'm not sure I see what the benefit would be. Who defines `is not meant to`?

philippjfr:
People generally cite this blog post:

>Leaving an __init__.py file empty is considered normal and even a good practice, if the package’s modules and sub-packages do not need to share any code. [Source](https://python-docs.readthedocs.io/en/latest/writing/structure.html)

and it is the consensus of most Python devs if you survey StackOverflow answers, reddit posts etc. `__init__.py` is really meant to be used for initialization code and declaring imports and defining `__all__`, here's what the Python docs say:

>In the simplest case, __init__.py can just be an empty file, but it can also execute initialization code for the package or set the __all__ variable, described later.

On balance I agree we should refactor param, the file is unwieldy and there's no reason for it really. If param was still small enough to be a single file I'd favor just dropping it all into `__init__.py` but that's not the case so I don't see any reason not to have better organization.

maximlt:
Indeed, because there's a lot of code in `__init__.py` the auto-completion suggests many irrelevant objects like `param.sys` or `param.OrderedDict`. Moving that code to another module would help. Or importing them *as* private objects (`import os as _os`). I'd probably prefer moving the code to another module, since anyway Param is already two files with `__init__.py` and `parameterized.py`, except if you're only interested in the `String` parameter :D

jbednar:
[...]

PR #852, opened 2023-09-27
Refactor `__init__.py`, mostly by moving Parameters to `parameters.py`

maximlt:
Fixes https://github.com/holoviz/param/issues/648

This is also adding `__all__` to the `__init__.py`.

When you start to touch so many lines it's tempting to do some formatting and even more refactoring, styling, imports sorting, etc.. I've decided to stick to address the limitations exposed in the issue; this can be done later.

I think Github supports ignoring specific commits on *blame*, I'll enable that once this is merged.


jbednar:
Thanks. The large-scale moves here make sense, but are hard to review, so I'll just trust you didn't insert anything beyond what was already there. :-)  

Are there special techniques to preserve `git blame` in a case like this? E.g. if the git history includes `git mv __init__.py parameters.py` followed by recreating `__init__.py`, will github know that the history of the classes in `parameters.py` can be found in older `__init__.py` versions? Now would be the time to be careful about that, and I don't know how best to achieve that.

PR #853, opened 2023-09-27
Refactor `__init__.py`, mostly by moving Parameters to `parameters.py` (take 2)

maximlt:
Fixes https://github.com/holoviz/param/issues/648
Supersedes https://github.com/holoviz/param/pull/852

This is also adding `__all__` to the `__init__.py`.

When you start to touch so many lines it's tempting to do some formatting and even more refactoring, styling, imports sorting, etc.. I've decided to stick to address the limitations exposed in the issue; this can be done later.

I think Github supports ignoring specific commits on *blame*, I'll enable that once this is merged.

Per Jim's comment in https://github.com/holoviz/param/pull/852#issuecomment-1736534254

> Are there special techniques to preserve git blame in a case like this? E.g. if the git history includes git mv __init__.py parameters.py followed by recreating __init__.py, will github know that the history of the classes in parameters.py can be found in older __init__.py versions? Now would be the time to be careful about that, and I don't know how best to achieve that.

I have used `git mv __init__.py parameters.py` in this PR (had never used it!) which can only help `git`. Now we usually squash commits when merging to *main* and I don't know to which extent that will confuse `git blame`. I guess the next step for me is to try that locally.


maximlt:
I squashed the commits locally merging to main, created another branch from that that I pushed to the repo. Browsing to the `parameters.py` file and enabling blame, Github doesn't show the history at all https://github.com/holoviz/param/blame/debugmain/param/parameters.py. However, Github does show the history when enabling blame on `parameters.py` from this branch (refactor_init_all2) https://github.com/holoviz/param/blame/refactor_init_all2/param/parameters.py.

So it seems **we should not squash** the commits of this PR when merging it into *main*!

On Github we have the following options when merging:

1. Create a merge commit (disabled currently on the repo but could be enabled temporarily)
2. ~~Squash and merge~~
3. Rebase and merge

I tried 1) and 3), both options allow to preserve blame for `parameters.py`: https://github.com/holoviz/param/blame/debugmain3/param/parameters.py and https://github.com/holoviz/param/blame/debugmain2/param/parameters.py.

So that's all good news, we have a way to preserve blame 👍 Thanks a lot @jbednar for catching that, it would have been very annoying loosing blame. Do you have a preference between the two available options?

jbednar:
Ok, great! I strongly prefer option 3 (rebasing), because it keeps the revision history linear on main. Thanks for following up on that!

maximlt:
Ok thanks I'm going to rebase and merge this PR then.

Issue #727, opened 2023-04-05
Document Param 2.0 breaking changes


maximlt:
Listing the breaking changes made in Param 2.0 to list them in the release notes:
- https://github.com/holoviz/param/pull/726
- https://github.com/holoviz/param/pull/729
- https://github.com/holoviz/param/pull/605
- https://github.com/holoviz/param/pull/744
- https://github.com/holoviz/param/pull/767
- https://github.com/holoviz/param/pull/797
- https://github.com/holoviz/param/pull/796
- and more that I missed for sure

philippjfr:
Not a breaking change but an announcement of a future breaking change:

- Param will enable `allow_refs` in future so any objects that are references, e.g. functions, parameters, reactive expressions and in the case of Panel widgets, will automatically resolve to their value when passed to a parameter. To disable this users will have to set `allow_refs=False` explicitly.

PR #855, opened 2023-09-27
Add release notes for Param 2.0

maximlt:
Fixes https://github.com/holoviz/param/issues/727

Well, these were not the easiest release notes to put together :D Marking this as draft for now as there's at least one last PR to merge and the reactive API being still a little in flight I wouldn't be surprised if they were more to come.


maximlt:
Updated after v.2.0.0rc6, includes now references support.

PR #862, opened 2023-10-02
Add docs around allow_refs feature

philippjfr:
Documents `allow_refs` and `nested_refs`.

### DIFF
--- a/doc/user_guide/How_Param_Works.ipynb
+++ b/doc/user_guide/How_Param_Works.ipynb
@@ -239,6 +239,125 @@
   },
   {
    "cell_type": "markdown",
+   "id": "b86f55e3-5ae5-4879-9b4c-273e859b6fec",
+   "metadata": {},
+   "source": [
+    "## References\n",
+    "\n",
+    "The custom attribute access allows accessing the concrete value of a `Parameter` but via the `.param` namespace we can also access the underlying Parameter value. In many scenarios the `Parameter` object can be treated as a reference that represents the current value of that parameter. Specifically we can tell a `Parameter` that it should `allow_refs`, i.e. it should be able to accept a reference to another `Parameter` and resolve it dynamically, ensuring that the two `Parameter` values remain synced:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fba9fbdc-0776-4cd3-8250-e00487593d57",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class B(Parameterized):\n",
+    "    \n",
+    "    p = Parameter(default=1, allow_refs=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "27beb80b-415f-41cb-b909-adb6822e7e9a",
+   "metadata": {},
+   "source": [
+    "Having declared a `Parameter` that allows references we can now pass the parameter `p` of `b1` to parameter `p` of `b2`:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "22a51c29-c286-4e0f-b1a6-1e02bc4149da",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "b1 = B(p=14)\n",
+    "b2 = B(p=b1.param.p)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5eab5e13-66df-4001-8d70-2cab70449641",
+   "metadata": {},
+   "source": [
+    "Inspecting `b2.p` we will see that `p` of `b2` now reflects the value of `b1.p`:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fdc58565-f327-405f-958d-b49caa70adb3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "b2.p"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7322e310-9105-48c9-a0b8-41cb80bcf4e6",
+   "metadata": {},
+   "source": [
+    "Even when we update the value of `b1.p` the value of `b2.p` will reflect the change:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "961aebc0-45b5-458d-919c-1ba6a3c286e7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "b1.p = 7\n",
+    "\n",
+    "b2.p"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "75c5eb72-3132-458c-9b30-b5cd1e64b63d",
+   "metadata": {},
+   "source": [
+    "If we explicitly set `b2.p` however the two values will become unsynced:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "2e02cb21-a4ab-48fb-a516-da17feaf0da0",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "b2.p = 3\n",
+    "\n",
+    "print(b1.p, b2.p)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "701e26c3-6eb0-4c76-b453-cc37be06ac41",
+   "metadata": {},
+   "source": [
+    "and any subsequent changes to `b1.p` will not be reflected by `b2`:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "50aedbea-7902-4f1a-8e98-6866a965eda8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "b1.p = 27\n",
+    "\n",
+    "b2.p"
+   ]
+  },
+  {
+   "cell_type": "markdown",
    "id": "c3e3d3aa",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


doc/user_guide/Parameters.ipynb
```diff
@@ -784,6 +784,156 @@
...
+    "\n",
+    "v.b"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "324475cf-65a3-4902-8508-57af19697b5f",
+   "metadata": {},
+   "source": [
+    "By declaring that `V.b` allows references we have made it possible to pass the Parameter `U.b`, which means `v.b` will reflect the value of `u.a`:"
```
maximlt, line 823: I think this is incorrect?

```suggestion
    "By declaring that `V.b` allows references we have made it possible to pass the Parameter `U.a`, which means `v.b` will reflect the value of `u.a`:"
```

doc/user_guide/Parameters.ipynb
```diff
@@ -784,6 +784,156 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "2579c782-8dfd-4ded-995f-dc8498f4d275",
+   "metadata": {},
+   "source": [
+    "## Parameter references\n",
+    "\n",
+    "Having looked at the difference between the `Parameter` value and the `Parameter` object let us consider the idea that a `Parameter` is a reference or proxy for its underlying value. Parameters may be declared to `allow_refs` which effectively means that they will accept other Parameter **objects** (and other valid references) and resolve and reflect their current values."
```
MarcSkovMadsen, line 794: Why is `allow_parameters` not a more "natural" name than `allow_refs`?

MarcSkovMadsen: Its almost impossible for me to envision what `allow_refs` means. How should I think about it? What problem does it solve?

MarcSkovMadsen: Is it really a way of specifying whether the value is the value given (like a parameter) or the value the given value references (like the value of a parameter)?

philippjfr: Panel has had internal support for `allow_refs` for quite a while (since 1.1.0) so you've almost certainly seen it and used it. The reason why it's not called `allow_parameters` is that it's not accurate. `refs` can be a number of things including parameters, bound functions, functions annotated with depends, reactive expressions, and even async generators. It's documented [here](https://panel.holoviz.org/how_to/interactivity/bind_component.html) and to some extent [here](https://panel.holoviz.org/how_to/interactivity/hvplot_interactive.html).

philippjfr: It's also not something most users will ever have to think about as it's primarily useful for Panel developers and quite advanced Panel users (i.e. folks writing Panel apps in classes).

philippjfr: >Is it really a way of specifying whether the value is the value given (like a parameter) or the value the given value references (like the value of a parameter)?

That's exactly right.

doc/user_guide/Parameters.ipynb
```diff
@@ -784,6 +784,156 @@
...
+   "metadata": {},
+   "source": [
+    "Note that `Parameter` types are not the only types of valid references. The full list of valid references include:\n",
+    "\n",
+    "- Class and instance `Parameter` objects\n",
+    "- Functions or methods annotated with `param.depends`\n",
+    "- Functions wrapped with `param.bind`\n",
+    "- Reactive expressions declared using `param.rx`\n",
+    "- Asynchronous generators\n",
+    "- Custom objects transformed into a valid reference with a hook registered with `param.parameterized.register_reference_transform`.\n",
```
maximlt, line 920: Maybe add `param.parameterized.register_reference_transform` to `doc/reference/parameterized_helpers.md`?

Generally we should try (after the release!) to better document how advanced users can extend Param and leverage it (e.g. what can I do with a method decorated with `@param.depends`), and separate those sections with those that are meant for normal Param users.

doc/user_guide/Parameters.ipynb
```diff
@@ -784,6 +784,156 @@
...
+   "id": "ae2d21b3-c78b-4648-8622-282c237a1d64",
+   "metadata": {},
+   "source": [
+    "Note that `Parameter` types are not the only types of valid references. The full list of valid references include:\n",
+    "\n",
+    "- Class and instance `Parameter` objects\n",
+    "- Functions or methods annotated with `param.depends`\n",
+    "- Functions wrapped with `param.bind`\n",
+    "- Reactive expressions declared using `param.rx`\n",
+    "- Asynchronous generators\n",
```
maximlt, line 919: Allowing bare asynchronous generators isn't going to cause problems with `param.Callable`? Is it the kind of case were users will have to set `allow_refs=False` and won't have the option between passing a normal async generator or a "reference async generator"?
Super edge-case I know, it's just to understand the consequences.

doc/user_guide/Parameters.ipynb
```diff
@@ -784,6 +784,156 @@
...
+    "- Custom objects transformed into a valid reference with a hook registered with `param.parameterized.register_reference_transform`.\n",
+    "\n",
+    "There are two utility functions which allow resolving the value of a reference and all parameters the reference depends on:"
```
maximlt, line 922: The order in the sentence (value/reference) is inverted compared to the order of the example below.

doc/user_guide/Parameters.ipynb
```diff
@@ -784,6 +784,156 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8cb30e85-5cfe-49d6-a641-f98c84a7acb2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from param.parameterized import resolve_ref, resolve_value\n",
```
maximlt, line 932: Maybe add `resolve_ref`, `resolve_value`to `doc/reference/parameterized_helpers.md`?

maximlt (approved):
LGTM!

In addition to the few comments I left:

- Did you think about adding `allow_refs` and `nested_refs` to the beginning of `Parameters.ipynb` where all the common Parameter slots are listed?
- It would be nice to demonstrate that passing references allows to avoid writing callbacks. Also it'd be nice to see you can watch a Parameter that's been set from a reference, to see that updating the reference does execute the callback (I had to try to make sure :D!). Param docs probably need some refactoring!

doc/user_guide/Parameters.ipynb
philippjfr: It is going to cause problems if you enable `allow_refs=True` but for a callable there's rarely ever a good reason to do so.

doc/user_guide/Parameters.ipynb
philippjfr: Done

doc/user_guide/Parameters.ipynb
philippjfr: The deeper point though is that currently we don't have a way to allow some references and not others, which I guess is a problem.

philippjfr: The trick I originally used in Panel was that I would first check if a value would pass `_validate` before I attempted to resolve the reference, the issue there is that `Dynamic` parameters (including `Number`) allow functions/generators and you want to allow references even on `param.Parameter` types.

maximlt: > It is going to cause problems if you enable allow_refs=True but for a callable there's rarely ever a good reason to do so.

Is `allow_refs` set to True for `pn.viewable.Viewer` objects?

maximlt:
Thanks for addressing my comments, feel free to merge!

philippjfr:
Haven't addressed your last comments. So will do that, merge and then cut an RC.

doc/user_guide/Parameters.ipynb
droumis, line 1: Here's my suggestion for new intro text, including a more descriptive header:

## Dynamically linking parameters

Building on the previous discussion about the dual roles of a `Parameter` - as both a value holder and a metadata container - let's explore how parameters can go a step further by acting as dynamic references to other parameters. When configured with `allow_refs=True`, a `Parameter` can serve as a live link to another `Parameter`, mirroring its current value. This capability enables more intricate relationships between parameters, allowing for automatic value synchronization and forming the basis for reactive programming.

droumis, line 1: I don't understand how the example used to demonstrate nested_refs is an example of nesting; `u1` and `u2` don't contain any references...?

I would have expected something more like this:

```python
class U(param.Parameterized):
    a = param.Number()

class V(param.Parameterized):
    b = param.Number(allow_refs=True)

class W(param.Parameterized):
    c = param.Number(allow_refs=True, nested_refs=True)

u = U(a=5)
v = V(b=u.param.a)
w = W(c=v.param.b)
```

droumis (approved):
I left some comments and suggestions. Besides the nesting example, the rest was mostly about wordsmithing. Feel free to take or leave any of it and then merge.

doc/user_guide/Parameters.ipynb
philippjfr: This is what demonstrates nesting, i.e. the references are nested inside some other container:

```
w = W(c=[u1.param.a, u2.param.a]) 
``

philippjfr: Chaining references like you have done doesn't require any special support.

doc/user_guide/How_Param_Works.ipynb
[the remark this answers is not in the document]
maximlt: > Any guards against circular referencing?

@philippjfr what is your take on bi-directionally linking Parameters in Param? I'd guess there's a way for this to be implemented, if so let's open an issue to discuss the API? I'd love Param 2.1 to have it to be able to deprecate Panel's `.link`.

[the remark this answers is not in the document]
philippjfr: Indeed, would like to see the same.

PR #863, opened 2023-10-02
Run tests on Python 3.12

maximlt:


ianthomas23:
It should be `3.12-dev` rather than `3.12`.

ianthomas23:
for the `setup-python` action I mean.

maximlt:
> It should be `3.12-dev` rather than `3.12`.

Oh I didn't see your message, instead I just tried setting `allow-prereleases` to `true` for the `setup-python` action (https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#allow-pre-releases) and it seems it worked.

maximlt:
It is! Merging.

Issue #872, opened 2023-10-06
2.0.0rc6 - param.ListSelector Issue


jgriessler:
#### ALL software version info
param.2.0.0rc6 (and older 2.0 rc)

#### Description of expected behavior and the observed behavior
param.ListSelector with check_on_set=False incorrectly updates the .objects when assigning an existing or new value to the parameter.  It adds the value as a list to the end of the .objects, even if it already exists. 
Works fine on 1.13.0. 
#### Complete, minimal, self-contained example code that reproduces the issue

```python
import param

class Test(param.Parameterized):
    mylist = param.ListSelector(objects=['a', 'b', 'c'], check_on_set=False,)

mytest = Test()
mytest.mylist = ['a']
print(mytest.param.mylist.objects)
['a', 'b', 'c', ['a']]
```

jgriessler:
I compared the 1.13 and 2.0.rc versions - looks like the ListSelector validation stuff has been restructured and the problem is that the ListSelector uses the Selector._ensure_value_is_in_objects() method although that expects a value, not the list of values (hence it doesn't find a match and just appends the list at the end).

jbednar:
Damn! I could swear we had already addressed that issue while preparing 2.0. Unfortunately I am travelling and cannot easily look into this soon, but hopefully @maximlt can spot the issue and address it (and add another test!).

maximlt:
Hi @jgriessler ! Thanks for trying out the release candidate and reporting this bug, that I can reproduce. We indeed thought we fixed it in https://github.com/holoviz/param/pull/817 but not entirely apparently. Definitely a release blocker!

---

To be honest, we didn't really expect anyone to exert the release candidates! Out of curiosity, and only if you don't mind, I'd have a couple of questions for you. Are you a Panel user trying out the latest version? What's your use case with `ListSelector(check_on_set=False)` ?

jgriessler:
Yes, using latest Panel. The app is just some private fun stuff, not mission critical, though use param quite a bit (and panel for driving the UI,  bokeh mostly for timeseries charts, holoviews, hvplot for adhoc exploration (and pandas, xarray for the data). 

Most if not all of the many classes are based on param.parameterized (I really like the realtime checking, watch-triggers and automatic widget creation via panel etc). 

What I use params for (besides real-time checking and automatic widgets): 
- lots of dependencies (mostly via @param.depends, but also some via param.watch, less often via pn.bind) to refresh data, UI, charts ... and some basic workflow pipelines
[...]

PR #874, opened 2023-10-08
Fix regression in ListSelector with `check_on_set=False`

maximlt:
Fixes https://github.com/holoviz/param/issues/872

I also found a pretty bad bug in `_instantiate_param_obj` where we were looping only through the `__slots__` defined on the Parameter class, ignoring the ones it inherits. Easy mistake, `kls.__slots__` should really return all the slots! This bug led `ListSelector._objects` being shared between a Parameterized class and instance, since the `_objects` slot is declared in a parent class.

```diff
-    for s in p.__class__.__slots__:
+    for s in p.__class__._all_slots_:
        v = getattr(p, s)
        if _is_mutable_container(v) and s != "default":
            setattr(p, s, copy.copy(v))
```

I also made sure that setting `None` with `check_on_set=False` and `allow_None=True` doesn't add `None` to the objects.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -488,7 +488,7 @@ def _instantiate_param_obj(paramobj, owner=None):
     p.watchers = {}
 
     # shallow-copy any mutable slot values other than the actual default
-    for s in p.__class__.__slots__:
+    for s in p.__class__._all_slots_:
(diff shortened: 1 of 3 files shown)


maximlt:
Merging to prepare a new release candidate.

PR #879, opened 2023-11-03
Drop the feather-format test dependency

musicinmybrain:
It is just a wrapper for `pyarrow.feather`, which we can use directly.


maximlt:
LGTM, thanks!

PR #880, opened 2023-11-03
Add “tables” to the tests-deser extra

musicinmybrain:
Enables one additional test.

https://github.com/holoviz/param/blob/b6a285b7205a869b5677b8571fd54c2a5ccc8709/tests/testfiledeserialization.py#L63-L66

https://github.com/holoviz/param/blob/b6a285b7205a869b5677b8571fd54c2a5ccc8709/tests/testfiledeserialization.py#L76

https://github.com/holoviz/param/blob/b6a285b7205a869b5677b8571fd54c2a5ccc8709/tests/testfiledeserialization.py#L205-L213

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,7 @@ tests-deser = [
     "odfpy",
     "feather-format",
     "pyarrow",
+    "tables",
 ]
 tests-examples = [
     "pytest",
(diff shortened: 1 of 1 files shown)


maximlt:
`tables` is actually installed when the CI runs on Linux:
https://github.com/holoviz/param/blob/e9b27813bb57b41df47e90fdaaef2e683626689a/pyproject.toml#L191-L196

And you can verify the test is not skipped:
https://github.com/holoviz/param/actions/runs/6761543745/job/18376521753#step:7:595

It's just that the dependencies declaration is a *little* convoluted in `pyproject.yml`, and that in the end the `tests-deser` isn't used at all. I'll merge your PR though since anyway `tables` should be in `tests-deser`.

Issue #875, opened 2023-10-23
In Windows, `import param` pops up a command line window


nglee:
On Windows GUI, doing `import param` calls [this line](https://github.com/holoviz/param/blob/v1.13.0/param/__init__.py#L46) which, in turn, pops up a command line window to run `git` commands. Our product is a GUI based product in Windows and it is awkward to see a command line window pop up and close. We are hoping that `param` could be updated so that it does not receive version information from git in runtime.

An alternative would be to have a separate `__version__.py` file that is created during package build. The file stores version information. In runtime, `param` will get the version information from `__version__.py` directly without calling `git` command to retrieve version information from server.

maximlt:
Hi, this line is no longer used in Param 2.0 that is meant to be released soon.

This however suggests a bug in `version.py`.

jbednar:
OMG, that's horrible! Popping up a command line window must have made sense for someone, but that's really bizarre. Definitely something we need to understand how to avoid, even for Param 1.x.

maximlt:
Hi @nglee, we have released Param 2.0.1 with a tentative fix for the issue you reported. Could you please try that new version and report whether it fixes your issue? Asking since I haven't been able to reproduce it.

PR #882, opened 2023-11-06
version.py: new process should not create a window

maximlt:
Attempts to address https://github.com/holoviz/param/issues/875

I'm not sure how to reproduce the bug exactly, and I'm not sure this fixes it.


jlstevens:
I didn't know this flag existed and I don't know if it fixes the issue. That said, it sounds like the right thing and I can't see how it would hurt to specify it.

jlstevens (approved):

maximlt:
Ok thanks, let's try that then!

PR #885, opened 2023-11-07
Don't use setuptools_scm if .git does not exists

Hoxbro:
I noticed `setuptools_scm` took too long to load:

`python -X importtime -c "import param" 2> log.txt && tuna log.txt` 

image


jbednar (approved):
Looks useful, thanks!

PR #890, opened 2023-11-12
Speeds up the Path and derived Parameters

### LINKED ISSUE #889: Speed up derived applications like Panel by speeding up Path Parameter
Panel=`main`, param=2.0.1, python=3.10,os=linux

In https://github.com/holoviz/panel/issues/5852 I'm reporting that for a simple app Panel is spending 2.75% of its load time resolve file paths. Upon investigation I can see its because the paths to css files etc is hold by a `Path` parameter. And that implementation resolves a string path to a pathlib Path every time its asked for its value.

image

**Please add some kind of caching to only resolve once**

MarcSkovMadsen:
Closes #889

### DIFF
--- a/param/parameters.py
+++ b/param/parameters.py
@@ -15,6 +15,7 @@ This file contains subclasses of Parameter, implementing specific
 parameter types (e.g. Number), and also imports the definition of
 Parameters and Parameterized classes.
 """
+from __future__ import annotations
 
 import numbers
 import os.path
@@ -30,6 +31,7 @@ import warnings
 
 from collections import OrderedDict
 from contextlib import contextmanager
+from functools import lru_cache
 
 from .parameterized import (
     Parameterized, Parameter, ParameterizedFunction, ParamOverrides, String,
@@ -2598,6 +2600,81 @@ class HookList(List):
 #   - use resolve_path(path_to_file=False) for paths to existing folders to be read,
 #     and normalize_path() for paths to new files to be written.
 
+def _get_default_search_paths()->typing.Tuple[str]:
+    # Its 10x faster to pickle a string than a Path.
+    # Thus we use os.getcwd instead of pathlib.Path.cwd
+    return (os.getcwd(), )
+
+@lru_cache
+def _resolve_path_fast(
+    path: str,
+    search_paths: typing.Tuple[str|pathlib.Path],
+    path_to_file: bool|None=None
+)->str:
+
+    _path = pathlib.Path(os.path.normpath(path))
+    ftype = "File" if path_to_file is True \
+        else "Folder" if path_to_file is False else "Path"
+
+    if _path.is_absolute():
+        if ((path_to_file is None  and _path.exists()) or
+            (path_to_file is True  and _path.is_file()) or
+            (path_to_file is False and _path.is_dir())):
+            return str(_path)
+        raise OSError(f"{ftype} '{path}' was not found.")
+
+    paths_tried = []
+    for prefix in search_paths:
+        try_path = (pathlib.Path(prefix) / _path).resolve()
+
+        if ((path_to_file is None  and try_path.exists()) or
+            (path_to_file is True  and try_path.is_file()) or
+            (path_to_file is False and try_path.is_dir())):
+            return str(try_path.absolute())
+
+        paths_tried.append(str(try_path))
+
+    raise OSError(
+        f"{ftype} {_path.name} was not found in the following place(s): {str(paths_tried)}."
+    )
+
+def _resolve_path(
+    path: str,
+    search_paths: typing.List[str|pathlib.Path]|None=None,
+    path_to_file: bool|None=None
+)->str:
+    """
+    Find the path to an existing file, searching the paths specified
+    in the search_paths parameter if the filename is not absolute, and
+    converting a UNIX-style path to the current OS's format if
+    necessary.
+
+    To turn a supplied relative path into an absolute one, the path is
+    appended to paths in the search_paths parameter, in order, until
+    the file is found.
+
+    An IOError is raised if the file is not found.
+
+    Similar to Python's os.path.abspath(), except more search paths
+    than just os.getcwd() can be used, and the file must exist.
+
+    Parameters
+    ----------
+    search_paths : tuple, default=(os.getcwd(),)
+        Tuple of paths to search the path from
+    check_exists: boolean, default=True
+        If True (default) the path must exist on instantiation and set,
+        otherwise the path can optionally exist.
+    """
+    if not search_paths:
+        _search_paths = _get_default_search_paths()
+    else:
+        _search_paths = tuple(search_paths)
+    return _resolve_path_fast(path, _search_paths, path_to_file)
+
+def _get_default_search_paths_as_list()->typing.List[str]:
+    return [str(path) for path in _get_default_search_paths()]
+
 class resolve_path(ParameterizedFunction):
     """
     Find the path to an existing file, searching the paths specified
@@ -2615,7 +2692,7 @@ class resolve_path(ParameterizedFunction):
     than just os.getcwd() can be used, and the file must exist.
     """
 
-    search_paths = List(default=[os.getcwd()], pickle_default_value=False, doc="""
+    search_paths = List(default=list(_get_default_search_paths_as_list()), pickle_default_value=False, doc="""
         Prepended to a non-relative path, in order, until a file is
         found.""")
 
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
## Additional Speed up if using faster attribute lookup.

If the implementation of `Path` would not use costly `self._resolve_path` and `self.search_paths` look ups then it would again be much faster.

The below was run on `this` branch and shows this.

image

param/parameters.py
```diff
@@ -2598,6 +2600,81 @@ def _validate_value(self, val, allow_None):
...
+    if not search_paths:
+        _search_paths = _get_default_search_paths()
+    else:
+        _search_paths = tuple(search_paths)
+    return _resolve_path_fast(path, _search_paths, path_to_file)
+
+def _get_default_search_paths_as_list()->typing.List[str]:
+    return [str(path) for path in _get_default_search_paths()]
+
 class resolve_path(ParameterizedFunction):
```
MarcSkovMadsen, line 2678: Somewhere we should test users that there is a overhead cost of using this class instead of the `_resolve_path` function. I don't know if we should even deprecate it?

maximlt:
There are broken tests so I assume this is still work in progress?

A test suite is not a benchmark so I'm not surprised we don't see performance penalties or regressions from the test suite itself. We now leverage `asv` to run some benchmark, feel free to contribute to it https://github.com/holoviz/param/tree/main/benchmarks.

MarcSkovMadsen:
I hoped to get some feedback first. For example its seems Param does not utilize caching today and the focus is minimum of code. This is a change in favor of speed. 

Thus the first step is really to spark the discussion with this PR. And I would like to know if you would like to accept this before I spend time finalizing.

The tests are breaking on macOS and Windows. To me it looks like the path strings are resolved in different But equivalent ways by os and pathlib.

Having thought about it, I assume parhlib was better as its "modern" Python. But having thought about it, I'm not sure. At least I should profile that too.

MarcSkovMadsen:
Ok. Using pathlib is slow. The image below slows the new implementation uncached on left and old implementation uncached on right.

image

MarcSkovMadsen:
I've added two benchmarks @maximlt. I've read the README and run the commands described there. But I don't know what to expect. It seems asv does not collect the new suite that I've added. I don't know if that is expected.

```bash
$ asv run
· Creating environments
· Discovering benchmarks
· Running 24 total benchmarks (1 commits * 1 environments * 24 benchmarks)
[ 0.00%] · For param commit eaa106b6 <v2.0.1^0>:
[ 0.00%] ·· Benchmarking virtualenv-py3.10
[ 2.08%] ··· Running (benchmarks.ImportSuite.timeraw_import_param--)........................
[52.08%] ··· benchmarks.ImportSuite.timeraw_import_param                                                                                                                                                                                                                                                                                       78.7±0.9ms
[54.17%] ··· benchmarks.ParameterSuite.time_instantiation                                                                                                                                                                                                                                                                                      35.8±0.1μs
[56.25%] ··· benchmarks.ParameterizedDependsInstantiateSuite.time_declarative_10_parameters_separate_cb                                                                                                                                                                                                                                       24.0±0.07μs
[58.33%] ··· benchmarks.ParameterizedDependsInstantiateSuite.time_declarative_10_parameters_shared_cb                                                                                                                                                                                                                                          23.9±0.1μs
[60.42%] ··· benchmarks.ParameterizedDependsInstantiateSuite.time_declarative_1_parameter                                                                                                                                                                                                                                                      20.0±0.1μs
[62.50%] ··· benchmarks.ParameterizedDependsInstantiateSuite.time_watch_10_parameters_separate_cb                                                                                                                                                                                                                                                 956±2μs
[64.58%] ··· benchmarks.ParameterizedDependsInstantiateSuite.time_watch_10_parameters_shared_cb                                                                                                                                                                                                                                                   801±3μs
[66.67%] ··· benchmarks.ParameterizedDependsInstantiateSuite.time_watch_1_parameter                                                                                                                                                                                                                                                             122±0.6μs
[68.75%] ··· benchmarks.ParameterizedDependsSuite.time_declarative_10_parameters_separate_cb                                                                                                                                                                                                                                                     1.48±0ms
[70.83%] ··· benchmarks.ParameterizedDependsSuite.time_declarative_10_parameters_shared_cb                                                                                                                                                                                                                                                    1.37±0.01ms
[72.92%] ··· benchmarks.ParameterizedDependsSuite.time_declarative_1_parameter                                                                                                                                                                                                                                                                    263±1μs
[75.00%] ··· benchmarks.ParameterizedDependsSuite.time_watch_10_parameters_separate_cb                                                                                                                                                                                                                                                           1.48±0ms
[77.08%] ··· benchmarks.ParameterizedDependsSuite.time_watch_10_parameters_shared_cb                                                                                                                                                                                                                                                          1.37±0.01ms
[79.17%] ··· benchmarks.ParameterizedDependsSuite.time_watch_1_parameter                                                                                                                                                                                                                                                                          264±1μs
[81.25%] ··· benchmarks.ParameterizedInstantiateSuite.time_100_parameters                                                                                                                                                                                                                                                                      63.6±0.3μs
[83.33%] ··· benchmarks.ParameterizedInstantiateSuite.time_10_parameters                                                                                                                                                                                                                                                                       23.9±0.1μs
[85.42%] ··· benchmarks.ParameterizedInstantiateSuite.time_1_parameters                                                                                                                                                                                                                                                                        20.1±0.1μs
[87.50%] ··· benchmarks.ParameterizedSetattrSuite.time_class                                                                                                                                                                                                                                                                                  6.37±0.07μs
[89.58%] ··· benchmarks.ParameterizedSetattrSuite.time_instance                                                                                                                                                                                                                                                                               5.77±0.01μs
[91.67%] ··· benchmarks.ParameterizedSuite.time_class_bare                                                                                                                                                                                                                                                                                     76.3±0.1μs
[93.75%] ··· benchmarks.ParameterizedSuite.time_class_with_100_parameter                                                                                                                                                                                                                                                                      11.2±0.04ms
[95.83%] ··· benchmarks.ParameterizedSuite.time_class_with_10_parameter                                                                                                                                                                                                                                                                       1.22±0.01ms
[97.92%] ··· benchmarks.ParameterizedSuite.time_class_with_1_parameter                                                                                                                                                                                                                                                                            208±1μs
[100.00%] ··· benchmarks.WatcherSuite.time_trigger 
```

maximlt:
I need to spend more time to review this (not sure I'll have time this week!). My feedback from working on Param 2.0:

- I'm not surprised there are performance improvements to gain, performance hasn't been a major focus, even if with the addition of the asv benchmark we've been able to catch some regressions and mitigate them.
- don't trust the test suite! It has many gaps, I can't count how many times I broke Panel or HoloViews without breaking the test suite.
- Param is so old that many of its bugs have turned into features, changing it requires great care. It's often worth adding tests before making any change, to capture the previous behavior and see how it's affected.

---

This piece of code breaks with the suggested changes, the last line used to raise an error and it no longer does. It shows that the difficulty with resources like file/directory paths is that they can be removed without Param knowing about it, which is why I think it always resolves. Maybe the way forward is to overload the newly added `check_exists` attribute (https://github.com/holoviz/param/pull/800) with a new `'once'` mode (better name welcome) that only checks whether the resource exists when the Parameter is instantiated.

```python
import os
import param
import pathlib

fp = pathlib.Path('exists.txt')
fp.write_text('')

class P(param.Parameterized):
    path = param.Path('exists.txt')

p = P()
print(p.path)

fp.unlink()

print(p.path)
```

MarcSkovMadsen:
Thanks for the feedback. I plan to give this a second iteration when I have gotten panel-chemistry working for Panel 1.x. First thing would be adding the use case your describe to the tests.

maximlt:
I quickly played around with another implementation (compared to `main`, not to this branch),  (ab)using `check_exists`.

```diff
diff --git a/param/parameters.py b/param/parameters.py
index ed11b53b..8852b79b 100644
--- a/param/parameters.py
+++ b/param/parameters.py
@@ -2706,9 +2706,11 @@ class Path(Parameter):
     ----------
     search_paths : list, default=[os.getcwd()]
         List of paths to search the path from
-    check_exists: boolean, default=True
-        If True (default) the path must exist on instantiation and set,
-        otherwise the path can optionally exist.
+    check_exists: boolean or 'once', default=True
+        If True (default) the path must exist on instantiation and whenever
+        get/set is called, if 'once' the path must only exist on instantiation
+        and on get the first resolved path is returned, if False the path can
+        optionally exist.
     """
 
     __slots__ = ['search_paths', 'check_exists']
@@ -2733,8 +2735,8 @@ class Path(Parameter):
             search_paths = []
 
         self.search_paths = search_paths
-        if check_exists is not Undefined and not isinstance(check_exists, bool):
-            raise ValueError("'check_exists' attribute value must be a boolean")
+        if check_exists is not Undefined and check_exists not in (True, False, 'once'):
+            raise ValueError("'check_exists' attribute value must be a boolean or 'once'")
         self.check_exists = check_exists
         super().__init__(default,**params)
         self._validate(self.default)
@@ -2749,16 +2751,23 @@ class Path(Parameter):
         else:
             if not isinstance(val, (str, pathlib.Path)):
                 raise ValueError(f'{_validate_error_prefix(self)} only take str or pathlib.Path types')
+            path = None
             try:
-                self._resolve(val)
:+                path = self._resolve(val)
             except OSError as e:
-                if self.check_exists:
+                if self.check_exists is True:
                     raise OSError(e.args[0]) from None
+            if path is not None and self.check_exists == 'once':
+                # Abusing check_exists to store the resolved path on parameter
+                # instantiation.
+                self.check_exists = ('once', path)
 
     def __get__(self, obj, objtype):
         """
         Return an absolute, normalized path (see resolve_path).
         """
+        if isinstance(self.check_exists, tuple):
+            return self.check_exists[1]
         raw_path = super().__get__(obj,objtype)
         if raw_path is None:
             path = None
```

With these benchmarks:

```python

class ParameterPathResolveSuite:

    def setup(self):
        import pathlib
        p = pathlib.Path('dummy')
        p.write_text('dummy')

        class P(param.Parameterized):
            x0 = param.Path(str(p))

        self.P = P
        self.p = P()

    def time_resolve_new(self):
        p = self.P()
        p.x0

    def time_resolve_same(self):
        self.p.x0


class ParameterPathResolveOnceSuite:

    def setup(self):
        import pathlib
        p = pathlib.Path('dummy')
        p.write_text('dummy')

        class P(param.Parameterized):
            x0 = param.Path(str(p), check_exists='once')

        self.P = P
        self.p = P()


    def time_resolve_new(self):
        p = self.P()
        p.x0

    def time_resolve_same(self):
        self.p.x0
```

Then I started to think about how this might be related to `readonly` and `constant`, to Parameter attribute inheritance and to `__set__`. To finally decide this could wait for January 🙃

PR #892, opened 2023-12-07
Fix typo in user guide: Outputs.ipynb

alfredocarella:
A small method names mixup.


jbednar (approved):
Thanks! The failing tests are due to Python 3.12, which isn't fully set up anyway. so I'm happy to merge.

PR #901, opened 2024-01-11
Make Comparator check for identity

philippjfr:
Seems like a crazy oversight, not all objects can be compared for equality so we should always check for identity first.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1710,6 +1710,8 @@ class Comparator:
 
     @classmethod
     def is_equal(cls, obj1, obj2):
+        if obj1 is obj2:
+            return True
         for eq_type, eq in cls.equalities.items():
             try:
                 are_instances = isinstance(obj1, eq_type) and isinstance(obj2, eq_type)


jbednar:
It certainly sounds safe to add this check.

maximlt:
I thought we didn't compare for identity because of mutable objects. Take the code below, right now the `debug` callback is called. With your changes, it no longer is.

```python
import param

class Dummy: pass

dummy = Dummy()

    d = param.ClassSelector(class_=Dummy)
    
    @param.depends('d', watch=True)
    def debug(self):
        print('updated d')

foo = Foo(d=dummy)
dummy.changed = True
foo.d = dummy  # triggers `debug` now, not with this PR
```

We could decide that's the right course of action, if so, I don't think it should be released in a *patch* release.

philippjfr:
Thanks, that's a good point. Personally I don't think this was intentional but I appreciate that it's a significant change in behavior. For mutable objects the correct way to trigger an event should always be to explicitly `trigger` an event. That is because if we ever did register an equality comparison for a particular type of object then the event would no longer get triggered because the equality check will always pass if the object has the same identity.

philippjfr:
As an example to illustrate this point, the two mutables that we do have equality comparisons defined for do not trigger an event:

```python
import param

    d = param.List()
    
    @param.depends('d', watch=True)
    def debug(self):
        print('updated d')

mutable = []

foo = Foo(d=mutable)
mutable.append(1)
foo.d = mutable # does not trigger `debug`
```

philippjfr:
The docs about trigger also make this point:

>Usually, a Watcher will be invoked only when a parameter is set (and only if it is changed, by default). What if you want to trigger a Watcher in other cases? For instance, if a parameter value is a mutable container like a list and you add or change an item in that container, Param’s set method will never be invoked, because in Python the container itself is not changed when the contents are changed. In such cases, you can trigger a watcher explicitly, using `.param.trigger(*param_names)`.

maximlt:
Yes, your points make sense to me.

We often see Param users, and in particular Panel users using Param, asking why updating their list/dict doesn't execute their callbacks. Expanding this behavior to "complex" (there's got to be a better name for this) will need to be carefully documented (with a big warning!).

bertcoerver:
Is there any progress on this? I'm learning param and I find it very strange and confusing that appending to a list or assigning a new key to a dictionary doesn't trigger a watcher. 

The `.param.trigger(*param_names)` workaround is also not very convenient. It makes the `param.parameterized.Event` that gets passed to the watched function (when using `self.param.watch(fn, [*param_names])`) much less useful, because `event.type == "triggered"` and `event.new == event.old`.

PR #908, opened 2024-02-08
Improvements for synchronous and asychronous generators

philippjfr:
This PR improves the support for synchronous and asychronous generators as references and dependencies. It also fixes a variety of bugs related to updating async generators. Specifically it implements the following:

## Enhancements

- Generators can now be used as references meaning that they are scheduled on a thread and then are applied asynchronously

```python

    value = param.Number(allow_refs=True)

def gen():
    i = 0
    while True:
        yield i
        time.sleep(0.1)
        i += 1

t = Test(value=gen)

t.param.value.rx()
```

gen_ref

- Add the ability to construct expressions directly from sync and async generators

gen_expr

- Add ability to Skip reference updates of all kinds

- Asynchronous generators can now be annotated with `param.depends`
- In IPython contexts we now define a default `async_executor` meaning that async functionality can now be used by default without importing Panel or defining your own executor.

## Bug fixes

- Fixes an issue with asynchronous generators not being correctly cleaned up when overridden
- Ensure replacing an async generator function does not try to update the parameter value with the function
- Ensure replacing an async generator function does not delete the reference to a newer task making it immortal
- Do not allow `param.Number` to accept generators since they are not actually supported for `Dynamic` parameters

## ToDo

- [x] Implements https://github.com/holoviz/panel/issues/6295
- [x] Add tests
- [x] Update docs


param/ipython.py
```diff
@@ -364,23 +365,47 @@ def __init__(self, reactive):
...
+            # We are in Jupyter and can piggyback the tornado IOLoop
+            from tornado.ioloop import IOLoop
+            ioloop = IOLoop.current()
+            event_loop = ioloop.asyncio_loop # type: ignore
+    except Exception:
+        pass
+    if event_loop is None:
+        event_loop = asyncio.get_event_loop()
+    if event_loop.is_running():
+        ioloop.add_callback(func)
```
Hoxbro, line 406: Can't this be unbound?

Have you thought about `asyncio.run` with `loop_factory`? This was added in 3.12, so a compatibility layer is needed for earlier versions; see https://github.com/dask/distributed/pull/7969 for inspiration.

tests/testrefs.py
```diff
@@ -198,3 +199,31 @@ async def gen_strings2():
     assert task1.done()
     assert not task2.done()
     assert p._param__private.async_refs['string'] is task2
+
+async def test_generator_ref_cancelled():
```
Hoxbro, line 203:
```suggestion
@pytest.mark.asyncio
async def test_generator_ref_cancelled():
```

philippjfr: We have a ton of asyncio tests that aren't marked as such.

philippjfr: We already configure `asyncio_mode = "auto"` in the pyproject.toml, so really don't think this is needed.

param/_utils.py
```diff
@@ -552,3 +588,22 @@ def _in_ipython():
...
+
+def async_executor(func):
+    try:
+        event_loop = asyncio.get_running_loop()
+    except RuntimeError:
+        event_loop = asyncio.new_event_loop()
+    if event_loop.is_running():
+        task = asyncio.ensure_future(func())
+        _RUNNING_TASKS.append(task)
+        task.add_done_callback(_cleanup_task)
```
Hoxbro, line 606:
```suggestion
        task.add_done_callback(_RUNNING_TASKS.discard)
```

And remove `_cleanup_task`

philippjfr: Confused by this, `_RUNNING_TASKS` is a list. Lists don't have discard.

doc/user_guide/Generators.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "\n",
+    "tidal_gauge()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1e8361a2-8af5-4fd0-ba42-646c8e988f87",
+   "metadata": {},
+   "source": [
+    "Instead of scheduling some periodic event on some event loop we can simply write a generator function to poll his sensor at some defined interval:"
```
ahuang11, line 39:
```suggestion
    "Instead of scheduling some periodic event on some event loop we can simply write a generator function to poll this sensor at some defined interval:"
```

param/_utils.py
```diff
@@ -219,6 +227,34 @@ def iscoroutinefunction(function):
...
+    def safe_next():
+        # Converts StopIteration to a sentinal value to avoid:
+        # TypeError: StopIteration interacts badly with generators and cannot be raised into a Future
+        try:
+            return next(sync_gen)
+        except StopIteration:
+            return done
+
+    while True:
+        if sys.version_info >= (3, 9):
```
Hoxbro, line 251: If this is gonna be in Param 2.1, maybe we should drop Python 3.8 support?

param/_utils.py
```diff
@@ -552,3 +588,18 @@ def _in_ipython():
         return True
     except NameError:
         return False
+
+_RUNNING_TASKS = set()
```
Hoxbro, line 592: As `_RUNNING_TASKS` is not a constant, shouldn't it be lowercase?

param/_utils.py
```diff
@@ -552,3 +588,18 @@ def _in_ipython():
...
+        event_loop = asyncio.get_running_loop()
+    except RuntimeError:
+        event_loop = asyncio.new_event_loop()
+    if event_loop.is_running():
+        task = asyncio.ensure_future(func())
+        _RUNNING_TASKS.add(task)
+        task.add_done_callback(_RUNNING_TASKS.discard)
+    else:
+        raise Exception(func)
+        event_loop.run_until_complete(func())
```
Hoxbro, line 605: This looks weird.

param/ipython.py
```diff
@@ -364,23 +364,46 @@ def __init__(self, reactive):
 
     def __call__(self):
         from param.depends import depends
-        from param.parameterized import resolve_ref
+        from param.parameterized import Undefined, resolve_ref
         from param.reactive import rx
 
+        handle = None
```
Hoxbro, line 370:
```suggestion
        handle: IPython.core.display_functions.DisplayHandle | None  = None
```
Was a bit confused when reading the code.  The import could be done by a `TYPE_HINTING`.

param/ipython.py
```diff
@@ -364,23 +364,46 @@ def __init__(self, reactive):
...
+            # We are in Jupyter and can piggyback the tornado IOLoop
+            from tornado.ioloop import IOLoop
+            ioloop = IOLoop.current()
+            event_loop = ioloop.asyncio_loop # type: ignore
+            if event_loop.is_running():
+                ioloop.add_callback(func)
+            else:
+                event_loop.run_until_complete(func())
+            return
+    except Exception:
```
Hoxbro, line 405:
```suggestion
    except (NameError, AttributeError):
```

If some API changes are made in the if statement, it can become hard to debug. `NameError` is to catch `get_ipython` and `AttributeError`  for `ip.kernel`.

param/_utils.py
```diff
@@ -46,6 +50,10 @@ class ParamFutureWarning(ParamWarning, FutureWarning):
     Always displayed.
     """
 
+class Skip(RuntimeError):
```
Hoxbro, line 53:
```suggestion
class SkipReferenceUpdate(RuntimeError):
```

Maybe be more explicit about what `Skip` does; this is only a suggestion.

param/_utils.py
```diff
@@ -184,7 +192,7 @@ def _validate_error_prefix(parameter, attribute=None):
 
 def _is_mutable_container(value):
     """True for mutable containers, which typically need special handling when being copied"""
-    return issubclass(type(value), (abc.MutableSequence, abc.MutableSet, abc.MutableMapping))
+    return issubclass(type(value), MUTABLE_TYPES)
```
Hoxbro, line 195: I am curious as to why `isinstance(value, MUTABLE_TYPES)` was not used here.

param/parameterized.py
```diff
@@ -1985,6 +1990,8 @@ def _setup_refs(self_, refs):
             ))
 
     def _update_ref(self_, name, ref):
+        if name in self_.self._param__private.async_refs:
+            self_.self._param__private.async_refs.pop(name).cancel()
         for _, watcher in self_.self._param__private.ref_watchers:
```
Hoxbro, line 1995:
```suggestion
        param_private = self_.self._param__private
        if name in param_private.async_refs:
            param_private.async_refs.pop(name).cancel()
        for _, watcher in param_private.ref_watchers:
```

It is mainly for readability, but it is more of a personal preference, so you can ignore it.

param/parameterized.py
```diff
@@ -2030,21 +2047,27 @@ async def _async_ref(self_, pname, awaitable):
...
                         self_.update({pname: new_obj})
             else:
                 with _syncing(self_.self, (pname,)):
-                    self_.update({pname: await awaitable})
+                    try:
+                        self_.update({pname: await awaitable})
+                    except Skip:
+                        pass
         except Exception as e:
             raise e
```
Hoxbro, line 2066:
```suggestion
```

param/_utils.py
philippjfr: Fair point, I do this relatively frequently but you're right I should change that and use lower case for mutable globals.

param/_utils.py
philippjfr: This will be user level as, so I like the short name here.

param/_utils.py
philippjfr: Excellent question I don't have a good answer for.

param/_utils.py
philippjfr: I would like that.

param/_utils.py
Hoxbro: I think it is too ambiguous to be useful. 

If it is a public API, I think it should be mentioned somewhere in the documentation.

param/_utils.py
```diff
@@ -552,3 +588,18 @@ def _in_ipython():
...
+    try:
+        event_loop = asyncio.get_running_loop()
+    except RuntimeError:
+        event_loop = asyncio.new_event_loop()
+    if event_loop.is_running():
+        task = asyncio.ensure_future(func())
+        _RUNNING_TASKS.add(task)
+        task.add_done_callback(_RUNNING_TASKS.discard)
+    else:
+        raise Exception(func)
```
philippjfr, line 604:
```suggestion
```

doc/user_guide/Generators.ipynb
jbednar: Also too many "some"s in the same sentence!

param/_utils.py
philippjfr: Will leave this in and then make a final decision on removing this or not before release.

philippjfr:
@jbednar I'll leave final review and merging up to you if you get a chance, otherwise I'll ask for approval from @Hoxbro and @maximlt before merging.

Hoxbro:
I'm still not a fan of the name `Skip`. 

Another thing is, do we need it to be an Exception? It almost seems like the user is doing something wrong each time they raise it. If it is an exception because it is easier to propagate the error to the right place for us as maintainers and does not have a real effect on the user if we change it to an object, I think we should change it.

philippjfr:
I still strongly prefer `Skip` and also think using it as an exception is the right choice. The only other option we have would be to allow the user to return the `Skip` sentinel value and there's a few reasons I don't like this:

1. The most important reason is that dependent functions can be nested in a number of ways and there is no obvious way to intercept a a skipped sentinel value that is returned inside another dependent function. This means user code would have to know about the skip sentinel value and explicitly check for it to avoid performing invalid operations on it and causing actual exceptions.
2. It does not clearly distinguish the skip case from a regular `return`/`yield` case.
3. Various forms of caching/memoization can be added to these functions and `Skip` should always bypass the cache.

Hoxbro:
Let's start from the back (I would say I'm playing devil's advocate for some of the arguments)

(3) A counterpoint would be if you do a heavy calculation in a function and then raise a `Skip` at the end. Having a sentinel value will actually remember that this function did return a `Skip`.

(2) Why is this important? 

I think this code with sentinal is cleaner,
``` python
# with sentinal value
value = func()
if value is Skip:
    value = None

# with exception
try:
    value = func()
except Skip: 
    value = None
```
Another thing is that `Skip` could be wrongly caught in `except RuntimeError` statement, albeit highly unlikely.

(1) I have a hard time coming up with an example that does not end with a `param.bind` (or similar) getting the returned `Skip`. Or at least a hard time if we assume the user can't support such cases in their nested functions. 

The problem you are describing is supporting suboptimal code. This could lead to a side effect function raising `Skip`, which is then raised to the top level even though this is not what you wanted, and then you are in a worse situation than with a sentinal value as you have no idea if the "correct" `Skip` was raised. All of this could be significantly harder if it is a watched function that raises `Skip` at a distance. 

Another reason for having it as a sentinal value is also that it will make it possible to type a hint that a function can return it; this is not possible with exceptions in Python. 

--- 

With that being said I generally don't have a problem with it being an exception. I just want to be sure that it is the best way to handle a skip, and if you feel that it is I have no objections to it. 

---
For naming maybe `SkipUpdate`. (Note: I wouldn't be that afraid of the length of the name as most modern editors have autocomplete.)

philippjfr:
>(3) A counterpoint would be if you do a heavy calculation in a function and then raise a Skip at the end. Having a sentinel value will actually remember that this function did return a Skip.

Seems like an anti-pattern to run some long calculation only to then skip, but I'll accept there will be (I would consider rare) cases where you don't know you want to skip until you get to the end of some long-running computation.

> (2) Why is this important?

It isn't super important but it's not our code I'm talking about but user code, i.e. I like the fact that `raise Skip()` stands out, while `return Skip` looks like any other return statement.

>(1) I have a hard time coming up with an example that does not end with a param.bind (or similar) getting the returned Skip. Or at least a hard time if we assume the user can't support such cases in their nested functions.

I'm not really following your point here in particular the stuff about `watch`+`Skip` since that combination really doesn't make sense and isn't what Skip is intended for. In the side-effectey world you don't have to `Skip`, you can just return early. What I am thinking about is scenarios where you might have a method that is useful both on its own and may be invoked by another method, e.g. take my crappy example

```python
class DatePlotter(param.Parameterized):

    date = param.CalendarDate()

    update = param.Event()

    ...

     @param.depends('date')
     def get_daily_data(self):
          if not self.update:
              raise Skip()
          return self.df[self.df.date==self.date]

     @param.depends('date')
     def plot_daily_data(self):
         daily_df = self.get_daily_data()
         return daily_df.hvplot()

dp = DailyPlotter()
pn.Row(dp.param, pn.widgets.Tabulator(dp), pn.pane.HoloViews(dp.plot_daily_data))
```

Here I can have a `get_daily_data` function that is both used directly and invoked by some other method for another purpose and I can write the `Skip` logic once. Both the consumer of `get_daily_data` and the consumer of `plot_daily_data` can rely on the fact that the data will only emit if the `update` parameter is triggered (i.e. when the button is clicked) without the user having to handle both cases. If we have a sentinel value returned by `get_daily_data` then `plot_daily_data` now also has to implement the Skip logic OR handle the Skip return value.

Hoxbro:
> Seems like an anti-pattern to run some long calculation only to then skip, but I'll accept there will be (I would consider rare) cases where you don't know you want to skip until you get to the end of some long-running computation.

Then, let's say it is a lookup in a database, which is slow. 

> It isn't super important but it's not our code I'm talking about but user code, i.e. I like the fact that raise Skip() stands out, while return Skip looks like any other return statement.

I really don't think that it needs to stand out. As I see it, a raised exception is usual when something goes wrong, and that is not what `Skip` is; it is part of the API. This could lead to confusion for non-advanced users, maybe they will even try to avoid it because `Exception == "bad"`.

> I can write the Skip logic once

I don't necessarily see this as a good thing. I think this could lead to weird side effects, like why a widget isn't updating because some weird nested function is raising `Skip`. I think the main concern for me is also the advantage a raise brings: It breaks the control flow, you don't know what will catch it, and that could lead to hard-to-debug problems. 
Having it as a sentinal value will, yes, introduce more code, but it will make users design their apps with this in mind upfront instead of having to figure it out in a debug session.  

In your example, there is no way to typehint that the `get_daily_data` can raise a Skip, which is okay when you have it just above the method, but what if it is inherited from a class in another file? This means that potentially every method/function call can raise a `Skip`, and you would have no idea which one is causing it. 

If we take your example and want to send a notification to the user that there is no data in `plot_daily_data`? This could quickly become ugly when having `Skip` as a raised exception. What if there is another function call inside  `plot_daily_data`, which also raises a `Skip` but for a different reason, and you want to send a different notification to the user? 

``` python
class DatePlotter(param.Parameterized):

    date = param.CalendarDate()

    update = param.Event()

    ...

     @param.depends('date')
     def get_daily_data(self) -> pd.DataFrame | Skip:
          if not self.update:
              return Skip
          return self.df[self.df.date==self.date]

     @param.depends('date')
     def plot_daily_data(self) -> hvplot | Skip:
         daily_df = self.get_daily_data()
         # We want to inform the user there is no daily data
         if daily_df is Skip:
              pn.state.notifications.info("No daily data")
              return Skip
         return daily_df.hvplot()

dp = DailyPlotter()
pn.Row(dp.param, pn.widgets.Tabulator(dp), pn.pane.HoloViews(dp.plot_daily_data))
```

An option could be to support both methods, e.g., both returning and raising `Skip`. 

---

Another name for it could be `NoUpdate`.

jbednar:
Exceptions usually are for error conditions, but the `Skip` exception seems analogous to the [StopIteration](https://docs.python.org/3/library/exceptions.html#StopIteration) exception, which is just for unusual control flows, not errors. The code examples above don't make a particularly strong case either way, so I'm inclined to support the way Philipp was originally motivated to write it (as an exception).

jbednar (approved):
Nice!

Hoxbro:
> Exceptions usually are for error conditions, but the Skip exception seems analogous to the [StopIteration](https://docs.python.org/3/library/exceptions.html#StopIteration) exception, which is just for unusual control flows, not errors

The difference is that this is built into the language itself and is called each time you run a for loop without raising an exception yourself. I never think I have raised a `StopIteration` error by myself. 

Another point is that people can catch `Skip` it with an `except Exception` by having it as an exception. And then they have to write code like this
``` python
try:
     # function which can both raise a ValueError and Skip
except Skip:
     raise
except Exception:
     # do something else 
```

Which users likely won't do, or even if they do, they can miss one.

Hoxbro:
The last thing I want to say is that using `Skip` as an exception is not a golden bullet that will always just work. It should be seen as a half-poisoned apple; it tastes good until you get to the poison side of it, and when that happens, it is too late to do anything.

param/_utils.py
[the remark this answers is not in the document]
philippjfr: Indeed.

doc/user_guide/References.ipynb
```diff
@@ -0,0 +1,239 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "w.param.trigger('run')\n",
+    "\n",
+    "v.b"
+   ]
+  }
+ ],
+ "metadata": {
```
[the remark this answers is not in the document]
philippjfr: Done.

param/_utils.py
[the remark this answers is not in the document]
philippjfr: Changed.

philippjfr:
>Then, let's say it is a lookup in a database, which is slow.

If the database lookup fails that may well be because of a timeout error or other temporary issues in which case caching would be bad.

As for your example above, I actually see the exception as a benefit here, because I can attach a reason in a straightforward and canonical way:

```python
class DatePlotter(param.Parameterized):

    date = param.CalendarDate()

    update = param.Event()

    ...

     @param.depends('date')
     def get_daily_data(self) -> pd.DataFrame | Skip:
          if not self.update:
              raise Skip("No updating")
          elif not self.date in self.df.date:
              raise Skip('No daily data')
          return self.df[self.df.date==self.date]

     @param.depends('date')
     def plot_daily_data(self) -> hvplot | Skip:
         try:
             daily_df = self.get_daily_data()
         except Skip as e:
              pn.state.notifications.info(str(e))
              raise e
         return daily_df.hvplot()

dp = DailyPlotter()
pn.Row(dp.param, pn.widgets.Tabulator(dp), pn.pane.HoloViews(dp.plot_daily_data))
```

philippjfr:
@Hoxbro and I had another discussion. He still consider the `Skip` pattern dangerous since someone might raise it deep inside some code and you might never know. My opinion is that `Skip` should never be used deep in some code and have add a cautionary note to the docs saying that `Skip` should be primarily used when wrapping other code to be used as a reactive reference and it should never appear in business logic. I'll trust our users enough not to abuse it but I guess time will tell if that trust was misplaced.

For now I'd like to continue here and with one approval and having addressed the review comments I'm going to merge.

PR #918, opened 2024-03-11
Make it possible to resolve rx expressions recursively

philippjfr:
I'm a little afraid at how "meta" this operation is but it allows doing some things which aren't otherwise possible with `.rx` (or rather extremely complicated). Specifically this allows writing expressions that themselves contain references and then recursively resolves them. The case where I encountered the need for this is a Panel ToDo app.

Let us say we have a layout of tasks:

```python
tasks = pn.Column(task1, task2)
```

Each task is structured like this:

```python
task = Row(Checkbox(), Markdown())
```

Now I can write a simple expression to gather up all checkboxes:

```python
checkboxes = tasks.param.objects.rx().rx.map(lambda task: task[0])
```

And if want to get the number of completed tasks I could do:

```python
sum(checkboxes.rx.map(lambda w: w.value))
```

Unfortunately this only updates when the list of checkboxes changes, i.e. NOT when each individual checkbox is ticked or unticked.

What I really want to do is for the expression to resolve the references (i.e. in this case the widgets) and update when any of the references change. Using `.rx.resolve` I can count the number of completed tasks with:

```python
sum(tasks.param.objects.rx().rx.map(lambda task: task[0]).rx.resolve())
```

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -164,11 +164,13 @@ def eval_function_with_deps(function):
             kwargs = {n: getattr(dep.owner, dep.name) for n, dep in kw_deps.items()}
     return function(*args, **kwargs)
 
-def resolve_value(value):
+def resolve_value(value, recursive=True):
     """
     Resolves the current value of a dynamic reference.
     """
-    if isinstance(value, (list, tuple)):
+    if not recursive:
+        pass
+    elif isinstance(value, (list, tuple)):
         return type(value)(resolve_value(v) for v in value)
     elif isinstance(value, dict):
         return type(value)((resolve_value(k), resolve_value(v)) for k, v in value.items())
@@ -2007,14 +2009,15 @@ class Parameters:
         updates = {}
         for pname, ref in self_.self._param__private.refs.items():
             # Skip updating value if dependency has not changed
-            deps = resolve_ref(ref, self_[pname].nested_refs)
+            recursive = self_[pname].nested_refs
+            deps = resolve_ref(ref, recursive)
             is_gen = inspect.isgeneratorfunction(ref)
             is_async = iscoroutinefunction(ref) or is_gen
             if not any((dep.owner is e.obj and dep.name == e.name) for dep in deps for e in events) and not is_async:
                 continue
 
             try:
-                new_val = resolve_value(ref)
+                new_val = resolve_value(ref, recursive)
             except Skip:
                 new_val = Undefined
             if new_val is Skip or new_val is Undefined:
@@ -2037,7 +2040,7 @@ class Parameters:
             return None, None, value, False
         ref = value
         try:
-            value = resolve_value(value)
+            value = resolve_value(value, recursive=pobj.nested_refs)
         except Skip:
             value = Undefined
         if is_async:
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -90,19 +90,19 @@ from functools import partial
 from types import FunctionType, MethodType
 from typing import Any, Callable, Optional
 
-from . import Event
 from .depends import depends
 from .display import _display_accessors, _reactive_display_objs
 from .parameterized import (
     Parameter, Parameterized, Skip, Undefined, eval_function_with_deps, get_method_owner,
     register_reference_transform, resolve_ref, resolve_value, transform_reference
 )
+from .parameters import Boolean, Event
 from ._utils import iscoroutinefunction, full_groupby
 
 
 class Wrapper(Parameterized):
     """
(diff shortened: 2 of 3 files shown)


param/parameterized.py
```diff
@@ -164,11 +164,13 @@ def eval_function_with_deps(function):
             kwargs = {n: getattr(dep.owner, dep.name) for n, dep in kw_deps.items()}
     return function(*args, **kwargs)
 
-def resolve_value(value):
+def resolve_value(value, recursive=True):
```
philippjfr, line 167: This was an oversight, it should not have recursively resolved values unless requested. Unfortunately changing the default behavior would cause potential regressions now since Panel and HoloViews both depend on it. So the migration plan will have to be for those libraries to explicitly set `recursive=True/False` and then eventually we can consider changing the default.

param/reactive.py
```diff
@@ -90,40 +90,92 @@
...
+    def __init__(self, **params):
+        self._watchers = []
+        super().__init__(**params)
+
+    def _resolve_value(self, *events):
+        nested = self.param.object.nested_refs
+        refs = resolve_ref(self.object, nested)
+        value = resolve_value(self.object, nested)
+        if self.recursive:
+            new_refs = [r for r in resolve_ref(value, nested) if r not in refs]
```
Hoxbro, line 150:
```suggestion
            new_refs = [r for r in resolve_ref(value, recursive=nested) if r not in refs]
```

Is clearer for me to understand by using keywords arguments here.

param/reactive.py
```diff
@@ -90,40 +90,92 @@
...
+            while new_refs:
+                refs += new_refs
+                value = resolve_value(value, nested)
```
Hoxbro, line 153:
```suggestion
                value = resolve_value(value, recursive=nested)
```

param/reactive.py
```diff
@@ -90,40 +90,92 @@
...
+                new_refs = [r for r in resolve_ref(value, nested) if r not in refs]
```
Hoxbro, line 154:
```suggestion
                new_refs = [r for r in resolve_ref(value, recursive=nested) if r not in refs]
```

param/reactive.py
```diff
@@ -90,40 +90,92 @@
...
+        self._watchers = []
+        super().__init__(**params)
+
+    def _resolve_value(self, *events):
+        nested = self.param.object.nested_refs
+        refs = resolve_ref(self.object, nested)
+        value = resolve_value(self.object, nested)
+        if self.recursive:
+            new_refs = [r for r in resolve_ref(value, nested) if r not in refs]
+            while new_refs:
```
Hoxbro, line 151:
```suggestion
            while new_refs := [r for r in resolve_ref(value, recursive=nested) if r not in refs]:
```
And delete the last line in the while-loop. (I know you like these walrus)

param/reactive.py
```diff
@@ -235,6 +287,27 @@ def pipe(self, func, /, *args, **kwargs):
         """
         return self._as_rx()._apply_operator(func, *args, **kwargs)
 
+    def resolve(self, nested=True, recursive=False):
```
Hoxbro, line 290:
```suggestion
    def resolve(self, *, nested=True, recursive=False):
```

So that we don't get a function call like this resolve(True, True)

tests/testreactive.py
```diff
@@ -433,6 +435,96 @@ def test_reactive_when_initial():
     p.param.trigger('event')
     assert integer.rx.value == 4
 
+def test_reactive_resolve():
+    p = Parameters(integer=3)
+    p2 = Parameters(parameter=p.param.integer)
+
+    prx = p2.param.parameter.rx()
+    assert prx.rx.value is p.param.integer
```
Hoxbro, line 443: Maybe it is better to use something other than numbers here. This will always be true if both results are 3.

philippjfr:
Going to merge so I can cut an RC. If you have review comments please still leave them and I'll follow up before release.

Hoxbro (commented):
I have left some minor comments.

tests/testreactive.py
philippjfr: `p.param.integer` is a Parameter object, not a number.

philippjfr:
Thanks, will make a new PR incorporating all your suggestions.

param/reactive.py
philippjfr: Indeed, although it's also a little confusing, `recursive` means something different in the context of `resolve_ref` than it does in the context of the `Resolver`.

Coderambling:
Nice. Is the intention for this to be part of 1.4?

Issue #867, opened 2023-10-05
Reactive: root executed while updating a branch


maximlt:
I was trying to understand what happens when a reactive expressions has multiple "reactive inputs".

```mermaid
graph TD;
    firstnamex-->append_lastname;
    lastnamex-->append_lastname;
```

Looking at the last cell of the image below, I'm surprised to see `'transforming first name johnny'` being printed while updating `lastnamex`, is that expected?

![image](https://github.com/holoviz/param/assets/35924738/2a0c540c-156a-4ed8-8c1c-5dcb335676e1)

```python
from param import rx

firstnamex = rx('john')
lastnamex = rx('lenon')

def debug(value, info=''):
    print(info, value)
    return value

def append_lastname(fn, ln):
    return fn + ' ' + ln

pfnx = firstnamex.rx.pipe(debug, info='transforming first name').title()
plnx = lastnamex.rx.pipe(debug, info='transforming last name').title()
together = pfnx.rx.pipe(append_lastname, ln=plnx).rx.pipe(debug, info='merged')

together

firstnamex.rx.value = 'johnny'
lastnamex.rx.value = 'hallyday'
```

philippjfr:
Seems to be to do with over-eagerly invalidating the root.

PR #919, opened 2024-03-13
Avoid overeager root invalidation on rx

philippjfr:
- [x] Fixes https://github.com/holoviz/param/issues/867
- [x] Add test

### DIFF
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -763,7 +763,9 @@ class rx:
         """
         if self._fn is not None:
             for _, params in full_groupby(self._fn_params, lambda x: id(x.owner)):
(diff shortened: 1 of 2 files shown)


param/reactive.py
```diff
@@ -790,6 +792,7 @@ def _resolve(self):
                     if obj is Skip:
                         raise Skip
             except Skip:
+                self._dirty = False
```
philippjfr, line 795: Technically unrelated and just an oversight.

maximlt (approved):
I just pushed a cosmetic change to make the test a little easier to understand. Other than that, LGTM.

PR #922, opened 2024-03-19
Add a page listing the active deprecations

maximlt:
Experimenting with adding this page, as I've been working on [HEP 2](https://github.com/holoviz/holoviz/pull/388) and think that it'd help maintainers (users too in fact!) if such a page existed and was maintained over time.

image

For example, I bumped a couple of warnings in https://github.com/holoviz/param/pull/921 and updated the page accordingly.

image

### DIFF
--- a/doc/reference/deprecations.md
+++ b/doc/reference/deprecations.md
@@ -1,0 +1,36 @@
+# Deprecations
+
+- `ParamFutureWarning` since `2.0.0` - Parameterized namespace / `instance._param_watchers` (getter and setter): use instead the property `inst.param.watchers`
+- `ParamFutureWarning` since `2.0.0` - Warn on failed validation of the *default* value of a Parameter after the inheritance mechanism has completed
+- `ParamFutureWarning` since `2.0.0` - Running unsafe operations **during Parameterized instance initialization**, instead run these operations after having called `super().__init__(**params)`:
+    - `instance.param.objects(instance=True)`
+    - `instance.param.trigger("<param_name>")`
+    - `instance.param.watch(callback, "<param_name>")`
+- `ParamDeprecationWarning` since `2.0.0` - Parameter slots / `List._class`: use instead `item_type`
+- `ParamDeprecationWarning` since `2.0.0` - Parameter slots / `Number.set_hook`: no replacement
+- `ParamDeprecationWarning` since `2.0.0` - `param.__init__` module / `param.produce_value`: no replacement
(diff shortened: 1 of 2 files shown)


Hoxbro:
Did you try putting this information in a table? 

I don't think we should mention old deprecations. This should ideally be handled by versioning the documentation. As we don't have that, I think it is ok for now. 

> users too in fact!

TBH, I don't think users will look at a webpage to see deprecations, but I understand your point.

maximlt:
> Did you try putting this information in a table?

Done in https://github.com/holoviz/param/pull/922/commits/07597faf9c3eb624f9062561f3c3d23ec168f4ce

image

> I don't think we should mention old deprecations. This should ideally be handled by versioning the documentation. As we don't have that, I think it is ok for now.

I should have explained why I included the deprecation history. It's almost the most important piece of information I need as a maintainer as I want an easy way to know for how long a feature has been deprecated. I had to do so much Github digging when releasing Param 2.0 through the release notes and Github issues/PRs.
In my mind, when a feature is eventually removed from the code base, it is to be removed from this page.

> TBH, I don't think users will look at a webpage to see deprecations, but I understand your point.

Just sharing these links I just found (the page I added in this PR is nowhere near these!):

- https://docs.sympy.org/latest/explanation/active-deprecations.html
- https://docs.pytest.org/en/7.1.x/deprecations.html

Hoxbro:
I think there is two conflicting perspective with your last comment, one is what is good for the developer, the second is what is good for the user. I don't see them as the same thing. 

> I should have explained why I included the deprecation history. It's almost the most important piece of information I need as a maintainer as I want an easy way to know for how long a feature has been deprecated. I had to do so much Github digging when releasing Param 2.0 through the release notes and Github issues/PRs.

This is looking from it as a developer. From a users perspective he doesn't care for previous versions deprecations, he care about _his_ version. 

> Just sharing these links I just found (the page I added in this PR is nowhere near these!):

May I ask you when did you last look at this? Because I haven't looked at them before, and I would say I use pytest a lot. 🙃

maximlt:
My main motivation for this PR was to experiment with how to help maintainers better handle deprecations over time, to serve the development of HEP 2. I picked Param because I know more about it's deprecation warnings than any other library, and because it has many active ones so it's a good use case.

I thought that having that as a listing in the docs could also be useful for users. The review shows that this view isn't shared. I think this discussion should take place in HEP 2 with a larger group, so I'm going to close this PR, I can still point to it from HEP 2 as a possible way to improve things.

philippjfr:
Personally I call out deprecations explicitly in the release notes but do potentially see value in an overview and/or migration guide.

PR #924, opened 2024-03-21
Add support for async and generator functions in rx.pipe

philippjfr:
We already support using asynchronous and generator functions as inputs to a reactive expression but until now we did not handle them when you use them as an input to an operation that is performed on the data. This meant that `.rx.pipe` and `.rx.map` would simply return the coroutine or generator as the value instead of evaluating it lazily.

Here we add support for these by scheduling the task on the event loop and then triggering an event when the function evaluates. Internally this is achieved by creating a `Trigger` which is added as one of the dependencies of the reactive expression such as that any consumer of the expression (whether Panel, the display hook or anything else) is notified when the async function or generator yields a new value.

In testing this I also found an oversight which meant that if you used a bound function as the input to `.rx.pipe` it would not extract its dependencies.

### DIFF
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -258,8 +258,12 @@ class reactive_ops:
         kwargs: mapping, optional
           A dictionary of keywords to pass to `func`.
         """
-        def apply(vs, *args, **kwargs):
-            return [func(v, *args, **kwargs) for v in vs]
+        if inspect.iscoroutinefunction(func):
+            async def apply(vs, *args, **kwargs):
+                return list(await asyncio.gather(*(func(v, *args, **kwargs) for v in vs)))
+        else:
+            def apply(vs, *args, **kwargs):
+                return [func(v, *args, **kwargs) for v in vs]
         return self._as_rx()._apply_operator(apply, *args, **kwargs)
 
     def not_(self):
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -81,6 +81,7 @@ display its repr.
 """
 from __future__ import annotations
 
+import asyncio
 import inspect
 import math
 import operator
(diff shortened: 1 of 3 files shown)


param/reactive.py
```diff
@@ -256,8 +258,12 @@ def map(self, func, /, *args, **kwargs):
         kwargs: mapping, optional
           A dictionary of keywords to pass to `func`.
         """
-        def apply(vs, *args, **kwargs):
-            return [func(v, *args, **kwargs) for v in vs]
+        if inspect.iscoroutinefunction(func):
```
Hoxbro, line 261: Should we raise an exception if a generator is used for maps? I don't think it makes much sense.

tests/testreactive.py
```diff
@@ -551,6 +576,24 @@ def gen():
...
+    rxgen = rxv.rx.pipe(gen)
+    rxgen.rx.watch()
+    assert rxgen.rx.value is param.Undefined
+    await asyncio.sleep(0.04)
+    assert rxgen.rx.value == 1
+    await asyncio.sleep(0.1)
+    assert rxgen.rx.value == 2
+    rxv.rx.value = 2
+    await asyncio.sleep(0.03)
+    assert rxgen.rx.value == 3
```
Hoxbro, line 595: I'm not entirely sure why the first one is 0.04, and the last one is 0.03. Is that just because it works?

Hoxbro (commented):
Some minor comments, but LGTM.

tests/testreactive.py
philippjfr: Solid question, really we need a `wait_until` utility like we have in Panel.

param/reactive.py
philippjfr: Agreed, I almost went down the road of implementing it but it was just a little too crazy to be sensible.

tests/testreactive.py
Hoxbro: Steal it

PR #929, opened 2024-04-02
Ensure refs can be updated by watcher of the same parameter

philippjfr:
@Hoxbro Please test and if possible add your test case.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1471,12 +1471,13 @@ class Parameter(_ParameterBase):
         item in a list).
         """
         name = self.name
-        if obj is not None and self.allow_refs and obj._param__private.initialized and name not in obj._param__private.syncing:
+        if obj is not None and self.allow_refs and obj._param__private.initialized:
+            syncing = name in obj._param__private.syncing
             ref, deps, val, is_async = obj.param._resolve_ref(self, val)
             refs = obj._param__private.refs
             if ref is not None:
                 self.owner.param._update_ref(name, ref)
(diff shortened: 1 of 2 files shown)


Hoxbro:
It does not currently work for my case:



https://github.com/holoviz/param/assets/19758978/12884486-72ca-4d32-bd55-c963a7723121




``` python
import panel as pn
import param

pn.extension(notifications=True)

class T(param.Parameterized):
    lst = param.List(allow_refs=True, allow_None=True)

    @param.depends("lst")
    def test(self):
        items = [pn.widgets.TextAreaInput(value=str(i)) for i in range(5)]
        with param.discard_events(self):
            self.lst = pn.rx(items).rx.resolve()
        return pn.Column(*items)


t = T()

pn.widgets.Button(name="print", on_click=lambda e: pn.state.notifications.info(str(t.lst))).servable()
pn.panel(t.test).servable()
```

philippjfr:
Oh, that's not surprising, in fact that's the correct behavior, because what you're doing in that example is effectively resetting the `lst` list every time. If you don't want to do that you need to initialize the widgets given the lst values:

```python
    @param.depends("lst")
    def test(self):
        lst = self.lst or range(5)
        items = [pn.widgets.TextAreaInput(value=str(i)) for i in lst]
        with param.discard_events(self):
            self.lst = pn.rx(items).rx.resolve()
        return pn.Column(*items)
```

Hoxbro:
Thank you. I will test it later.

Issue #937, opened 2024-04-19
`name` of a Parameterized instance overridden when is the default of a `readonly/constant` Parameter


maximlt:
```python
import param

class A(param.Parameterized):
    x = param.Number()

a1 = A(name="FOO1")
a2 = A(name="FOO2")

class B(param.Parameterized):
    const = param.Parameter(a1, constant=True)
    ro = param.Parameter(a2, readonly=True)

b = B()

print(f'{b.const.name=}')
print(f'{b.ro.name=}')
```

Param 1.13:
```
b.const.name='A00005'
b.ro.name='FOO2'
```

Now:
```
b.const.name='A00005'
b.ro.name='A00006'
```

---

The `name` of a Parameterized instance set as the `default` of a `readonly` Parameter is unexpectedly overridden. Bisecting pointed at https://github.com/holoviz/param/pull/776 that stopped automatically setting `instantiate` to `True` when `constant` is `True`.

However, I'm also surprised to see the `name` being updated for a `constant` Parameter! As currently the default value of a Parameter is deep copied only when `instantiate=True`.  So in the example above, `instantiate` is always `False, there's no deep copy of `a1` or `a2`.

https://github.com/holoviz/param/blob/dae8fc4a7b47079468ad5da0b486a779e963351b/param/parameterized.py#L1922-L1926

So the change I'd suggest would be to override the Parameter name only when a deep copy is required which doesn't break any test and would lead to this output for the example:

```
b.const.name='FOO1'
b.ro.name='FOO2'
```

---

The piece of code in `Parameterized._instantiate_param` that overrides the name of a Parameterized class on instantiation originates from [the first commit](https://github.com/holoviz/param/commit/f10b324f6af737fd0173908b1b4d49d6989d8be4#diff-bbc8fd9e66cd584fd9bac415ea1af35c57128c14524e99c7706efdf4a770e555R1164) on Github. One could argue that, even when deep-copying, overriding `name` isn't correct and could be left as a responsibility of the users implementing their Parameterized classes.

---

Did I already say I'd like `name` to be removed from Param? :)

jbednar:
> Did I already say I'd like name to be removed from Param?

[...]

PR #938, opened 2024-04-19
Only override the name of a Parameterized instance on Parameter instantiation when `instantiate=True`

maximlt:
Fixes https://github.com/holoviz/param/issues/937

Logging this change in case we agree on the suggested change in the issue.

- [x] Add tests


philippjfr (approved):
The change makes sense to me.

maximlt:
Unrelated test failure.

Issue #940, opened 2024-05-22
RX: output not updated when indexing a DataFrame with value_start/value_end


maximlt:
Not sure to which extent this is a Panel/Param issue.

```python
import numpy as np
import pandas as pd
import panel as pn

df = pd.DataFrame(np.random.random((100, 2)))
w_i = pn.widgets.IntRangeSlider(start=1, end=99)
dfr = pn.rx(df)
out = dfr[w_i.param.value_start:w_i.param.value_end].head()
out
```

![Monosnap screencast 2024-05-22 14-13-09_gifski](https://github.com/holoviz/param/assets/35924738/c4f615c5-0026-4d2b-80c9-6b09caafb20c)

PR #944, opened 2024-06-13
Recursively resolve references on args and kwargs passed to an reactive operation

maximlt:
Attempts to fix https://github.com/holoviz/param/issues/940

Which can be reproduced with:

```python
import param

start = param.rx(1)
end = param.rx(10)
l = list(range(20))
lrx = param.rx(l)
sx = lrx[start:end]
print(sx.rx.value)  # required to reproduce
start.rx.value = 4
print(sx.rx.value)  # [1, 2, ...], not [4, 5, ...] ! 
```

I think that the reactive expressions of the `start:end` *slice* passed to `__getitem__` were not resolved, which means that no watcher was set to invalidate the expression when `start` is updated. Resolving references recursively fixed the issue and broke no other test, but well, it might be a big hammer 🤔 Anyhow, I've also added another similar test that would have failed without this change, this time with a *tuple* containing reactive expressions passed to Numpy's `__getitem__`.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -196,11 +196,11 @@ def resolve_ref(reference, recursive=False):
     """
     if recursive:
         if isinstance(reference, (list, tuple, set)):
-            return [r for v in reference for r in resolve_ref(v)]
+            return [r for v in reference for r in resolve_ref(v, recursive)]
         elif isinstance(reference, dict):
-            return [r for kv in reference.items() for o in kv for r in resolve_ref(o)]
+            return [r for kv in reference.items() for o in kv for r in resolve_ref(o, recursive)]
         elif isinstance(reference, slice):
-            return [r for v in (reference.start, reference.stop, reference.step) for r in resolve_ref(v)]
+            return [r for v in (reference.start, reference.stop, reference.step) for r in resolve_ref(v, recursive)]
     reference = transform_reference(reference)
     if hasattr(reference, '_dinfo'):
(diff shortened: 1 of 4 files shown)


philippjfr (approved):
I think this is probably a regression introduced when the recursive argument was added to `resolve_ref` and then set to False by default.

Please also look at `resolve_ref` and make sure that recursive is passed down recursively.

param/parameterized.py
```diff
@@ -196,11 +196,11 @@ def resolve_ref(reference, recursive=False):
...
     if recursive:
         if isinstance(reference, (list, tuple, set)):
-            return [r for v in reference for r in resolve_ref(v)]
+            return [r for v in reference for r in resolve_ref(v, recursive)]
         elif isinstance(reference, dict):
-            return [r for kv in reference.items() for o in kv for r in resolve_ref(o)]
+            return [r for kv in reference.items() for o in kv for r in resolve_ref(o, recursive)]
         elif isinstance(reference, slice):
-            return [r for v in (reference.start, reference.stop, reference.step) for r in resolve_ref(v)]
+            return [r for v in (reference.start, reference.stop, reference.step) for r in resolve_ref(v, recursive)]
```
philippjfr, line 203: Think you missed one down below.

maximlt: That one you mean? I wasn't sure about it.

```python
            refs.extend(resolve_ref(arg))
```

Issue #956, opened 2024-07-27
Add .set or .update method to reactive expression


MarcSkovMadsen:
I often have the need to update a reactive expression. Over and over again I find that I lack a .set or .update method. I need that to be able to update via `.bind` or in lambda expressions. Its friction I have to create a set or update function my self.

Please add an `.set` method.

```python
expression.rx.set(value) # same effect as expression.rx.value=value
```

philippjfr:
We originally had this but then removed it because `.rx.value` setting seemed enough but I'd agree. Having a way to update the value with an expression rather than a statement adds a lot of convenience.

PR #957, opened 2024-07-28
Add .rx.set method

philippjfr:
Implements https://github.com/holoviz/param/issues/956

### DIFF
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -391,6 +391,13 @@ class reactive_ops:
 
     # Operations to get the output and set the input of an expression
 
+    def set(self, value):
+        """
+        Sets the input of the pipeline to a new value. Equivalent
+        to ``.rx.value = value``.
+        """
+        self.value = value
+
     @property
     def value(self):
         """
--- a/tests/testreactive.py
+++ b/tests/testreactive.py
@@ -158,10 +158,16 @@ def test_reactive_empty_construct():
     i.rx.value = 2
     assert i.rx.value == 2
 
-def test_reactive_set_new_value():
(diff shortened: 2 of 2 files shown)


param/reactive.py
```diff
@@ -391,6 +391,13 @@ def ternary(condition, _):
```
MarcSkovMadsen, line 391: Nice. Remember to document :-)

param/reactive.py
```diff
@@ -391,6 +391,13 @@ def ternary(condition, _):
 
     # Operations to get the output and set the input of an expression
 
+    def set(self, value):
+        """
+        Sets the input of the pipeline to a new value. Equivalent
+        to ``.rx.value = value``.
```
maximlt, line 397:
```suggestion
        to `.rx.value = value`.
```

maximlt (commented):
@philippjfr is this ready?

jbednar:
Are people ok with the linguistic ambiguity between "set" as a collection of items and "set" as "assign"? I find `.rx.assign` to  more obviously indicate the latter meaning, but it's maybe less guessable? I.e. if I think about what I want to do to assign a value, "set" is indeed the word I think of, but then when I encounter `.rx.set`, I'm not sure which of the two meanings it would be. Happy with whatever other people agree on...

maximlt:
FWIW I'm fine with the naming of `.rx.set` (`.rx.pipe(set | list | ...)` being I believe the correct way to cast).

MarcSkovMadsen:
I like set because its the term used in React Docs.

maximlt:
I'm ok with `rx.set()`. How about you @philippjfr?

Issue #958, opened 2024-08-07
Annotate commonly used param decorators


gandhis1:
See https://github.com/holoviz/panel/issues/7078 for more details

The highest ROI public interface typing improvements tend to be decorators as untyped decorators basically nuke the user's own type information, thus undermining static analysis checks in their own code.

List of decorators:

- `param.depends`
- `param.parametrized.output`
- `param.parameterized.iscoroutinefunction` (needs to be a `TypeGuard` just like the `inspect` variant)
- Any others that are widely used?

FYI @MarcSkovMadsen

PR #959, opened 2024-08-07
Annotate depends and accept_arguments decorators

gandhis1:
Addresses https://github.com/holoviz/param/issues/958

@MarcSkovMadsen @philippjfr


MarcSkovMadsen:
Was this closed on purpose @gandhis1 ?

gandhis1:
Yes, it needs more work / has some issues in its current state

Issue #960, opened 2024-08-14
Make ParameterizedFunction copy parameter references


Dsolnik:
In the constructor for a [ParameterizedFunction](https://github.com/holoviz/param/blob/31f5717457b8a53e979f5e5cc5e0c724c69f7f61/param/parameterized.py#L4453), all of the values at the current point in time are copied. Unfortunately, this doesn't copy over references.

#### Is your feature request related to a problem? Please describe.
It's frustrating that I need to add back in all the different previous parameters back.

#### Describe the solution you'd like
I'd like to copy over references, or add a flag to copy them.

#### Describe alternatives you've considered
```python
def instance(fn, **kwargs):
    params = fn.param.objects()
    params.update(**kwargs)
    return obj.instance(**params)
```

Maybe we could make an option to the instance method/

Dsolnik:
https://github.com/holoviz/param/pull/961

philippjfr:
Sorry you never got a reply here or on your PR. In general I think this breaks the contract of an instance. Using refs like this would link one instance to all other instances, which seems unexpected. Adding a flag may be okay. Wouldn't blame you if you've forgotten or no longer care but can you describe the motivating use case?

PR #961, opened 2024-08-14
pass references along in `ParameterizedFunction.instance`

Dsolnik:
Previously, only all of the values at the current point in time are copied. Unfortunately, this doesn't copy over references. It would be nice to copy those over too.


Dsolnik:
https://github.com/holoviz/param/issues/960

PR #962, opened 2024-08-14
Annotate depends and accept_arguments decorators

gandhis1:
It took some tinkering but I finally have this working. Addresses https://github.com/holoviz/param/issues/958

@MarcSkovMadsen @philippjfr 

Results:

```
>> mypy param/_utils.py --follow-imports=silent
Success: no issues found in 1 source file
```

```
>> mypy param/depends.py --follow-imports=silent
param/depends.py:142: error: Item "str" of "Parameter | str" has no attribute "owner"  [union-attr]
param/depends.py:144: error: Item "str" of "Parameter | str" has no attribute "owner"  [union-attr]
param/depends.py:144: error: Item "str" of "Parameter | str" has no attribute "name"  [union-attr]
```

### DIFF
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -1,22 +1,28 @@
+from __future__ import annotations
+
 import asyncio
 import collections
 import contextvars
 import datetime as dt
-import inspect
 import functools
+import inspect
 import numbers
 import os
 import re
 import sys
 import traceback
 import warnings
-
-from collections import defaultdict, OrderedDict
+from collections import OrderedDict, abc, defaultdict
 from contextlib import contextmanager
 from numbers import Real
 from textwrap import dedent
 from threading import get_ident
-from collections import abc
+from typing import TYPE_CHECKING, Callable, Concatenate, ParamSpec, TypeVar
+
+if TYPE_CHECKING:
+    _P = ParamSpec("_P")
+    _R = TypeVar("_R")
+    CallableT = TypeVar("CallableT", bound=Callable)
 
 DEFAULT_SIGNATURE = inspect.Signature([
     inspect.Parameter('self', inspect.Parameter.POSITIONAL_OR_KEYWORD),
@@ -282,12 +288,14 @@ def flatten(line):
             yield element
 
 
-def accept_arguments(f):
+def accept_arguments(
+    f: Callable[Concatenate[CallableT, _P], _R]
+) -> Callable[_P, Callable[[CallableT], _R]]:
     """
     Decorator for decorators that accept arguments
     """
     @functools.wraps(f)
-    def _f(*args, **kwargs):
+    def _f(*args: _P.args, **kwargs: _P.kwargs) -> Callable[[CallableT], _R]:
         return lambda actual_f: f(actual_f, *args, **kwargs)
     return _f
 
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -22,8 +22,8 @@ from typing import TYPE_CHECKING, Callable, TypeVar
 if TYPE_CHECKING:
     from typing_extensions import Concatenate, ParamSpec
 
-    _P = ParamSpec("_P")
-    _R = TypeVar("_R")
+    P = ParamSpec("P")
+    R = TypeVar("R")
     CallableT = TypeVar("CallableT", bound=Callable)
 
 DEFAULT_SIGNATURE = inspect.Signature([
@@ -291,13 +291,13 @@ def flatten(line):
 
 
 def accept_arguments(
-    f: Callable[Concatenate[CallableT, _P], _R]
-) -> Callable[_P, Callable[[CallableT], _R]]:
+    f: Callable[Concatenate[CallableT, P], R]
+) -> Callable[P, Callable[[CallableT], R]]:
     """
     Decorator for decorators that accept arguments
     """
     @functools.wraps(f)
-    def _f(*args: _P.args, **kwargs: _P.kwargs) -> Callable[[CallableT], _R]:
+    def _f(*args: P.args, **kwargs: P.kwargs) -> Callable[[CallableT], R]:
         return lambda actual_f: f(actual_f, *args, **kwargs)
     return _f
 
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -17,9 +17,11 @@ from contextlib import contextmanager
 from numbers import Real
 from textwrap import dedent
 from threading import get_ident
-from typing import TYPE_CHECKING, Callable, Concatenate, ParamSpec, TypeVar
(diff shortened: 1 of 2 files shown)


param/depends.py
```diff
@@ -1,16 +1,46 @@
...
+@overload
+def depends(
+    *dependencies: Parameter, watch: bool = ..., on_init: bool = ..., **kw: Parameter
+) -> Callable[[CallableT], DependsFunc[CallableT]]:
+    ...
 
 @accept_arguments
-def depends(func, *dependencies, watch=False, on_init=False, **kw):
+def depends(
+    func: CallableT, /, *dependencies: Dependency, watch: bool = False, on_init: bool = False, **kw: Parameter
```
gandhis1, line 42: Without the `/`, Mypy complains:

```
param/depends.py:40: error: Argument 1 to "accept_arguments" has incompatible type "Callable[[CallableT@depends, VarArg(Parameter | str), DefaultNamedArg(bool, 'watch'), DefaultNamedArg(bool, 'on_init'), KwArg(Parameter)], Callable[[CallableT@depends], DependsFunc[CallableT@depends]]]"; expected "Callable[[CallableT, VarArg(Parameter | str), DefaultNamedArg(bool, 'watch'), DefaultNamedArg(bool, 'on_init'), KwArg(Parameter)], Callable[[CallableT], DependsFunc[CallableT]]]"  [arg-type]
param/depends.py:40: note: This is likely because "depends" has named arguments: "func". Consider marking them positional-only
```

This looks like a Mypy bug; variadic arguments via `*` should necessarily mean that `func` is positional-only. However, adding in `/` I believe makes no meaningful difference at runtime, so I just went ahead and did that.

param/depends.py
```diff
@@ -1,16 +1,46 @@
...
+@overload
+def depends(
+    *dependencies: str, watch: bool = ..., on_init: bool = ...
+) -> Callable[[CallableT], DependsFunc[CallableT]]:
+    ...
+
+@overload
+def depends(
+    *dependencies: Parameter, watch: bool = ..., on_init: bool = ..., **kw: Parameter
+) -> Callable[[CallableT], DependsFunc[CallableT]]:
```
gandhis1, line 37: The intention with these overloads is to prevent usage that violates the validation that `depends` applies internally at runtime, for example, this fails:

```python
@depends("test", foobar=Parameter())
def test(a: int, b: str) -> None:
    print("foobar")
```

```
param/depends.py:154: error: No overload variant of "depends" matches argument types "str", "Parameter"  [call-overload]
param/depends.py:154: note: Possible overload variants:
param/depends.py:154: note:     def depends(*dependencies: str, watch: bool = ..., on_init: bool = ...) -> Callable[[CallableT], DependsFunc[CallableT]]
param/depends.py:154: note:     def depends(*dependencies: Parameter, watch: bool = ..., on_init: bool = ..., **kw: Parameter) -> Callable[[CallableT], DependsFunc[CallableT]]
```

While this works:

```
@depends(Parameter(), foobar=Parameter())
def test(a: int, b: str) -> None:
    print("foobar")
```

gandhis1:
Is it intentional that this project is still on Python 3.8 while `panel` is on python 3.10? The python 3.8 build is failing here

param/_utils.py
```diff
@@ -1,22 +1,30 @@
...
 from numbers import Real
 from textwrap import dedent
 from threading import get_ident
-from collections import abc
+from typing import TYPE_CHECKING, Callable, TypeVar
+
+if TYPE_CHECKING:
+    from typing_extensions import Concatenate, ParamSpec
+
+    _P = ParamSpec("_P")
```
hoxbro, line 25: Why do you use underscore here? Is this the norm? 

It is "hidden" behind `TYPE_CHECKING`, so it will be private for users. It also adds extra overhead to reading the type hinting itself.

hoxbro (commented):
> Is it intentional that this project is still on Python 3.8 while panel is on python 3.10?

Yes. Panel depends on projects that have dropped support for Python 3.9. Param is more of a fundamental tool, so we don't want to drop Python versions before it is necessary. See [HEP1](https://github.com/holoviz/holoviz/blob/main/hep/hep1.md)

param/_utils.py
gandhis1: It is typical for `typeshed`, for example: https://github.com/python/typeshed/blob/main/stdlib/urllib/request.pyi#L51

> It is "hidden" behind TYPE_CHECKING, so it will be private for users.

Not if they are also using `TYPE_CHECKING`:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from param._utils import P
```

And regardless of `TYPE_CHECKING` it still becomes visible in IDE auto-complete, interspersed with symbols that are part of the public API (as opposed to being namespaced with an underscore prefix):

image

(Granted, underscore is just a convention, they could just as well import it anyway.)

That said, I don't really care either way, so I'll just remove the underscores

philippjfr (approved):
Really appreciate this, thanks for all your efforts so far!

param/_utils.py
hoxbro: Thank you for the clarification. 

Whatever the norm should be done, and if you say it is making them private, then so be it.  

If you want to avoid repeating complex types, you are welcome to create a `_typing.py` file, though it does not need to be in this PR. 

Do you have a preferred way to test type hints? If possible, I would like your help in setting it up. 

And like Philipp said, I also appreciate your work on these PRs! :100:

gandhis1: > Do you have a preferred way to test type hints? 

For these PRs I generally am just running `mypy path/to/file/changed.py --follow-imports=silent` and making sure that the errors after are a strict subset of the errors that existed before. I also just create some dummy code that uses the decorator to see whether it works as expected.

For something more automated - and I'm mostly thinking in the context of `panel` - we could set up a Mypy run that's part of CI. We could start with some subset of the `examples` folder that we can get zero issues on, then extend to all the examples, and after we reach 100% no issues, we can switch to Mypy `strict` and repeat. The point would be to prioritize the public API - which `examples` would encompass - over the internal library code.

I figure `examples` would work a lot better than me manually creating functions in test code and applying `@param.depends` etc. on it.

Only after that would I try to Mypy `panel` or `param` internal code as that is a much bigger lift and could require some actual code changes to be type-compatible, etc. So we'd probably want to run the examples CI check with `--follow-imports=silent` so that it doesn't prematurely start checking internal code.

hoxbro: I have opened a PR for Panel here https://github.com/holoviz/panel/pull/7151

hoxbro (approved):

gandhis1:
Mergeable? Build seems to be flaky, assuming there's no good reason for it to fail on 3.12 windows and pass on 3.12 mac

PR #966, opened 2024-08-31
refactor: Don’t directly import numpy and add `gen_types` decorator

hoxbro:
The changes made in this PR are not to directly import non-essential imports (numpy in param) but to only check for the case when the user has already imported them.

This comes down to two things:

1) For param, this is to "hide" numpy imports in functions or generators. This is strictly not necessary, but people check import time with `python -X importtime -c 'import param'`, and because we do an actual import of numpy, we are heavily penalized by it. The random modules (numpy and stdlib) also seem unnecessary to import, only to use the import for not outputting anything with `pprint`. 

2) The second part involves using a new decorator, which overloads `isinstance` and `issubclass` and still makes it function like a generator with and without calling it. This approach can be used in our other libraries, e.g., HoloViews. Here, we currently import pandas to set up different class selectors, which heavily penalizes the import time. We currently require pandas, but that may not always be the case.

### DIFF
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -620,7 +620,10 @@ class _GeneratorIsMeta(type):
     def __subclasscheck__(cls, sub):
         return issubclass(sub, tuple(cls.types()))
 
-class _GeneratorIs(metaclass=_GeneratorIsMeta): ...
+class _GeneratorIs(metaclass=_GeneratorIsMeta):
+    @classmethod
+    def __iter__(cls):
+        yield from cls.types()
 
 def gen_types(gen_func):
     """Decorator that creates a type using _GeneratorIsMeta with a specified gen_types function."""
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -616,11 +616,11 @@ def async_executor(func):
 
 def anyinstance(obj, class_tuple_generator):
     if inspect.isgeneratorfunction(class_tuple_generator):
-        return any(isinstance(obj, i) for i in class_tuple_generator())
+        class_tuple_generator = tuple(class_tuple_generator())
     return isinstance(obj, class_tuple_generator)
 
 
 def anysubclass(obj, class_tuple_generator):
     if inspect.isgeneratorfunction(class_tuple_generator):
-        return any(issubclass(obj, i) for i in class_tuple_generator())
+        class_tuple_generator = tuple(class_tuple_generator())
     return issubclass(obj, class_tuple_generator)
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -612,3 +612,15 @@ def async_executor(func):
         task.add_done_callback(_running_tasks.discard)
     else:
         event_loop.run_until_complete(func())
+
+
+def anyinstance(obj, class_tuple_generator):
+    if inspect.isgeneratorfunction(class_tuple_generator):
+        return any(isinstance(obj, i) for i in class_tuple_generator())
+    return isinstance(obj, class_tuple_generator)
+
+
+def anysubclass(obj, class_tuple_generator):
+    if inspect.isgeneratorfunction(class_tuple_generator):
+        return any(issubclass(obj, i) for i in class_tuple_generator())
+    return issubclass(obj, class_tuple_generator)
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -612,3 +612,19 @@ def async_executor(func):
         task.add_done_callback(_running_tasks.discard)
     else:
         event_loop.run_until_complete(func())
+
+class _GeneratorIsMeta(type):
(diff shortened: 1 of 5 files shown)


param/parameters.py
```diff
@@ -109,7 +111,7 @@ def guess_param_types(**kwargs):
         elif isinstance(v, tuple):
             if all(_is_number(el) for el in v):
                 params[k] = NumericTuple(**kws)
-            elif all(isinstance(el, dt_types) for el in v) and len(v)==2:
+            elif all(anyinstance(el, _dt_types) for el in v) and len(v)==2:
```
hoxbro, line 114:
```suggestion
            elif len(v) == 2 and all(anyinstance(el, _dt_types) for el in v):
```

param/parameterized.py
```diff
@@ -3838,17 +3851,16 @@ def container_script_repr(container,imports,prefix,settings):
     return rep
 
 
-def empty_script_repr(*args): # pyflakes:ignore (unused arguments):
-    return None
+def _np_random():
+    if "numpy" in sys.modules:
```
hoxbro, line 3855:
```suggestion
    if "numpy.random" in sys.modules:
```

param/_utils.py
```diff
@@ -612,3 +612,15 @@ def async_executor(func):
         task.add_done_callback(_running_tasks.discard)
     else:
         event_loop.run_until_complete(func())
+
+
+def anyinstance(obj, class_tuple_generator):
```
maximlt, line 617: Could an approach like this be used to avoid creating these functions? https://peps.python.org/pep-3119/#overloading-isinstance-and-issubclass

maximlt: It seems it'd be something like:

```python

import sys

class GeneratorIsMeta(type):
    def __instancecheck__(cls, inst):
        return isinstance(inst, tuple(cls.gen_types()))

    def __subclasscheck__(cls, sub):
        return issubclass(sub, tuple(cls.gen_types()))


class DtTypes(metaclass=GeneratorIsMeta):
    @classmethod
    def gen_types(cls):
        yield dt.datetime
        yield dt.date
        if "numpy" in sys.modules:
            import numpy as np
            yield np.datetime64

class IntTypes(metaclass=GeneratorIsMeta):
    @classmethod
    def gen_types(cls):
        yield int
        if "numpy" in sys.modules:
            import numpy as np
            yield np.integer
```

hoxbro: Thank you. 

I still thinks it is pretty advanced (even though you don't need it to be iterable in your example). 

You could likely move the advanced logic into a decorator, but then you would need to import that into our other libraries. My main point with accepting iterator is not so much for param itself but for HoloViews.

maximlt: > My main point with accepting iterator is not so much for param itself but for HoloViews.

Oh yeah I agree users shouldn't have to pass this custom class, a generator seems appropriate. Instead I was wondering if there couldn't be a way to use the metaclass approach internally to avoid the special `anyisinstance` and `anyissubclass` functions, they seem to be easy to forget in the future.

hoxbro: I have switched to a metaclass approach with a decorator for easy access.

param/parameters.py
```diff
@@ -2184,18 +2184,20 @@ def _validate(self, val):
     def _validate_class_(self, val, class_, is_instance):
         if (val is None and self.allow_None):
             return
+        if (is_instance and isinstance(val, class_)) or (not is_instance and issubclass(val, class_)):
+            return
+
         if isinstance(class_, tuple):
-            class_name = ('(%s)' % ', '.join(cl.__name__ for cl in class_))
+            class_name = ('({})'.format(', '.join(cl.__name__ for cl in class_)))
+        elif inspect.isgenerator(class_):
```
hoxbro, line 2192: Likely I need to update this with the metaclass aproach.

philippjfr (approved):
Looks good, just had one comment.

param/parameterized.py
```diff
@@ -3839,17 +3852,13 @@ def container_script_repr(container,imports,prefix,settings):
...
+def _no_script_repr():
     # Suppress scriptrepr for objects not yet having a useful string representation
-    import numpy
-    script_repr_reg[random.Random] = empty_script_repr
-    script_repr_reg[numpy.random.RandomState] = empty_script_repr
-
-except ImportError:
-    pass # Support added only if those libraries are available
+    if random := sys.modules.get("random"):
+        yield random.Random
```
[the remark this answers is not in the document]
hoxbro: `import random` was an import at the top of the file.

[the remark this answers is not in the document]
philippjfr: I missed the walrus operator, ignore me.

PR #969, opened 2024-10-07
Added missing `super().__init_subclass__` call in `_ParameterBase.__init_subclass__`

JRRudy1:
Currently the `_ParameterBase.__init_subclass__` method does not forward to the method on its superclass, so any other base classes in the MRO do not have a chance to perform any necessary subclass initialization steps. This PR adds the missing call.

This is particular relevant when defining a `Parameter` subclass with generic type parameters by inheriting from `typing.Generic` (or equivalently by using the new generics syntax added in Python 3.12). Inheriting from `typing.Generic[T]` where `T` is a `typing.TypeVar` typically allows a class to be parameterized by a concrete type, but this does not work when `_ParameterBase` is in the MRO since the `__init_subclass__` call is not properly propagated up the MRO.

The following example raises an `AttributeError` in the current `param` release, but works correctly after this PR:
```python

from typing import Generic, TypeVar
from param import Selector

T = TypeVar('T')

class GenericSelector(Selector, Generic[T]):
    pass

# raises `AttributeError: type object 'GenericSelector' has no attribute '__parameters__'`:
IntSelector = GenericSelector[int]
```

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -971,6 +971,7 @@ class _ParameterBase(metaclass=ParameterMetaclass):
 
     @classmethod
     def __init_subclass__(cls):
+        super().__init_subclass__()
         # _update_signature has been tested against the Parameters available
         # in Param, we don't want to break the Parameters created elsewhere
         # so wrapping this in a loose try/except.


MarcSkovMadsen:
If using generics does mypy and other type checkers then know an IntSelector parameter is an int?

JRRudy1:
> If using generics does mypy and other type checkers then know an IntSelector parameter is an int?

Probably not... unfortunately the `param` library doesn't seem to integrate very well with static type checkers, but hopefully it will get better someday (I was about to link issue #&#8203;376 before realizing that was you haha). I can often trick my IDE (PyCharm) by declaring params like `a: int = IntSelector(...)`, or `a: T = GenericSelector(...)` in a generic context, though it is far from perfect and probably doesn't fly with mypy. But even in cases where generics (and type-hints in general) don't serve a practical purpose, they can often significantly help with readability/maintainability anyways.

That said, the motivation for this PR isn't really specific to handling generic classes, that just happened to be the case that led me here.

maximlt (approved):
This looks like an oversight, thanks!

PR #970, opened 2024-10-10
Declare compatibility with Python 3.13

maximlt:
Let's see!

PR #971, opened 2024-10-14
ci: Update CI to use Pixi

hoxbro:
Changes the developer/CI from Hatch to Pixi. This will align it with our other repos.

Supersede #970


doc/developer_guide.md
```diff
@@ -1,41 +1,189 @@
-# Developer guide
+# Developer Guide
```
hoxbro, line 1: This has been copied from Panel. I should have updated it to match the param, but there could potentially be something I haven't updated correctly.

.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 # This is the configuration for pre-commit, a local framework for managing pre-commit hooks
 #   Check out the docs at: https://pre-commit.com/
 
-default_stages: [commit]
+default_stages: [pre-commit]
```
hoxbro, line 1: Stop emitting a warning.

doc/user_guide/Parameter_Types.ipynb
```diff
@@ -717,8 +717,8 @@
     "class P(param.Parameterized):\n",
     "    p = param.Path('Parameter_Types.ipynb')\n",
     "    f = param.Filename('Parameter_Types.ipynb')\n",
-    "    d = param.Foldername('lib', search_paths=['/','/usr','/share'])\n",
-    "    o = param.Filename('output.csv', check_exists=False)\n",
+    "    d = param.Foldername('lib', search_paths=['/','/usr','/share'], check_exists=False)\n",
```
hoxbro, line 720: Changes made in this file are so we can run this notebook on Windows, too.

pyproject.toml
```diff
@@ -39,13 +40,7 @@ examples = [
     "pandas",
     "panel",
 ]
-doc = [
```
hoxbro, line 42: I have removed the developer-focused optional dependencies `doc` and `lint`.

tests/testreactive.py
```diff
@@ -590,7 +590,7 @@ def gen():
 async def test_reactive_gen_pipe():
     def gen(val):
         yield val+1
-        time.sleep(0.05)
+        time.sleep(0.1)
```
hoxbro, line 593: Avoid the flaky Windows test; use the same fix as https://github.com/holoviz/param/pull/967.

.github/workflows/docs.yaml
```diff
@@ -2,50 +2,71 @@ name: docs
...
-      - name: graphviz
-        run: sudo apt install graphviz graphviz-dev
-      - name: env setup
-        run: |
-          python -m pip install --upgrade pip
-          python -m pip install hatch
-      - name: build docs
-        run: hatch -v run docs:build
+
+  docs_publish:
```
maximlt, line 58: Why is this workflow divided in two jobs?

.github/workflows/nightly_lock.yaml
maximlt, line 1: I wonder to which extent this is needed for repositories that a low development activity like Param. I guess it doesn't hurt so much.

.github/workflows/test.yaml
```diff
@@ -5,71 +5,128 @@ on:
...
+      fail-fast: false
+      matrix: ${{ fromJson(needs.setup.outputs.matrix) }}
+    timeout-minutes: 30
+    steps:
+      - uses: holoviz-dev/holoviz_tasks/pixi_install@v0
+        with:
+          environments: ${{ matrix.environment }}
+      - name: Test Unit
+        run: |
+          pixi run -e ${{ matrix.environment }} test-unit $COV
```
maximlt, line 98: Maybe add a `test-unit-cov` command? This is also useful to run as a developer and not just for the CI.

pixi.toml
```diff
@@ -0,0 +1,138 @@
...
+panel = "*"
+
+# =============================================
+# =================== TESTS ===================
+# =============================================
+[feature.test-core.dependencies]
+pytest = "*"
+pytest-asyncio = "*"
+pytest-cov = "*"
+pytest-github-actions-annotate-failures = "*"
```
maximlt, line 65: Not a big fan of this plugin. It's also definitely not part of the `core` deps needed to tun the tests.

scripts/conda/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+PACKAGE="param"
+
+python -m build -s .
+
+VERSION=$(python -c "import $PACKAGE; print($PACKAGE._version.__version__)")
```
maximlt, line 9: Out of curiosity, is there a reason this is not `PACKAGE.__version__`?

hoxbro: This is to be sure that the automatic version of the file is generated. `param.__version__` is more "complex" with `try/except`.

.github/workflows/docs.yaml
maximlt: Ah ok, I've never seen that fail before during a docs build. Do you mind simplifying this workflow into one single job? Let's keep things simple!

.github/workflows/build.yaml
```diff
@@ -2,81 +2,141 @@ name: packages
...
+defaults:
+  run:
+    shell: bash -e {0}
+
+env:
+  PYTHON_VERSION: "3.11"
+  PACKAGE: "param"
 
 jobs:
+  waiting_room:
```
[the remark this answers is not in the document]
maximlt: > I have limited it to lead maintainers (and me).

Ok. Where is the least of lead maintainers defined?

> I use it to double-check that the version tag is correct.

Too bad we no longer trust the system to produce the right version tag, but that's fine.

.github/workflows/nightly_lock.yaml
maximlt: > Away to get a snapshot of the environment. This can be used to see what package caused a failure. It could also be used to see how the param environments looked a year ago.

Ah yes I missed that the lock file was pushed to S3 too. Sounds useful, when you know about it ;)

.github/workflows/test.yaml
maximlt: I liked when you said that you moved as much as possible from the config that was on the CI workflows to the pixi config file (e.g. env vars) which is why I was a little surprised to see that there in `test.yaml`.

> I'm cautious about adding more tasks, which could confuse new contributors.

My experience so far has been to copy/paste the commands displayed in the developer guide. I have not relied on `pixi task list` which I think just gives a simple list of tasks available without their definition, and that's pretty cluttered already anyway.

I'm fine with how things are anyway here.

pixi.toml
maximlt: > I'm a big fan of it. Reduces the need to look into the logs.

My experience of it so far has been to give me the impression that, basically, every Panel PR is failing as it also displays errors that occur during flaky tests.

---

> The same could be set about pytest-cov. The plugin is a no-op if you are not in a Github action and only have pytest as a dependency.

Reminds me of the discussion we recently had about `hatch-conda-build ` :D `pytest-cov` can be used by a user, I don't think `pytest-github-actions-annotate-failures` can?

But anyway, I'm again fine with how this is.

pyproject.toml
```diff
@@ -111,118 +101,15 @@ include = [
...
-
 [tool.pytest.ini_options]
+addopts = "--color=yes"
 python_files = "test*.py"
 filterwarnings = [
   "error",
 ]
 xfail_strict = "true"
 asyncio_mode = "auto"
+asyncio_default_fixture_loop_scope="funcion"
```
maximlt, line 112: Typo:

```suggestion
asyncio_default_fixture_loop_scope="function"
```

maximlt (commented):
Except a typo and a simplification I'd like to see in the docs workflow, I'm fine with these changes.

@hoxbro I would strongly encourage you to update the HoloViz operating guide (https://holoviz.org/contributing.html#operating-guide) now that the workflows you created are used in the majority of places and `pyctdev` is no longer used in the main repositories. There are many behaviors and explanations as to why things are how they are that are just in your mind and that would be worth sharing with others, current and future contributors/maintainers. If you have custom tooling that is used to inspect/interact with the system you created, it would also be nice to share it in a HoloViz repo.

.github/workflows/build.yaml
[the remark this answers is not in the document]
maximlt: > PyPi also strongly encourages it ([ref](https://docs.pypi.org/trusted-publishers/using-a-publisher/)).

I'm sorry I don't understand, this page links to using a trusted publisher. Are we doing that?

pixi.toml
maximlt: > The correct thing to be frustrated about is the flaky test themselves 🙂

Both things are frustrating.

maximlt (approved):
Thanks @hoxbro for doing this work! I'm going to get this merged now as I'd like the next release to use this new infrastructure.

I don't necessarily agree with the increased complexity some workflows got (e.g. publishing waiting room, docs build in two jobs) but I'm not the one doing all the hard work setting up this infra and maintaining it across HoloViz, so it's alright!!

.github/workflows/build.yaml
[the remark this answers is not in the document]
hoxbro: > I'm sorry I don't understand, this page links to using a trusted publisher. Are we doing that?

Just to be sure: 

- The waiting room job itself is to ensure that all the building steps succeed. For example, if the PyPi build step fails, the conda package will _not_ be released. 
- Inside the waiting room is an environment `publish`, which people must be approved before release. This was what I tried to reference in the PyPI ref.
image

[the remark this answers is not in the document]
maximlt: Ok thanks I got it.

> The waiting room job itself is to ensure that all the building steps succeed. For example, if the PyPi build step fails, the conda package will not be released.

It sounds like it's more for manually checking the two builds, since otherwise it could be automated.

> Inside the waiting room is an environment publish, which people must be approved before release. This was what I tried to reference in the PyPI ref.

Ok I understand better how you use the environment to add this protection rule. Seems not so related to the trusted publisher thing for which `environment` is optional but `permissions` with `id-token: write` is not.

image

By the way I think you should add Jean-Luc and we probably need to update the list of maintainers.

image

All of that sounds like it would be worth being documented somewhere, it's kind of hard to follow.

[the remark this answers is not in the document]
hoxbro: > manually checking the two builds

It is automatically done. If the environment were not there, it would do everything without interaction. 

> All of that sounds like it would be worth being documented somewhere

Yes. I will update the holoviz site (soon).

PR #975, opened 2024-11-08
Extend DataFrame parameter support to other libraries

philippjfr:
As we begin to broaden support for different DataFrame libraries in our ecosystem we need to start at the bottom and work our way up. This PR makes it possible for `param.DataFrame` to be extended with support for other DataFrame libraries and adds initial support for accepting `polars` `DataFrame` and `LazyFrame` objects. For now we maintain backwards compatibility here and require the user to explicitly define which libraries they want to support by providing a `libraries` argument.


param/parameters.py
```diff
@@ -2294,31 +2298,54 @@ class DataFrame(ClassSelector):
...
         self._validate(self.default)
 
+    @property
+    def class_(self):
+        types = ()
+        if 'pandas' in self.libraries and 'pandas' in sys.modules:
+            import pandas as pd
+            types += (pd.DataFrame,)
+        if 'polars' in self.libraries and 'polars' in sys.modules:
+            import polaras as pl
```
hoxbro, line 2339:
```suggestion
            import polars as pl
```

philippjfr:
@hoxbro @MarcSkovMadsen would love your thoughts. I'd like to think about Narwhals support but am wondering if that may be better implemented as a new parameter type.

param/parameters.py
```diff
@@ -2374,6 +2401,8 @@ def _validate(self, val):
     def serialize(cls, value):
         if value is None:
             return None
+        if hasattr(value, 'to_dicts'):
```
hoxbro, line 2404:
```suggestion
        if hasattr(value, 'collect'):
            value = value.collect()  # Polars LazyFrame
        if hasattr(value, 'to_dicts'):
```

MarcSkovMadsen:
My POC for how this should be done from param over panel to bokeh is in https://github.com/panel-extensions/panel-graphic-walker/blob/89d6bac66d119d6ad1f75dfe27ec984d18d895ed/src/panel_gwalker/_tabular_data.py#L1

My broader vision for tabular support is in https://github.com/narwhals-dev/narwhals/issues/1289#issuecomment-2453285606.

**My recommendation would by to first implement this e2e in panel-graphic-walker**. That is a place where its really easy to get this working and experimenting e2e. Its also a practical example that will help us have a specific discussion solving real problems.

philippjfr:
@hoxbro @maximlt I know you two discussed this before and your initial suggestion was that we should introduce a new Parameter class, because this is potentially confusing and over-complicated. I'm wary of creating a new parameter given the history of situations like CalendarDate, Date and the ObjectSelector, Selector mess, which in my experience also just leads to confusion.

Overloading the Parameter does carry risk of confusion too and backward compatibility is a bit hard to reason about. So my concerns with this implementation are:

- We now allow any (eager) DataFrame type by default. This does break backward compatibility.
- If you don't have Narwhals installed then it falls back to pandas. This is potentially confusing since support for e.g. Polars, depends on another library being installed. At minimum I guess we need to improve the error messages to suggest installing Narwhals if it falls down the pandas validation path.
- Should we even attempt to validate rows and columns if `allow_lazy=True`.
- What do we do about (de)serialization? By default it'll always use pandas during deserialization now.

ghostiee-11:
Hey @philippjfr,

Implementing `DataFrameLike` as a new parameter class, separate from `DataFrame`, per the agreed direction. Mapping it to each concern raised in this thread so nothing is left open:

**Backward compatibility (breaking the pandas contract).** `DataFrame` is not touched. `param.DataFrame()` still accepts pandas and only pandas, so code that calls `.iloc` / `.groupby` without a type check keeps working. No migration, opt-in flag, or deprecation cycle is needed because nothing existing changes.

**Different APIs / "isinstance everywhere".** `DataFrameLike` passes the native object through unchanged; it does not silently wrap in Narwhals. Authors who want a unified API call `nw.from_native(self.df)` themselves; pandas-only authors keep using `DataFrame`. The parameter makes "I accept any frame" declarable without hiding the API difference.

**Narwhals-optional / confusing pandas fallback.** `DataFrameLike` requires Narwhals by definition. No silent fallback to pandas validation. If Narwhals is missing, instantiation raises naming the missing dependency, so backend support is never secretly conditional on another install.

**Validating rows/columns when `allow_lazy=True`.** Column and ordered checks still run on LazyFrames (`nw` exposes `.columns` without collecting). Row-count validation is a no-op for lazy frames; `.collect()` is never called implicitly. Documented in the docstring.

**(De)serialization.** The one point I want a steer on. Native round-trip via the original library by default, with a Narwhals backend-agnostic path available, is my lean. Flagging it explicitly rather than deciding unilaterally.

**Class proliferation (CalendarDate/Date, ObjectSelector/Selector).** That confusion came from names reading as interchangeable for one concept. `DataFrame` ("strict pandas") and `DataFrameLike` ("anything Narwhals recognises") name a real semantic split, so it should not reproduce that.

**Lazy handling without hasattr probing.** Detection goes through `nw.from_native` (which classifies eager vs lazy across pandas, Polars, PyArrow, Dask, DuckDB, cuDF, Modin) rather than `hasattr(value, 'collect')` checks.

**panel-graphic-walker prototype.** Validating the same API there in parallel since it already has a Narwhals tabular layer to integrate against.

Opening a draft PR with the implementation plus tests shortly so there is running code to review against. `DataFrame` stays exactly as is.

PR #977, opened 2024-11-13
Improved param and rx docs

MarcSkovMadsen:
Attempt to make `.param`, `.rx` and `.rx()` easier to use by improving docstrings.

Currently I meet a lot of dead ends because of missing or very sparse docstrings.

For example `.rx` in example below has no docstring which makes it almost impossible to use without already knowing how.

```python
import param

class MyClass(param.Parameterized):
    value = param.Parameter()

my_class = MyClass(value=0)
rx_value = my_class.param.value.rx
```

Its the same for `.param`. There is no docstring.

I do believe it would help users tremendously if there where docstring and improved docstrings. It would make the `.param` and `.rx` namespaces so much more inviting and easy to use.

## Guiding Principles

- Use Numpy style docstrings (would strongly prefer google docstrings because they are better as tooltips)
- Add types where they are easy to add
- Document in style similar to Panel docstrings
  - Reference link
  - Basic examples that can run
- Ok to make smaller refactors of code if they can be done without breaking changes.


param/reactive.py
```diff
@@ -755,14 +787,56 @@ def __init__(
         ]
         self._setup_invalidations(depth)
         self._kwargs = kwargs
-        self.rx = reactive_ops(self)
+        self._rx = reactive_ops(self)
```
MarcSkovMadsen, line 790: This enables me to provide a docstring. Don't know how to do that without it.

MarcSkovMadsen:
As I am not a regular contributor to param it would help me to get a review, feedback and merge of this "smaller" PR.

Then I would feel confident making a larger PR more systematically going through `.param` and `.rx` namespaces and improving the docstrings and types.

param/parameterized.py
```diff
@@ -2611,7 +2759,38 @@ def serialize_value(self_, pname, mode='json'):
         serializer = Parameter._serializers[mode]
         return serializer.serialize_parameter_value(self_or_cls, pname)
 
-    def deserialize_parameters(self_, serialization, subset=None, mode='json'):
+    def deserialize_parameters(self_, serialization, subset=None, mode='json') -> dict:
```
MarcSkovMadsen, line 2758: Now I better understand why I've never felt confident starting to use some of these methods.

They feel so "unsupported" when there is no useful docstring. And without types you have to experiment to figure out what goes in and comes out.

.gitignore
```diff
@@ -36,3 +36,4 @@ benchmarks/param
 # pixi
 .pixi
 pixi.lock
+script.*
```
hoxbro, line 39:
```suggestion
```

I don't think personal files like this should be added to the `.gitignore`.

You can add it to a global `.gitignore` on your computer, this is what I do, an example of how to set it up [here](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/).

hoxbro (commented):
Thank you so much! Love it.

param/parameterized.py
```diff
@@ -2260,12 +2294,50 @@ def set_default(self_,param_name,value):
         cls = self_.cls
         setattr(cls,param_name,value)
 
-    def add_parameter(self_, param_name, param_obj):
+    def add_parameter(self_, param_name: str, param_obj: Parameter):
```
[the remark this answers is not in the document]
MarcSkovMadsen: I don't think so. The tests pass?

param/reactive.py
MarcSkovMadsen: `reactive_ops` (reactive operators) and `rx` (reactive expressions) are two different classes and should not have the same docstring. I actually did not understand there was a difference before I started to inspect the code and add docstrings.

MarcSkovMadsen: I added pytests for the 3 docstrings that should be identical. Thx.

param/parameterized.py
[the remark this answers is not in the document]
hoxbro: More to postpone the evaluation of the annotations, https://peps.python.org/pep-0563/. 

My original thinking was it was faster with that enabled, but I haven't tested it.

hoxbro (approved):
Let's see if @philippjfr has anything to say.

.gitignore
MarcSkovMadsen: removed

tests/testreactive.py
```diff
@@ -767,3 +768,7 @@ def test_reactive_callback_resolve_accessor():
     dfx = rx(df)
     out = dfx["name"].str._callback()
     assert out is df["name"].str
+
+def test_docstrings_in_sync():
```
hoxbro, line 772:
```suggestion
def test_docstrings_in_sync():
     # The docstring needs to be explicitly written to work with LSP. 
```

param/parameterized.py
```diff
@@ -2260,12 +2295,50 @@ def set_default(self_,param_name,value):
...
+
+        class P(param.Parameterized):
+            a = param.Number()
+            b = param.String()
+
+
+        p = P()
+        ```
+
+        Add a new parameter to the class via the class:
```
maximlt, line 2327: "to the class via the class"?

param/parameterized.py
```diff
@@ -2309,12 +2382,46 @@ def params(self_, parameter_name=None):
...
+        Updates one or more parameters of this object or class.
+
+        This method allows you to set the parameters of the object or class using a dictionary,
+        an iterable, or keyword arguments in the form of param=value. The specified parameters
+        will be updated to the given values.
+
+        This method can also be used as a context manager to temporarily set and then reset
+        parameter values.
+
+        Reference: https://param.holoviz.org/user_guide/Parameters.html#other-parameterized-methods
```
maximlt, line 2394: Param has an API reference. What do you think about linking to the ref instead of to the user guide?

https://param.holoviz.org/reference/param/generated/param.parameterized.Parameters.update.html

param/parameterized.py
```diff
@@ -2597,7 +2704,45 @@ class or instance that contains an iterable collection of
             for obj in sublist:
                 obj.param.set_dynamic_time_fn(time_fn,sublistattr)
 
-    def serialize_parameters(self_, subset=None, mode='json'):
+    def serialize_parameters(self_, subset=None, mode='json')->str:
```
maximlt, line 2707: How about just typing the return value of `JSONSerialization.serialize_parameters`? At the moment it seems only JSON is supported here so `str` is correct at this level but other formats may be added in the future.

param/parameterized.py
```diff
@@ -2597,7 +2704,45 @@ class or instance that contains an iterable collection of
...
+            A list of parameter names to serialize. If None, all parameters will be serialized. Defaults to None.
+        mode (str, optional):
+            The serialization format. By default, only 'json' is supported. Defaults to 'json'.
+
+        Raises:
+            ValueError: If the `mode` is not valid.
+
+        Returns
+        -------
+        dict: The serialized string
```
maximlt, line 2722: That's not a dict.

param/parameterized.py
```diff
@@ -2309,12 +2382,46 @@ def params(self_, parameter_name=None):
...
+        This method allows you to set the parameters of the object or class using a dictionary,
+        an iterable, or keyword arguments in the form of param=value. The specified parameters
+        will be updated to the given values.
+
+        This method can also be used as a context manager to temporarily set and then reset
+        parameter values.
+
+        Reference: https://param.holoviz.org/user_guide/Parameters.html#other-parameterized-methods
+
+        Examples:
```
maximlt, line 2396: Shouldn't this be formatted as follows?

```suggestion
        Examples
        --------
```

param/parameterized.py
```diff
@@ -1320,6 +1320,41 @@ def schema(self, safe=False, subset=None, mode='json'):
...
     def rx(self):
+        """
+        The reactive namespace.
+
+        Provides reactive versions of the operations that cannot be made reactive through overloading, such as
+        `.rx.and_` and `.rx.bool`. Call it (`()`) to obtain a reactive expression.
+
+        Reference: https://param.holoviz.org/user_guide/Reactive_Expressions.html#special-methods-on-rx
+
+        Examples:
```
maximlt, line 1331:
```suggestion
        Examples
        --------
```

param/parameterized.py
```diff
@@ -2260,12 +2295,50 @@ def set_default(self_,param_name,value):
...
+
+        Parameters
+        ----------
+        param_name : str
+            The name of the parameter to add.
+        param_obj : Parameter
+            The Parameter object to add.
+
+        Example
+        -------
```
maximlt, line 2313:
```suggestion
        Examples
        --------
```

param/parameterized.py
```diff
@@ -2597,7 +2704,45 @@ class or instance that contains an iterable collection of
             for obj in sublist:
                 obj.param.set_dynamic_time_fn(time_fn,sublistattr)
 
-    def serialize_parameters(self_, subset=None, mode='json'):
+    def serialize_parameters(self_, subset=None, mode='json')->str:
+        """Returns the serialized parameters of the Parameterized object.
+
+        Parameters
+        ----------
+        subset (list, optional):
```
maximlt, line 2712: Seems to be that's not how Numpy Docstrings are usually formatted? Based on https://numpydoc.readthedocs.io/en/latest/format.html#parameters.

```suggestion
        subset: list, optional
```

param/parameterized.py
```diff
@@ -2597,7 +2704,45 @@ class or instance that contains an iterable collection of
...
+            A list of parameter names to serialize. If None, all parameters will be serialized. Defaults to None.
+        mode (str, optional):
+            The serialization format. By default, only 'json' is supported. Defaults to 'json'.
+
+        Raises:
+            ValueError: If the `mode` is not valid.
```
maximlt, line 2718: Personal opinion but I usually don't find listing the errors raised super useful.

param/parameterized.py
```diff
@@ -2597,7 +2704,45 @@ class or instance that contains an iterable collection of
...
+        Returns
+        -------
+        dict: The serialized string
+
+        Reference
+        ---------
+        For more details visit https://param.holoviz.org/user_guide/Serialization_and_Persistence.html#serializing-with-json
+
+        Example
+        -------
```
maximlt, line 2729:
```suggestion
        Examples
        --------
```

param/parameterized.py
```diff
@@ -2611,7 +2756,38 @@ def serialize_value(self_, pname, mode='json'):
         serializer = Parameter._serializers[mode]
         return serializer.serialize_parameter_value(self_or_cls, pname)
 
-    def deserialize_parameters(self_, serialization, subset=None, mode='json'):
+    def deserialize_parameters(self_, serialization, subset=None, mode='json') -> dict:
+        """
+        Deserialize the given serialized data. You may use the deserialized to create a Parameteried object or
+        update the parameters of a Parameterized object.
+
+        Parameters:
```
maximlt, line 2764: This docstring is not formatted following the Numpy style.

maximlt (changes requested):
I found inconsistencies in the formatting of the docstrings added in this PR. I also wonder if the links shouldn't point to the reference API instead of to the user guides.

Before merging, I'd recommend building the dev site from this branch and checking that the generated API ref is correctly formatted.

param/parameterized.py
philippjfr: I probably wouldn't call this `Reference:`, rather I'd suggest `Read more:` or something. The user guide is going to be more helpful than the reference guide here though since the API docs will just be generated from this docstring.

philippjfr (changes requested):
This is well overdue, so thanks for doing it. Let's make sure that we consistently use NumPy style docstrings though, right now this is an odd mix.

MarcSkovMadsen:
> This is well overdue, so thanks for doing it. Let's make sure that we consistently use NumPy style docstrings though, right now this is an odd mix.

If think I should investigate whether ruff or something else can validate this. I'm not using Numpy Docstrings very much. So I will fail at this.

MarcSkovMadsen:
Thanks so much for the feedback. Will correct it ASAP.

param/parameterized.py
MarcSkovMadsen: How would you formulate it.

param/parameterized.py
MarcSkovMadsen: I've removed the type annotation.

MarcSkovMadsen:
> This is well overdue, so thanks for doing it. Let's make sure that we consistently use NumPy style docstrings though, right now this is an odd mix.

I've updated the docstrings. Also means the code blocks are now `>>>` prefixed.

param/parameterized.py
```diff
@@ -4206,6 +4378,37 @@ def __init__(self, **params):
...
+        >>> my_instance = MyClass(value=0)
+
+        Access the `value` parameter of `my_instance`:
+
+        >>> my_instance.param.value  # the Parameter instance
+
+        Note that this is different from the current `value` of `my_instance`:
+
+        >>> my_instance.value  # 0, the current parameter value
+        """
```
maximlt, line 4411:
```suggestion
        >>> my_instance.value  # the current parameter value
        0
        """
```

param/parameterized.py
```diff
@@ -1320,6 +1320,43 @@ def schema(self, safe=False, subset=None, mode='json'):
...
     def rx(self):
+        """
+        The reactive namespace.
+
+        Provides reactive versions of operations that cannot be made reactive through operator overloading, such as
+        `.rx.and_` and `.rx.bool`. Calling this namespace (`()`) returns a reactive expression.
+
+        Parameters
+        ----------
+        None
```
maximlt, line 1331: Is this required?

param/parameterized.py
```diff
@@ -1320,6 +1320,43 @@ def schema(self, safe=False, subset=None, mode='json'):
...
+
+        Examples
+        --------
+        Create a Parameterized instance:
+
+        >>> import param
+        >>> class P(param.Parameterized):
+        >>>     a = param.Number()
+        >>>     b = param.String()
+        >>> p = P(a=1, b="hello")
```
maximlt, line 1350: Looks like `b` is not needed for this example

```suggestion
        >>> p = P(a=1)
```

param/parameterized.py
```diff
@@ -2309,13 +2375,47 @@ def params(self_, parameter_name=None):
...
+        This method can also be used as a context manager to temporarily set and then reset parameter values.
+
+        Parameters
+        ----------
+        **params : dict or iterable or keyword arguments
+            The parameters to update, provided as a dictionary, iterable, or keyword arguments in `param=value` format.
+
+        References
+        ----------
+        User Guide: https://param.holoviz.org/user_guide/Parameters.html#other-parameterized-methods
```
maximlt, line 2392:
```suggestion
        User Guide
        ----------
        https://param.holoviz.org/user_guide/Parameters.html#other-parameterized-methods
```

param/parameterized.py
```diff
@@ -1320,6 +1320,43 @@ def schema(self, safe=False, subset=None, mode='json'):
...
+        https://param.holoviz.org/user_guide/Reactive_Expressions.html#special-methods-on-rx
+
+        Examples
+        --------
+        Create a Parameterized instance:
+
+        >>> import param
+        >>> class P(param.Parameterized):
+        >>>     a = param.Number()
+        >>>     b = param.String()
```
maximlt, line 1349:
```suggestion
        ...     a = param.Number()
        ...     b = param.String()
```

param/parameterized.py
```diff
@@ -2260,12 +2297,41 @@ def set_default(self_,param_name,value):
...
+            The Parameter object to add.
+
+        Examples
+        --------
+        Create a Parameterized class:
+
+        >>> import param
+        >>> class P(param.Parameterized):
+        >>>     a = param.Number()
+        >>>     b = param.String()
```
maximlt, line 2321:
```suggestion
        ...     a = param.Number()
        ...     b = param.String()
```

param/parameterized.py
```diff
@@ -2309,13 +2375,47 @@ def params(self_, parameter_name=None):
...
+        User Guide: https://param.holoviz.org/user_guide/Parameters.html#other-parameterized-methods
+
+        Examples
+        --------
+        Create a Parameterized instance:
+
+        >>> import param
+        >>> class P(param.Parameterized):
+        >>>     a = param.Number()
+        >>>     b = param.String()
```
maximlt, line 2401:
```suggestion
        ...     a = param.Number()
        ...     b = param.String()
```

param/parameterized.py
```diff
@@ -2598,6 +2698,41 @@ class or instance that contains an iterable collection of
...
+        https://param.holoviz.org/user_guide/Serialization_and_Persistence.html#serializing-with-json
+
+        Examples
+        --------
+        Create a Parameterized instance and serialize its parameters:
+
+        >>> import param
+        >>> class P(param.Parameterized):
+        >>>     a = param.Number()
+        >>>     b = param.String()
```
maximlt, line 2727:
```suggestion
        ...     a = param.Number()
        ...     b = param.String()
```

param/parameterized.py
```diff
@@ -2260,12 +2297,41 @@ def set_default(self_,param_name,value):
...
+        >>> p = P()
+
+        Add a new parameter to the class via the class:
+
+        >>> P.param.add_parameter('c', param.Tuple(default=(1, 2, 3)))
+        >>> print(p.c)
```
maximlt, line 2327:
```suggestion
        >>> p.c
```

param/parameterized.py
```diff
@@ -2260,12 +2297,41 @@ def set_default(self_,param_name,value):
...
+        (1, 2, 3)
+
+        Add a new parameter to the class via the instance:
+
+        >>> p.param.add_parameter('d', param.Tuple(default=(3, 2, 1)))
+        >>> print(p.d)
```
maximlt, line 2333:
```suggestion
        >>> p.d
```

param/parameterized.py
maximlt: Ok I see what you mean now, I'll make suggestions to make it just a little bit more explicit.

param/parameterized.py
```diff
@@ -2260,12 +2297,41 @@ def set_default(self_,param_name,value):
...
+        --------
+        Create a Parameterized class:
+
+        >>> import param
+        >>> class P(param.Parameterized):
+        >>>     a = param.Number()
+        >>>     b = param.String()
+        >>> p = P()
+
+        Add a new parameter to the class via the class:
```
maximlt, line 2324:
```suggestion
        Add a new parameter to the class `P` via the class namespace `P.param`:
```

param/parameterized.py
```diff
@@ -2260,12 +2297,41 @@ def set_default(self_,param_name,value):
...
+
+        >>> P.param.add_parameter('c', param.Tuple(default=(1, 2, 3)))
+        >>> print(p.c)
+        (1, 2, 3)
+
+        Add a new parameter to the class via the instance:
```
maximlt, line 2330:
```suggestion
        Add a new parameter to the class `P` via the instance namespace `p.param`:
```

maximlt (changes requested):
Thanks for the updates Marc!

I see that Panel also has its examples code doctest formatted (`>>> <code>`). What's your experience using this format? I know that it sometimes annoys me as copy/pasting it (from a website, a docstring tooltip in an IDE, etc.) doesn't always strip `>>>` so you have to manually remove them. IIRC the `ipython` interpreter strips them out but not the base python REPL. The advantage of this format is that it enables running doctest which might be a good idea. By the way, I have no strong opinion over all of that, just trying to gauge the user experience.

MarcSkovMadsen:
I also Think the are hard to copy. But I changed them because of request to Alice with nu not docstrings. I believe that is the format.

MarcSkovMadsen:
It is the format

image

MarcSkovMadsen:
I believe I got the feedback from @maximlt above that `>>>` codestyle is not very helpful as its hard to copy examples. I agree.

So even though `>>>` is numpy docstyle. I plan to overrule this by restructuredText code blocks. Please let me know if you disagree @hoxbro , @philippjfr,  @maximlt as implementing one style and then have to change it is a massive task. Thanks.

maximlt:
> So even though >>> is numpy docstyle. I plan to overrule this by restructuredText code blocks.

Could you show what that format looks like in the code? I'd also like to know how it will end up being displayed:

- in VSCode
- in JupyterLab
- on the reference API of Param's website (you could just make a change to your branch, trigger a dev doc build from Github, and post a screenshot)

MarcSkovMadsen:
> > So even though >>> is numpy docstyle. I plan to overrule this by restructuredText code blocks.
> 
> Could you show what that format looks like in the code? I'd also like to know how it will end up being displayed:
> 
> * in VSCode
> * in JupyterLab
> * on the reference API of Param's website (you could just make a change to your branch, trigger a dev doc build from Github, and post a screenshot)

I'm simply not prepared to do that work again. I tried to analyze situation when started doing this for Panel. And the learning is there is no way that works great. I can focus on VS Code to be honest. The other part is too much.

maximlt:
> I'm simply not prepared to do that work again. I tried to analyze situation when started doing this for Panel. And the learning is there is no way that works great. If you are happy with Panel I'll do it like that. That is what I can offer.

Ok sure I understand, this is all a bit of a mess.

> So even though `>>>` is numpy docstyle. I plan to overrule this by restructuredText code blocks.

Could you show how you plan to change it with an example? For example, how will this docstring look like with the change you plan to make?

```
        """
        Return the serialized parameters of the Parameterized object.

        Parameters
        ----------
        subset : list, optional
            A list of parameter names to serialize. If None, all parameters will be serialized. Defaults to None.
        mode : str, optional
            The serialization format. By default, only 'json' is supported. Defaults to 'json'.

        Returns
        -------
        Any
            The serialized value.

        User Guide
        ----------
        https://param.holoviz.org/user_guide/Serialization_and_Persistence.html#serializing-with-json

        Examples
        --------
        Create a Parameterized instance and serialize its parameters:

        >>> import param
        >>> class P(param.Parameterized):
        >>>     a = param.Number()
        >>>     b = param.String()
        >>> p = P(a=1, b="hello")

        Serialize parameters:

        >>> serialized_data = p.param.serialize_parameters()
        >>> print(serialized_data)
        {"name": "P00002", "a": 1, "b": "hello"}
        """
```

Then I can look a little bit what would the user experience be on various platforms (VSCode, JLab, website, etc.).

---

Just linking something that I wasn't aware of, the `sphinx-copybutton` Sphinx extension has lots of ways to configure what gets copied in the end, possibly stripping out `>>> `: https://sphinx-copybutton.readthedocs.io/en/latest/use.html. I'm not sure why Pandas doesn't use that (check https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sort_values.html).

param/parameterized.py
```diff
@@ -2309,13 +2375,47 @@ def params(self_, parameter_name=None):
...
         """
-        For the given dictionary or iterable or set of param=value
-        keyword arguments, sets the corresponding parameter of this
-        object or class to the given value.
+        Update one or more parameters of this object or class.
+
+        Allows setting the parameters of the object or class using a dictionary, an iterable, or keyword arguments
+        in the form of `param=value`. The specified parameters will be updated to the given values.
+
+        This method can also be used as a context manager to temporarily set and then reset parameter values.
```
jbednar, line 2383: Does `update` let you set a bunch of parameter values at once without triggering events until all of them have been set? I may be confused about some other method, but if this is what lets you do that, it should presumably be mentioned in the docstring.

jbednar (approved):
This looks very useful; thanks @MarcSkovMadsen !

param/parameterized.py
maximlt: Yes, usually we recommend updating multiple parameters at once with `update` instead of sequentially, specially in a Panel app where this can lead to multiple re-renders.

jbednar: I'm not sure how best to express that in terms that make sense at the Param level, and didn't see much in the code that would let me state precisely what is deferred during the update.  @philippjfr, can we say that events are not delivered to watchers until all parameter values are updated?

param/parameterized.py
MarcSkovMadsen: I've improved it even further in my code. Thx for the suggestion.

param/parameterized.py
MarcSkovMadsen: fixed in my version.

param/parameterized.py
MarcSkovMadsen: fixed in my version. Thx.

param/parameterized.py
MarcSkovMadsen: Fixed in my version. Thx.

param/parameterized.py
MarcSkovMadsen: Fixed in my version. Thx.

param/parameterized.py
MarcSkovMadsen: added in my version. Thx a lot.

param/parameterized.py
MarcSkovMadsen: Added in my version. Great suggestion. Thx.

param/parameterized.py
MarcSkovMadsen: > Yes, usually we recommend updating multiple parameters at once with `update` instead of sequentially, specially in a Panel app where this can lead to multiple re-renders.

I'm always in doubt about `.update`. Does it batch updates. I've never seen that explained or demonstrated.

MarcSkovMadsen:
> > I'm simply not prepared to do that work again. I tried to analyze situation when started doing this for Panel. And the learning is there is no way that works great. If you are happy with Panel I'll do it like that. That is what I can offer.
> 
> Ok sure I understand, this is all a bit of a mess.
> 
> > So even though `>>>` is numpy docstyle. I plan to overrule this by restructuredText code blocks.
> 
> Could you show how you plan to change it with an example? For example, how will this docstring look like with the change you plan to make?
> 
> ```
>         """
>         Return the serialized parameters of the Parameterized object.
> 
>         Parameters
>         ----------
>         subset : list, optional
>             A list of parameter names to serialize. If None, all parameters will be serialized. Defaults to None.
>         mode : str, optional
>             The serialization format. By default, only 'json' is supported. Defaults to 'json'.
> 
>         Returns
>         -------
>         Any
>             The serialized value.
> 
>         User Guide
>         ----------
>         https://param.holoviz.org/user_guide/Serialization_and_Persistence.html#serializing-with-json
> 
>         Examples
>         --------
>         Create a Parameterized instance and serialize its parameters:
> 
>         >>> import param
>         >>> class P(param.Parameterized):
>         >>>     a = param.Number()
>         >>>     b = param.String()
>         >>> p = P(a=1, b="hello")
> 
>         Serialize parameters:
> 
>         >>> serialized_data = p.param.serialize_parameters()
>         >>> print(serialized_data)
>         {"name": "P00002", "a": 1, "b": "hello"}
>         """
> ```
> 
> Then I can look a little bit what would the user experience be on various platforms (VSCode, JLab, website, etc.).
> 
> Just linking something that I wasn't aware of, the `sphinx-copybutton` Sphinx extension has lots of ways to configure what gets copied in the end, possibly stripping out `>>> `: https://sphinx-copybutton.readthedocs.io/en/latest/use.html. I'm not sure why Pandas doesn't use that (check https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sort_values.html).

The below is how it looks in my version of VS Code on windows with pylance. I don't have the latest version of VS Code.

https://github.com/user-attachments/assets/170a9117-ccbc-43ad-8ed9-950b4fac899e

param/parameterized.py
MarcSkovMadsen: I've removed it.

param/parameterized.py
MarcSkovMadsen: I does update everything and then execute callbacks:

```python
import param

class MyParam(param.Parameterized):
    a = param.String()
    b = param.String()

my_param = MyParam(a="0. Hello", b="0. World")

@param.depends(my_param.param.a, watch=True)
def print_b(a):
    print(my_param.b)

my_param.a="1. Hello"
my_param.b="1. World"
my_param.param.update(a="2. Hello",b="2. World")
```

```bash
0. World
2. World
```

MarcSkovMadsen: I*ve updated the docstring based on the feedback.

maximlt (changes requested):
LGTM other than 1 pending suggestion. Let's preserve the original Numpy style for the examples, I think we can tweak the sphinx-copybutton extension to work in a decent way, and hopefully VSCode will make it easier to copy/paste these snippets. If we end up needing to change the style, it shouldn't be too hard to write a script to re-write the docstrings automatically, stripping `>>>` and `...` (the other way around, yes!).

PR #978, opened 2024-11-13
refactor: Change linting to Ruff

hoxbro:
Aligning with our other repos.


param/parameterized.py
```diff
@@ -340,7 +340,7 @@ def edit_constant(parameterized):
         p.constant = False
     try:
         yield
-    except:
+    except Exception:
         raise
```
hoxbro, line 344: Likely this can just be removed or will be removed when running something like pyupgrade

tests/testdateparam.py
```diff
@@ -11,7 +11,7 @@
 
 try:
     import numpy as np
-except:
+except ImportError:
```
maximlt, line 14: What do you think about catching `ModuleNotFoundError` instead?

hoxbro: I have update every ImportError in ccaf158 to ModuleNotFoundError

param/parameterized.py
```diff
@@ -26,7 +26,7 @@
 # Allow this file to be used standalone if desired, albeit without JSON serialization
 try:
     from . import serializer
-except ImportError:
+except ModuleNotFoundError:
```
hoxbro, line 29: I'm not entirely sure about this one.

param/version.py
```diff
@@ -517,7 +519,7 @@ def get_setupcfg_version():
     """
     try:
         import configparser
-    except ImportError:
+    except ModuleNotFoundError:
         import ConfigParser as configparser # python2 (also prevents dict-like access)
```
hoxbro, line 523: We can likely remove this one...

param/parameterized.py
hoxbro: Tested it. Needs to be an `ImportError`

```suggestion
except ImportError:
```

hoxbro: Though even with ImportError it still does not work...

``` python-traceback
❯ python -c "import param"  # with ModuleNotFoundError
  File "<string>", line 1, in <module>
  File "/home/shh/projects/holoviz/repos/param/param/__init__.py", line 4, in <module>
    from .depends import depends  # noqa: api import
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/shh/projects/holoviz/repos/param/param/depends.py", line 9, in <module>
    from .parameterized import (
  File "/home/shh/projects/holoviz/repos/param/param/parameterized.py", line 28, in <module>
    from . import serializer
ImportError: cannot import name 'serializer' from partially initialized module 'param' (most likely due to a circular import) (/home/shh/projects/holoviz/repos/param/param/__init__.py)

param on  refactor_ruff [$✘] via 🅒 holoviz
❯ python -c "import param"  # with ImportError
  File "<string>", line 1, in <module>
  File "/home/shh/projects/holoviz/repos/param/param/__init__.py", line 4, in <module>
    from .depends import depends  # noqa: api import
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/shh/projects/holoviz/repos/param/param/depends.py", line 9, in <module>
    from .parameterized import (
  File "/home/shh/projects/holoviz/repos/param/param/parameterized.py", line 1027, in <module>
    class Parameter(_ParameterBase):
  File "/home/shh/projects/holoviz/repos/param/param/parameterized.py", line 1170, in Parameter
    _serializers = {'json': serializer.JSONSerialization}
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'JSONSerialization'
```

hoxbro: Should we remove the try/except here? Nobody has complained about it, and that line was changed in June of 202.

maximlt: From what I've heard, and IIRC, Param was designed to be used super easily by simply copy-pasting one or two files in a project, which works as it has no external dependency. It's likely it made people's lives easier 10/15 years ago, these days we expect users to `pip/conda install` their dependencies and in the vast majority of cases it works fine.
So I think that yes we should remove the `try/except` and generally lift the assumption that you can copy-paste some modules and it works, if it's baked in other places. I also wouldn't be surprised if one day we added some dependencies to Param (e.g. `typing-extensions` if we finally do some typing work in Param).

cc @philippjfr @jbednar @jlstevens

jlstevens: @maximlt is right in that param was originally designed to be easily vendored-in to any codebase. I personally have no strong feelings about this: while being easy to avoid an extra dependencies can be nice, it is also true that the majority of users will be using a package manager.

hoxbro: To be sure, the question right now is not about adding an extra dependency but about removing the `try/except` around the serializer import, which has been broken since 2021.

jlstevens: Right, was just confirming what Maxime said and given this import has been broken so long, the issue is rather moot!

maximlt (approved):

param/parameterized.py
jbednar: I agree that it is no longer important to be able to paste in the two main Param files directly into a project. Param was first written before conda and even before pip, and adding a dependency was much more of a big issue in those days. Happy to see this and any similar try/excepts removed, when they depend only on files in param itself.

PR #980, opened 2024-11-22
refactor: Remove `_dict_update`

hoxbro:
I think this function does exactly what a normal `dict` does if the first argument is a dict.

### DIFF
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -203,15 +203,6 @@ def _is_mutable_container(value):
     return issubclass(type(value), MUTABLE_TYPES)
 
 
-def _dict_update(dictionary, **kwargs):
-    """
-    Small utility to update a copy of a dict with the provided keyword args.
-    """
-    d = dictionary.copy()
-    d.update(kwargs)
-    return d
-
-
 def full_groupby(l, key=lambda x: x):
     """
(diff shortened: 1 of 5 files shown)


param/_utils.py
```diff
@@ -203,15 +203,6 @@ def _is_mutable_container(value):
     return issubclass(type(value), MUTABLE_TYPES)
```
hoxbro, line 203: Unrelated but this looks more correct to me
```suggestion
    return isinstance(value, MUTABLE_TYPES)
```

param/_utils.py
```diff
@@ -203,15 +203,6 @@ def _is_mutable_container(value):
...
 
 
-def _dict_update(dictionary, **kwargs):
```
hoxbro, line 206: I haven't checked if any downstream uses this.

maximlt (approved):

jbednar (approved):
Seems fine to me, though I do always worry if I can't remember why it was done this way originally. Maybe some ancient issue with an earlier Python version?

hoxbro:
> Seems fine to me, though I do always worry if I can't remember why it was done this way originally. Maybe some ancient issue with an earlier Python version?

Either that, it was more complex at some point, or just forgetting that this is possible.

PR #982, opened 2024-11-23
Add Ruff D rule

MarcSkovMadsen:
Builds on #977. So review that one first.

The only thing I've touched is in pyproject.toml. The rest is auto fixed by Ruff when I run `ruff check`.

### DIFF
--- a/param/serializer.py
+++ b/param/serializer.py
@@ -13,7 +13,7 @@ class UnsafeserializableException(Exception):
     pass
 
 def JSONNullable(json_type):
-    "Express a JSON schema type as nullable to easily support Parameters that allow_None"
+    """Express a JSON schema type as nullable to easily support Parameters that allow_None"""
     return {'anyOf': [ json_type, {'type': 'null'}] }
 
 
@@ -25,7 +25,7 @@ class Serialization:
 
     @classmethod
     def schema(cls, pobj, subset=None):
-        raise NotImplementedError        # noqa: unimplemented method
+        raise NotImplementedError
 
     @classmethod
     def serialize_parameters(cls, pobj, subset=None):
@@ -33,7 +33,7 @@ class Serialization:
         Serialize the parameters on a Parameterized object into a
         single serialized object, e.g. a JSON string.
         """
-        raise NotImplementedError        # noqa: unimplemented method
+        raise NotImplementedError
 
     @classmethod
     def deserialize_parameters(cls, pobj, serialized, subset=None):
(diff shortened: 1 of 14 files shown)


pyproject.toml
```diff
@@ -119,8 +119,23 @@ fix = true
...
 ignore = [
+    "D100", # Missing docstring in public module"
+    "D101", # Missing docstring in public class"
+    "D102", # Missing docstring in public method
+    "D103", # Missing docstring in public function
+    "D104", # Missing docstring in public package"
+    "D105", # Missing docstring in magic method"
+    "D107", # Missing docstring in `__init__`"
+    "D200", # One-line docstring should fit on one line"
+    "D203", # one-blank-line-before-class and `no-blank-line-before-class` (D211) are incompatible.
```
MarcSkovMadsen, line 132: I'm proposing D211 over D203. What I care most about is that we have a rule enforcing consistency.

pyproject.toml
```diff
@@ -119,8 +119,23 @@ fix = true
...
+    "D205", # 1 blank line required between summary line and description
+    "D212", # multi-line-summary-first-line. Alternative is to ignore D213 `multi-line-summary-second-line`.
```
MarcSkovMadsen, line 134: I'm proposing D213 over D212. Again - I don't care so much about the specific rule as I care about having the rule and consistency.

hoxbro:
Just changing base branch so it easier to review

param/serializer.py
```diff
@@ -25,37 +25,37 @@ class Serialization:
 
     @classmethod
     def schema(cls, pobj, subset=None):
-        raise NotImplementedError        # noqa: unimplemented method
+        raise NotImplementedError
```
hoxbro, line 28: I'm unsure why this has changed; maybe it's because of  `--unsafe-fixes`. I don't think they should be added back

tests/utils.py
```diff
@@ -138,6 +139,7 @@ async def async_wait_until(fn, timeout=5000, interval=100):
         Waiting interval, by default 100
 
     Adapted from pytest-qt.
+
```
hoxbro, line 142: I don't understand this empty line, as other docstrings do not necessarily have it. But if this is what ruff say, so be it.

hoxbro (approved):

hoxbro (approved):

param/serializer.py
MarcSkovMadsen: Ruff complained that specific rules should be used with noqa. So I removed them. Forgot to explain that.

maximlt (approved):

pyproject.toml
maximlt: Removed the line

PR #983, opened 2024-11-23
Enable Ruff d200 rule

MarcSkovMadsen:
Continues from #982. So review that one first.

I've re-renabled rule `D200` in pyproject.toml and applied `ruff check --unsafe-fixes`.

Then checked the changes manually.

### DIFF
--- a/numbergen/__init__.py
+++ b/numbergen/__init__.py
@@ -1,6 +1,4 @@
-"""
-Callable objects that generate numbers according to different distributions.
-"""
+"""Callable objects that generate numbers according to different distributions."""
 
 import random
 import operator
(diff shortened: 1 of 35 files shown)


hoxbro (approved):

maximlt:
@MarcSkovMadsen in case it wasn't clear, I did not suggest to open one PR per rule:

> I'm not necessarily saying one rule at a time. It's just I'd like to avoid having to review all the docstrings of the complete API at once, as it's a lot, and I'm quite sure there are a few traps along the way that could slow you down or even block you entirely if done in one PR, while if you iterate we can be much more efficient and release improvements quickly.

I think it's totally fine if you open one PR that enables multiple rules at the same time, specially if it's only about formatting. Things will get more challenging for reviewers as soon as you'll touch the real meat which is the content of the docstrings.

cc @hoxbro

maximlt (approved):

MarcSkovMadsen:
Are you expecting me to squash and merge?

hoxbro:
It was a bit hard to rebase so maybe take a look that everything is ok and _then_ merge it.

PR #992, opened 2024-12-25
Enable ruff rules

MarcSkovMadsen:
This PR Enables the docstring rules:

- [D104](https://docs.astral.sh/ruff/rules/undocumented-public-package/): Checks for undocumented public package definitions.
- [D107](https://docs.astral.sh/ruff/rules/undocumented-public-init/): Checks for public __init__ method definitions that are missing docstrings.
- [D400](https://docs.astral.sh/ruff/rules/ends-in-period/): First line should end with a period.
- [D401](https://docs.astral.sh/ruff/rules/non-imperative-mood/): First line of docstring should be in imperative mood.

This is just a step on the way to improving the content of the docstrings - especially for `.rx` and `.param` methods. But first I would like Ruff docstring support to be able to help me.

### DIFF
--- a/param/version.py
+++ b/param/version.py
@@ -134,22 +134,22 @@ class Version:
 
     @property
     def release(self):
-        """Return the release tuple"""
+        """Return the release tuple."""
         return self.fetch()._release
 
     @property
     def commit(self):
-        """A specification for this particular VCS version, e.g. a short git SHA"""
+        """A specification for this particular VCS version, e.g. a short git SHA."""
         return self.fetch()._commit
 
     @property
     def commit_count(self):
-        """Return the number of commits since the last release"""
+        """Return the number of commits since the last release."""
         return self.fetch()._commit_count
 
     @property
     def dirty(self):
-        """True if there are uncommited changes, False otherwise"""
+        """True if there are uncommited changes, False otherwise."""
         return self.fetch()._dirty
 
 
@@ -259,7 +259,7 @@ class Version:
 
 
     def _update_from_vcs(self, output):
-        """Update state based on the VCS state e.g the output of git describe"""
+        """Update state based on the VCS state e.g the output of git describe."""
         split = output[1:].split('-')
         dot_split = split[0].split('.')
         for prefix in ['a','b','rc']:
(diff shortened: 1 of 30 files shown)


param/version.py
maximlt, line 1: Please revert the changes to this module and configure `ruff` to ignore it. There's no need to touch it, in fact we should deprecate it.

maximlt (commented):
@MarcSkovMadsen why is there a mix of Google and Numpy styles?

```
    def __init__(self,lhs,rhs,operator,reverse=False,**args):
        """Initialize a BinaryOperator with operands, an operator, and optional arguments.

        Args:
            lhs: The left-hand side operand, which can be a NumberGenerator or a number.
            rhs: The right-hand side operand, which can be a NumberGenerator or a number.
            operator (Callable): The binary operator to apply to the operands.
            reverse (bool, optional): If `True`, swaps the left and right operands. Defaults to `False`.
            **args: Optional keyword arguments to pass to the operator when it is called.

        Notes
        -----
            It is currently not possible to set parameters in the superclass during
            initialization because `**args` is used by this class itself.
```

For this PR, and the others you have in preparation, I'm curious and would like to know more about your process. Do you sometimes use an LLM? Or just look at the current docstrings + doc + code ?

param/version.py
philippjfr: Agreed, let's ignore and deprecate asap.

MarcSkovMadsen:
Its a mistake. I'm so used to Google docstring style.

For this PR I dont remember using a llm a lit as I expected simple to fix issues.

For the other PRs I used an llm. I started by explaining I'm updating docstring of HoloViz Param and which used numpy docstring style. I then copy (parts of) the class or function to give it context and then ask to Update the docstring. I then copy the proposed docstring back and start reviewing it. Often i remove repeated text, add a Documentation section or change the examples. I've run all the examples in Ipython. Sometimes I copy the code back to the llm and ask it to improve it.

param/version.py
philippjfr: Happy to do this myself btw. Appreciate your efforts here.

philippjfr: Hmm, I'll leave the changes it made but did exclude it. The changes are minor and at least consistent.

MarcSkovMadsen:
Thanks for fixing the Google Docstring style. As far as I can see the failing macos tests is due to some import change by Philipp.

Let me know if there is more I should do here. Thanks.

philippjfr:
>As far as I can see the failing macos tests is due to some import change by Philipp.

It's not, the conda-forge build chain broke pyarrow via libprotobuff and libabseil. Have to wait until they fix it.

PR #994, opened 2024-12-26
Improve reactive docstrings

MarcSkovMadsen:
This PR continues from https://github.com/holoviz/param/pull/992. Please review that one first.

The aim of this PR is to make it easier to use reactive expressions by

- Improving the docstrings
- Selectively adding type annotations.

### DIFF
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -338,7 +338,6 @@ class reactive_ops:
         >>> rx_buffer.rx.value
         [2, 3, 4]
         """
-
         items = []
         def collect(new, n):
             items.append(new)
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -1,83 +1,85 @@
 """
-reactive API.
+Reactive API for Dynamic Expression Pipelines.
 
-`rx` is a wrapper around a Python object that lets users create
-reactive expression pipelines by calling existing APIs on an object with dynamic
-parameters or widgets.
+`rx` provides a wrapper around Python objects, enabling the creation of
+reactive expression pipelines that dynamically update based on changes to their
+underlying parameters or widgets. This approach simplifies the implementation of
+dynamic, interactive behavior in Python code, particularly for data manipulation
+and visualization tasks.
 
-An `rx` instance watches what operations are applied to the object
-and records these on each instance, which are then strung together
-into a chain.
+## Overview
 
-The original input to an `rx` object is stored in a mutable list and can be
-accessed via the `_obj` property. The shared mutable data structure
-ensures that all `rx` instances created from the same object can
-hold a shared reference that can be updated, e.g. via the `.value`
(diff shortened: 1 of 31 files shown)


param/reactive.py
```diff
@@ -212,21 +228,116 @@ def __init__(self, reactive):
...
+
+        Examples
+        --------
+        Create two reactive expressions and combine them using `and_`:
+
+        >>> import param
+        >>> a = param.rx(True)
+        >>> b = param.rx(False)
+        >>> result = a.rx.and_(b)
+        >>> result.rx.value
```
MarcSkovMadsen, line 262: In ipython or notebook the output of `result.rx.value` would look the same as the output of the reactive expression `result`. Should I write `result` instead because its shorter and what should be used in those environments? Or should I not because it would not be useful in the basic Pyton REPL?

param/reactive.py
```diff
@@ -1004,7 +1705,7 @@ def evaluate(*args, **kwargs):
             return bind(evaluate, *params)
         return evaluate
 
-    def _clone(self, operation=None, copy=False, **kwargs):
+    def _clone(self, operation=None, copy=False, **kwargs) -> 'rx':
```
MarcSkovMadsen, line 1708: FYI. The `rx' type annotation here made a huge difference in VS Code interpretability of the code.

tests/testreactive.py
```diff
@@ -768,8 +768,3 @@ def test_reactive_callback_resolve_accessor():
     dfx = rx(df)
     out = dfx["name"].str._callback()
     assert out is df["name"].str
-
-def test_docstrings_in_sync():
```
MarcSkovMadsen, line 772: I removed this test because I do believe the docstrings in fact should be different:

Users our `Parameter.rx` would expect a docstring with examples focused on Parameter usage. Users of `rx.rx` property would expect examples focused on `rx` usage.

param/reactive.py
```diff
@@ -462,33 +1108,69 @@ def cb(value):
...
+    to any bound parameters or reactive expressions, ensuring that its output
+    remains up-to-date.
+
+    Similar to `functools.partial`, arguments can also be bound to constants,
+    leaving a simple callable object. When `watch=True`, the function is
+    automatically evaluated whenever any bound parameter or reactive expression changes.
+
+    Parameters
+    ----------
+    function : callable, generator, async generator, or coroutine
```
MarcSkovMadsen, line 1125: I really tried to figure out what *function* arguments are supported. But its difficult.

philippjfr: I think you covered it.

param/reactive.py
philippjfr: I'd stick with it as is.

philippjfr (approved):

PR #997, opened 2024-12-28
Enhance documentation of param namespace

MarcSkovMadsen:
Continues from https://github.com/holoviz/param/pull/992. Please review and merge that one first.

Focus is on updating the docstrings of `Parameters` class and its methods, i.e. the `.param` namespace.
I've added type annotations too.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -3200,7 +3200,7 @@ class Parameters:
         else:
             return param_obj._force(slf, cls)
 
-    def get_value_generator(self_,name: str)->Any: # pylint: disable-msg=E0213
+    def get_value_generator(self_,name: str) -> Any: # pylint: disable-msg=E0213
         """
         Retrieve the value or value-generating object of a named parameter.
 
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -3575,8 +3575,13 @@ class Parameters:
                 getattr(watchers[what], action)(watcher)
 
     def watch(
-        self_, fn, parameter_names: list[str], what: str='value', onlychanged: bool=True,
-        queued: bool=False, precedence: int=0
+        self_,
+        fn,
+        parameter_names: Union[str, list[str]],
+        what: str='value',
+        onlychanged: bool = True,
+        queued: bool = False,
+        precedence: int = 0,
     ) -> Watcher:
         """
         Register a callback function to be invoked for parameter events.
@@ -3591,8 +3596,8 @@ class Parameters:
             The callback function to invoke when an event occurs. This function
             will be provided with `Event` objects as positional arguments, allowing
             it to determine the triggering events.
-        parameter_names : list[str]
-            A list of parameter names to watch for events.
+        parameter_names : str or list[str]
+            A parameter name or a list of parameter names to watch for events.
         what : str, optional
             The type of change to watch for. By default, this is 'value', but it
             can be set to other slots such as 'constant'. Default is 'value'.
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2631,8 +2631,8 @@ class Parameters:
             - `True`: Return instance-specific parameters, creating them if necessary. This
             requires the instance to be fully initialized.
             - `False`: Return class-level parameters without creating instance-specific copies.
-            - `'existing'`: Return only the instance parameters that already exist, avoiding
-            creation of new instance-specific parameters.
+            - `'existing'`: Returns a mix of instance parameters that already exist and
+            class parameters, avoiding creation of new instance-specific parameters.
 
         Returns
         -------
@@ -2645,14 +2645,6 @@ class Parameters:
             If the method is called on a `Parameterized` instance that has not been
             fully initialized. Ensure `super().__init__(**params)` is called in the
             constructor before triggering watchers.
-
-        Notes
-        -----
-        - This method distinguishes between class-level and instance-specific parameters.
-        - Instance-specific parameters are lazily created; they are not initialized unless
-        explicitly requested or accessed.
-        - When `instance='existing'`, only parameters already initialized at the instance level
-        will be returned, while class-level parameters remain unaffected.
 
         Examples
         --------
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2672,9 +2672,12 @@ class Parameters:
         """
         if self_.self is not None and not self_.self._param__private.initialized and instance is True:
             raise RuntimeError(
-                'Cannot access instance parameters before the Parameterized instance '
-                'is fully initialized. Ensure `super().__init__(**params)` is called, or '
-                'use `.param.objects(instance=False)` for class parameters.'
+                'Looking up instance Parameter objects (`.param.objects()`) until '
+                'the Parameterized instance has been fully initialized is not allowed. '
+                'Ensure you have called `super().__init__(**params)` in your Parameterized '
+                'constructor before trying to access instance Parameter objects, or '
+                'looking up the class Parameter objects with `.param.objects(instance=False)` '
+                'may be enough for your use case.',
             )
 
         pdict = self_._cls_parameters
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -3281,8 +3281,7 @@ class Parameters:
         -------
         Any
             The current value of the parameter, the last generated value for
-            `Dynamic` parameters, or a list of inspected values for composite
-            parameters.
+            `Dynamic` parameters.
 
         Examples
         --------
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -4097,7 +4097,6 @@ class Parameters:
 
         Notes
         -----
-        - This method is experimental and subject to change in future versions.
         - The generated representation assumes the necessary imports are provided
         for evaluation with `eval`.
 
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -3627,12 +3627,6 @@ class Parameters:
         Watcher
             The `Watcher` object that encapsulates the registered callback.
 
-        Raises
-        ------
-        ValueError
-            If a negative precedence is provided, which is reserved for internal
-            watchers.
-
         See Also
         --------
         Watcher : Contains detailed information about the watcher object.
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -29,7 +29,7 @@ from html import escape
 from itertools import chain
 from operator import itemgetter, attrgetter
 from types import FunctionType, MethodType
-from typing import Type, Union, Literal # When python 3.9 support is dropped replace Union with |
+from typing import Any, Type, Union, Literal, Iterable, Callable # When python 3.9 support is dropped replace Union with |
 
 from contextlib import contextmanager
 from logging import DEBUG, INFO, WARNING, ERROR, CRITICAL
@@ -273,9 +273,9 @@ def _batch_call_watchers(parameterized, enable=True, run=True):
 
     Args:
         parameterized: The object whose watchers are being managed.
-        enable (bool, optional): If `True`, enable batching of events. Defaults to `True`.
+        enable (bool, optional): If `True`, enable batching of events. Default is `True`.
         run (bool, optional): If `True`, execute accumulated watchers on exit.
-                              If `False`, they remain queued. Defaults to `True`.
+                              If `False`, they remain queued. Default is `True`.
 
     Yields
     ------
@@ -1950,16 +1950,23 @@ class Parameters:
         for k, v in state.items():
             setattr(self, k, v)
 
-    def __getitem__(self_, key):
-        """Retrieve the class or instance parameter by key.
+    def __getitem__(self_, key: str) -> Parameter:
+        """
+        Retrieve a Parameter by its key.
 
-        Args:
-            key: The name of the parameter to retrieve.
+        This method allows access to a class or instance Parameter using its name.
+        If accessed on an instance, the returned Parameter will be instantiated.
+
+        Parameters
+        ----------
+        key : str
+            The name of the Parameter to retrieve.
 
         Returns
         -------
-            The parameter associated with the given key. If accessed on an instance,
-            returns the instantiated parameter.
+        Parameter
+            The Parameter associated with the given key. If accessed on an instance,
+            the method returns the instantiated parameter.
         """
         inst = self_.self
         if inst is None:
@@ -2797,7 +2804,10 @@ class Parameters:
                           if (name, watcher.what) in event_dict]
                 with _batch_call_watchers(self_.self_or_cls, enable=watcher.queued, run=False):
                     self_._execute_watcher(watcher, events)
-
+    # Please update the docstring with better description and examples
+    # I've (MarcSkovMadsen) not been able to understand this. Its probably because I lack context.
+    # Its not mentioned in the documentation.
+    # The pytests do not make sense to me.
     def set_dynamic_time_fn(self_,time_fn,sublistattr=None):
         """
         Set time_fn for all Dynamic Parameters of this class or
@@ -2839,24 +2849,30 @@ class Parameters:
             for obj in sublist:
                 obj.param.set_dynamic_time_fn(time_fn,sublistattr)
 
-    def serialize_parameters(self_, subset=None, mode='json'):
+    def serialize_parameters(self_, subset: Union[Iterable[str], None]=None, mode='json'):
         """
         Return the serialized parameters of the Parameterized object.
 
         Parameters
         ----------
-        subset : list, optional
-            A list of parameter names to serialize. If None, all parameters will be serialized. Defaults to None.
+        subset : iterable, optional
+            An iterable of parameter names to serialize. If None, all parameters will be serialized.
+            Default is None.
         mode : str, optional
-            The serialization format. By default, only 'json' is supported. Defaults to 'json'.
+            The serialization format. By default, only 'json' is supported. Default is 'json'.
 
         Returns
         -------
         Any
             The serialized value.
 
-        User Guide
-        ----------
+        Raises
+        ------
+        ValueError
+            If the specified serialization mode is not supported.
+
+        Documentation
+        -------------
         https://param.holoviz.org/user_guide/Serialization_and_Persistence.html#serializing-with-json
 
         Examples
@@ -2872,24 +2888,67 @@ class Parameters:
         Serialize parameters:
 
         >>> serialized_data = p.param.serialize_parameters()
-        >>> print(serialized_data)
-        {"name": "P00002", "a": 1, "b": "hello"}
-
+        >>> serialized_data
+        '{"name": "P00002", "a": 1, "b": "hello"}'
         """
-        self_or_cls = self_.self_or_cls
         if mode not in Parameter._serializers:
             raise ValueError(f'Mode {mode!r} not in available serialization formats {list(Parameter._serializers.keys())!r}')
+        self_or_cls = self_.self_or_cls
         serializer = Parameter._serializers[mode]
         return serializer.serialize_parameters(self_or_cls, subset=subset)
 
-    def serialize_value(self_, pname, mode='json'):
-        self_or_cls = self_.self_or_cls
+    def serialize_value(self_, pname: str, mode: str='json'):
+        """
+        Serialize the value of a specific parameter.
+
+        This method serializes the value of a given parameter on a Parameterized
+        object using the specified serialization mode.
+
+        Parameters
+        ----------
+        pname : str
+            The name of the parameter whose value is to be serialized.
+        mode : str, optional
+            The serialization format to use. By default, only 'json' is supported.
+            Default is 'json'.
+
+        Returns
+        -------
+        Any
+            The serialized value of the specified parameter.
+
+        Raises
+        ------
+        ValueError
+            If the specified serialization mode is not supported.
+
+        Documentation
+        -------------
+        https://param.holoviz.org/user_guide/Serialization_and_Persistence.html#serializing-with-json
+
+        Examples
+        --------
+        Serialize the value of a specific parameter:
+
+        >>> import param
+        >>> class P(param.Parameterized):
+        ...     a = param.Number()
+        ...     b = param.String()
+        >>> p = P(a=1, b="hello")
+
+        Serialize the value of parameter 'a':
+
+        >>> serialized_value = p.param.serialize_value('a')
+        >>> serialized_value
+        '1'
+        """
         if mode not in Parameter._serializers:
             raise ValueError(f'Mode {mode!r} not in available serialization formats {list(Parameter._serializers.keys())!r}')
+        self_or_cls = self_.self_or_cls
         serializer = Parameter._serializers[mode]
         return serializer.serialize_parameter_value(self_or_cls, pname)
 
-    def deserialize_parameters(self_, serialization, subset=None, mode='json') -> dict:
+    def deserialize_parameters(self_, serialization, subset: Union[Iterable[str], None]=None, mode: str='json') -> dict:
         """
         Deserialize the given serialized data. This data can be used to create a
         `Parameterized` object or update the parameters of an existing `Parameterized` object.
@@ -2897,21 +2956,26 @@ class Parameters:
         Parameters
         ----------
         serialization : str
-            The serialized parameter data as a JSON string.
-        subset : list of str, optional
-            A list of parameter names to deserialize. If `None`, all parameters will be
-            deserialized. Defaults to `None`.
+            The serialized parameter data.
+        subset : iterable of str, optional
+            An iterable of parameter names to deserialize. If `None`, all parameters will be
+            deserialized. Default is `None`.
         mode : str, optional
             The serialization format. By default, only 'json' is supported.
-            Defaults to 'json'.
+            Default is 'json'.
 
         Returns
         -------
         dict
             A dictionary with parameter names as keys and deserialized values.
 
-        User Guide
-        ----------
+        Raises
+        ------
+        ValueError
+            If the specified serialization mode is not supported.
+
+        Documentation
+        -------------
         https://param.holoviz.org/user_guide/Serialization_and_Persistence.html#serializing-with-json
 
         Examples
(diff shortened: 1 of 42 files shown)


param/parameterized.py
```diff
@@ -2622,7 +2804,10 @@ def _batch_call_watchers(self_):
                           if (name, watcher.what) in event_dict]
                 with _batch_call_watchers(self_.self_or_cls, enable=watcher.queued, run=False):
                     self_._execute_watcher(watcher, events)
-
+    # Please update the docstring with better description and examples
+    # I've (MarcSkovMadsen) not been able to understand this. Its probably because I lack context.
+    # Its not mentioned in the documentation.
+    # The pytests do not make sense to me.
     def set_dynamic_time_fn(self_,time_fn,sublistattr=None):
```
MarcSkovMadsen, line 2811: I hope someone else than me will update the docstring. I've not been able to understand this method. Its probably because I lack context. Its not mentioned in the documentation. The pytests are not valuable to me.

param/parameterized.py
```diff
@@ -2815,6 +3178,9 @@ def values(self_, onlychanged=False):
         vals.sort(key=itemgetter(0))
         return dict(vals)
 
+    # Please update the docstring with better description and examples
+    # I've (MarcSkovMadsen) not been able to understand this. Its probably because I lack context.
+    # Its not mentioned in the documentation or pytests
     def force_new_dynamic_value(self_, name): # pylint: disable-msg=E0213
```
MarcSkovMadsen, line 3184: I hope someone else than me will update this docstring. I lack context and cannot find it in the documentation or pytests.

maximlt:
> I don't know why this fails on windows + python 3.12

These are unrelated test failures.

@MarcSkovMadsen can you resolve the conflicts?

philippjfr:
Your efforts on these PRs is heroic, but please make PRs like this self-contained in future the merge conflicts are a nightmare.

maximlt:
Thanks @philippjfr !!! 🙏 I didn't feel brave enough to deal with these crazy conflicts on my first day back from holidays :D

MarcSkovMadsen:
> Your efforts on these PRs is heroic, but please make PRs like this self-contained in future the merge conflicts are a nightmare.

I would also like to avoid merge conflicts. But technically I don't know how to do this if one PR has to build on another PR. How can I do this differently? (At work this almost always work. Why does it not work here?)

philippjfr:
> I was asked not to make one big PR.

I think this is quite context dependent, I would certainly have suggested that improving docstrings should have been a single PR, though separate PRs would have been fine if they didn't build on each other. For code I get the fact that each PR has to build on the other but I'm a little perplexed why that would be needed for docstrings. Certainly I would **never** suggest creating more than two separate PRs that depend on each other because you end up with this awful daisy chain of merge conflicts. This was probably the worst case scenario of that because the main issues were in the base PR, i.e. the Ruff PR with all the wrongly formatted docstrings. In any case, we're mostly done now.

maximlt:
Yes I suggested multiple PRs as I'm pretty sure that documenting some parts isn't going to be straightforward (e.g. documenting `parameters.py` without duplicating all the docstrings and making the module much longer to import) and I was concerned this would block merging any improvements at all! I also think we should give @jbednar a chance to chime in before merging all these docstring changes (even if it's just to say "go ahead!").

> As far as I can see the merge conflicts have been resolved

FYI not in https://github.com/holoviz/param/pull/998

Oh and I've just noticed you added type hints. It's nice but also more work for reviewers (shouldn't we set up `mypy` before adding type hints?). From a quick look, I saw for example that the docstring of `watch` shows that `what` is optional but the type hint doesn't. Which one is correct?

```python
    def watch(
        self_, fn, parameter_names: list[str], what: str='value', onlychanged: bool=True,
        queued: bool=False, precedence: int=0
    ) -> Watcher:
        ...
        what : str, optional
            The type of change to watch for. By default, this is 'value', but it
            can be set to other slots such as 'constant'. Default is 'value'.
```

MarcSkovMadsen:
> I saw for example that the docstring of `watch` shows that `what` is optional but the type hint doesn't. Which one is correct?
> 
> 
> 
> ```python
> 
>     def watch(
> 
>         self_, fn, parameter_names: list[str], what: str='value', onlychanged: bool=True,
> 
>         queued: bool=False, precedence: int=0
> 
>     ) -> Watcher:
> 
>         ...
> 
>         what : str, optional
> 
>             The type of change to watch for. By default, this is 'value', but it
> 
>             can be set to other slots such as 'constant'. Default is 'value'.
> 
> ```

Thx.

I think the type annotation and docstring is correct and consistent:

- You dont type annotate that an argument has a default value and its optional to provide a value
- the docstring "optional" does not refer to a type annotation. It means a default value has been set and that its optional to provide a value.

maximlt:
Ah true! I got confused with `typing.Optional`.

philippjfr:
As for the general typing question. I certainly would like to fully type param in the near future. I'm personally okay with adding a few types here or there until we start on that effort but if we want to add some basic mypy validation to the test suite first I'm also okay with that.

param/parameterized.py
```diff
@@ -3381,18 +3966,67 @@ def debug(self_,msg,*args,**kw):
...
+        -------
+        None
+
+        Raises
+        ------
+        Exception
+            If the logging level is `WARNING` and warnings are treated as
+            exceptions (`warnings_as_exceptions` is True).
+
+        Notes
```
maximlt, line 4006: Redundant with the text above.

param/parameterized.py
```diff
@@ -3381,18 +3966,67 @@ def debug(self_,msg,*args,**kw):
         """
         self_.__db_print(DEBUG,msg,*args,**kw)
 
-    def log(self_, level, msg, *args, **kw):
+    def log(
```
maximlt, line 3969: That doesn't fit on a single line?

param/parameterized.py
```diff
@@ -3381,18 +3966,67 @@ def debug(self_,msg,*args,**kw):
...
         """
-        Print msg merged with args as a message at the indicated logging level.
+        Log a message at the specified logging level.
 
-        Logging levels include those provided by the Python logging module
-        plus VERBOSE, either obtained directly from the logging module like
-        `logging.INFO`, or from parameterized like `param.parameterized.INFO`.
+        This method logs a message constructed by merging `msg` with `args` at
+        the indicated logging level. It supports logging levels defined in
+        Python's `logging` module.
```
maximlt, line 3981: The original text described `VERBOSE` but the new one doesn't. Why?

param/parameterized.py
```diff
@@ -3246,19 +3693,157 @@ def _watch(self_, fn, parameter_names, what='value', onlychanged=True, queued=Fa
...
+            The precedence level of the watcher. Lower precedence values are executed
+            earlier. User-defined watchers must use positive precedence values.
+            Default is 0.
+
+        Returns
+        -------
+        Watcher
+            The `Watcher` object encapsulating the registered callback.
+
+        Raises
```
maximlt, line 3804: Let's remove these please.

param/parameterized.py
```diff
@@ -3246,19 +3693,157 @@ def _watch(self_, fn, parameter_names, what='value', onlychanged=True, queued=Fa
...
+        parameter_names : str or list of str
+            The name(s) of the parameters to monitor. Can be a single parameter
+            name, a list of parameter names, or a tuple of parameter names.
+        what : str, optional
+            The type of change to watch for. Must be 'value'. Default is 'value'.
+        onlychanged : bool, optional
+            If True (default), the callback is only invoked when the parameter value
+            changes. If False, the callback is invoked even when the parameter is
+            set to its current value.
+        queued : bool, optional
```
maximlt, line 3789: Same comment as `watch`.

param/parameterized.py
```diff
@@ -3453,11 +4087,75 @@ def _state_pop(self_):
...
+
+        Raises
+        ------
+        NotImplementedError
+            If the method is called at the class level instead of an instance
+            of `Parameterized`.
+
+        Notes
+        -----
+        - This method is experimental and subject to change in future versions.
```
maximlt, line 4138: I don't think that's still true.

param/parameterized.py
```diff
@@ -4383,51 +5081,125 @@ def __setstate__(self, state):
...
-
-    Message formatting: Each Parameterized instance has several
-    methods for optionally printing output. This functionality is
-    based on the standard Python 'logging' module; using the methods
-    provided here, wraps calls to the 'logging' module's root logger
-    and prepends each message with information about the instance
-    from which the call was made. For more information on how to set
-    the global logging level and change the default message prefix,
-    see documentation for the 'logging' module.
+    Base class for named objects with observable Parameters.
```
maximlt, line 5084: `observable` isn't used in Param's terminology.

param/parameterized.py
```diff
@@ -4383,51 +5081,125 @@ def __setstate__(self, state):
...
+
+    The `Parameterized` base class simplifies your codebase, making it more robust and maintainable,
```
maximlt, line 5086: I think this part is a bit weird in the `Parameterized` docstring: "rich, interactive applications ... HoloViz ... ".

param/parameterized.py
```diff
@@ -4383,51 +5081,125 @@ def __setstate__(self, state):
...
+
+    Features
+    --------
+    1. **Parameters**
+    - Support for default, constant, and readonly values.
+    - Validation, documentation, custom labels, and parameter references.
+    2. **`param` Namespace**
+    - Provides utility methods for tasks like adding parameters, updating parameters,
+    debugging, pretty-printing, logging, serialization, deserialization, and more.
+    3. **Observer Pattern**
```
maximlt, line 5098: Again, "observer pattern" is terminology not used in Param's documentation.

param/parameterized.py
```diff
@@ -4383,51 +5081,125 @@ def __setstate__(self, state):
...
+    Changed my_number from 5 to 7
+
+    `watch` is the most low level, reactive api. For most use cases we recommend
+    using the higher level `depends`, `bind` or `rx` apis.
     """
 
     name = String(default=None, constant=True, doc="""
-        String identifier for this object.""")
+        String identifier for this object.
+        Default is the object's class name plus a unique integer""")
```
maximlt, line 5155: That's not true, at the class level the default is just the object's class name. Please don't change this.

param/parameterized.py
```diff
@@ -4383,51 +5081,125 @@ def __setstate__(self, state):
...
+        **Default Naming**
+
+        >>> obj.name
+        'MyClass12345'  # Default name: class name + unique identifier.
+
+        **Handling Invalid Parameters**
+
+        If a keyword does not match a defined parameter, it raises a TypeError:
+
+        >>> obj = MyClass(nonexistent_param=42)  # TypeError: MyClass.__init__() got an unexpected keyword argument 'nonexistent_param'
```
maximlt, line 5195: No need to document this imo, this is standard Python.

param/parameterized.py
```diff
@@ -4383,51 +5081,125 @@ def __setstate__(self, state):
...
+        >>> class MyClass(param.Parameterized):
+        ...     value = param.Number(default=10, bounds=(0, 20))
+        >>> obj = MyClass(value=15)
+
+        The `value` parameter is set to 15 for this instance, overriding the default.
+
+        **Default Naming**
+
+        >>> obj.name
+        'MyClass12345'  # Default name: class name + unique identifier.
```
maximlt, line 5189:
```suggestion
        'MyClass00001'  # Default name: class name + unique identifier.
```

param/parameterized.py
```diff
@@ -4383,51 +5081,125 @@ def __setstate__(self, state):
...
+        Parameters can be supplied as keyword arguments (`param_name=value`), overriding
+        their default values for this specific instance. Any parameters not explicitly
+        set will retain their defined default values.
+
+        If no `name` parameter is provided, the instance's `name` attribute will default
+        to a unique string composed of the class name followed by a unique 5-digit suffix.
+
+        Parameters
+        ----------
+        **params : dict
```
maximlt, line 5170: See https://numpydoc.readthedocs.io/en/latest/format.html#parameters

```suggestion
        **params
```

param/parameterized.py
```diff
@@ -4383,51 +5081,125 @@ def __setstate__(self, state):
...
+
+    Add a watcher to respond to parameter updates:
+
+    >>> def callback(event):
+    ...     print(f"Changed {event.name} from {event.old} to {event.new}")
+    >>> obj.param.watch(callback, 'my_number')
+    >>> obj.my_number = 7
+    Changed my_number from 5 to 7
+
+    `watch` is the most low level, reactive api. For most use cases we recommend
```
maximlt, line 5149: I'm not sure `watch` is what we call a "reactive" API. Also I think it's usually written API or APIs, not api or apis.

param/parameterized.py
```diff
@@ -4383,51 +5081,125 @@ def __setstate__(self, state):
 
 class Parameterized(metaclass=ParameterizedMetaclass):
     """
```
maximlt, line 5083: The Numpy docstring guide seems to say that you should document `__init__` in the class docstring (here), see https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring. That's what Pandas does for example https://github.com/pandas-dev/pandas/blob/b8a4691647a8850d681409c5dd35a12726cd94a1/pandas/core/frame.py#L505.
So the `Parameters` section. you added to `__init__` should be moved to the class docstring. Apparently you could also add an `Attributes` section to document `name` and `param`.

maximlt (changes requested):
Note that there are some comments I haven't repeated in all the places where they were relevant (e.g. when `Returns` is specified with `None`).

param/parameterized.py
```diff
@@ -1933,7 +1938,29 @@ def watchers(self_, value):
         self_.self._param__private.watchers = value
 
     @property
-    def self_or_cls(self_):
+    def self_or_cls(self_) -> Union['Parameterized', Type['Parameterized']]:
```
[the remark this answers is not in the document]
MarcSkovMadsen: Its not the intent of these docstring PRs to change the api.

I would suggest that when this is made private the docstring is removed. As long as its public it should have a proper docstring.

param/parameterized.py
```diff
@@ -1947,19 +1974,23 @@ def __setstate__(self, state):
...
             setattr(self, k, v)
 
-    def __getitem__(self_, key):
-        """Retrieve the class or instance parameter by key.
+    def __getitem__(self_, key: str) -> Parameter:
+        """
+        Retrieve a Parameter by its key.
+
+        This method allows access to a class or instance Parameter using its name.
+        If accessed on an instance, the returned Parameter will be instantiated.
```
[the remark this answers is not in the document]
MarcSkovMadsen: I've removed the line

param/parameterized.py
```diff
@@ -2611,26 +2641,63 @@ def _cls_parameters(self_):
...
         """
         if self_.self is not None and not self_.self._param__private.initialized and instance is True:
             raise RuntimeError(
-                'Looking up instance Parameter objects (`.param.objects()`) until '
-                'the Parameterized instance has been fully initialized is not allowed. '
-                'Ensure you have called `super().__init__(**params)` in your Parameterized '
-                'constructor before trying to access instance Parameter objects, or '
-                'looking up the class Parameter objects with `.param.objects(instance=False)` '
-                'may be enough for your use case.',
+                'Cannot access instance parameters before the Parameterized instance '
```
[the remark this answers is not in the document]
MarcSkovMadsen: In my judgement the existing string was very hard to understand and very long. To me this string is much easier to comprehend.

param/parameterized.py
MarcSkovMadsen: I see this differently. The signature is `def __init__(self, **params):` thus in principle you can provide any argument you want. But its functionality of `param` - not standard python that raises an exception:

```python
  File "/home/jovyan/repos/private/param/script.py", line 4, in <module>
    obj = MyClass(nonexistent_param=42)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jovyan/repos/private/param/param/parameterized.py", line 5217, in __init__
    refs, deps = self.param._setup_params(**params)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jovyan/repos/private/param/param/parameterized.py", line 1775, in override_initialization
    ret = fn(self_, *args, **kw)
          ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jovyan/repos/private/param/param/parameterized.py", line 2093, in _setup_params
    raise TypeError(
TypeError: MyClass.__init__() got an unexpected keyword argument 'nonexistent_param'
```

 If this exception is not documented its not clear to the user what to expect.

param/parameterized.py
MarcSkovMadsen: Changed to `watch` is the most low level, event-driven API. For most use cases we recommend
    using the higher level `depends`, `bind` or `rx` APIs.

param/parameterized.py
MarcSkovMadsen: Thx. I'm a little bit in doubt what to do here because 

- the numpy recommendation was created long time before for example VS Code was a thing. In VS Code I believe I have seen the class docstring and the `__init__` docstring being used at different times by pylance tooltips.
- In param docs the `__init__` method is documented. See for example https://param.holoviz.org/reference/param/generated/param.Action.html#param.Action.__init__. If I remove the docstring from `__init__` to param docs will be empty there. Might not be a problem.

What should I do?

MarcSkovMadsen:
> Note that there are some comments I haven't repeated in all the places where they were relevant (e.g. when `Returns` is specified with `None`).

Sorry. I think I missed the the "`Returns` is specified with `None`" comment. What does this mean?

param/parameterized.py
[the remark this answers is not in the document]
maximlt: I don't think it's public.

param/parameterized.py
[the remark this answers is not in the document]
maximlt: Please do not make unrelated code changes in a PR dedicated to touching the docstrings only.

param/parameterized.py
maximlt: Ok yes true.

Note that the signature is in fact dynamically altered in Parameterized subclasses (except when they have an init signature different than `def __init__(self, **params): ` so that users in a REPL can get autocomplete on the constructor.

param/parameterized.py
maximlt: Someone will need to look more into that, without forgetting Jupyter as it's a major target for HoloViz packages.

> In param docs the __init__ method is documented

Yes I already mentioned that documenting the Parameters is tricky.

maximlt:
> Sorry. I think I missed the the "Returns is specified with None" comment. What does this mean?

This means that I'm not sure it's useful to document that?

```
Returns
--------
None
```

MarcSkovMadsen:
> > Sorry. I think I missed the the "Returns is specified with None" comment. What does this mean?
> 
> This means that I'm not sure it's useful to document that?
> 
> ```
> Returns
> --------
> None
> ```

I sort of agree. The problem is that if its not include then the user does not know if its intended or forgotten.

We have started using ruff at work. And with the rules that are standard at work its actually required to type annotate `-> None` for functions that do not return anything. Again because you don't know if the type annotation was forgotten.

param/parameterized.py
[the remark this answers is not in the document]
MarcSkovMadsen: I changed it because I got the error while testing the examples. I did not understand the error message. So I spent time trying to improve it. I do believe its an improvement.

Would it be ok to include anyways? Thx.

param/parameterized.py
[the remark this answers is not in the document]
MarcSkovMadsen: For me its public because it can be accessed via `param.parameterized.self_or_class`. There is no leading `_` in the path.

For me as a consumer of the library that would mean its public and something i can use.

param/parameterized.py
```diff
@@ -29,6 +29,7 @@
 from itertools import chain
 from operator import itemgetter, attrgetter
 from types import FunctionType, MethodType
+from typing import Any, Union, Literal, Iterable, Callable # When python 3.9 support is dropped replace Union with |
```
maximlt, line 32: I learned that `typing.Callable` and `typing.Iterable` are sort of deprecated, instead one should import them from `collections.abc`. See https://docs.python.org/3/library/typing.html#typing.Iterable.

maximlt:
> I sort of agree. The problem is that if its not include then the user does not know if its intended or forgotten.

Ok yes.

param/parameterized.py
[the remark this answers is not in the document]
maximlt: The most important message imo is that `super().__init__(**params)` has to be called *before* `self.param.objects()`. I don't think the updated message conveys that. I still prefer the original message overall.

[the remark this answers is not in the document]
maximlt: Besides that, I know I'm annoying with asking not to mix different changes in a single PR, but as a release manager this is the only sane way to produce an accurate changelog. This PR title for example isn't correct, it is really "Enhance docstrings of the .param namespace + Parameterized class + Add type hints + Enhance error message of .param.objects()" 🙃

param/parameterized.py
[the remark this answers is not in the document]
maximlt: I guess it's fine for this one. But please do not make the mistake of assuming that an API that doesn't start with `_` is necessarily public:

- This is just a convention, packages can adopt other ones (https://pandas.pydata.org/docs/reference/index.html)
- HoloViz packages don't do a great job at separating their private vs public API
- Param just recently started to make that distinction clearer (https://github.com/holoviz/param/pull/768)

For Param, I'd say the real public API is anything that is documented on the website.

maximlt:
Hi @MarcSkovMadsen, I'd like to ask you if you intend to work on this PR again in the near future? I'm cooking up the 2.3.0 release and would like to get this one in if possible. Let me know if you're not available to work on it, in which case I'd be happy to take it over to finish it.

MarcSkovMadsen:
Thx. @maximlt . Its more likely I won't have time in short time then I will.

I would be happy to see this finalized by you in short term or by me in longer term (months).

param/parameterized.py
maximlt: Skipping it for now.

param/parameterized.py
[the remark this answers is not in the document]
maximlt: Reverted, let's do this in another PR if need be.

param/parameterized.py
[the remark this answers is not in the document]
maximlt: Removed for now.

param/parameterized.py
```diff
@@ -2611,26 +2641,63 @@ def _cls_parameters(self_):
...
+            fully initialized. Ensure `super().__init__(**params)` is called in the
+            constructor before triggering watchers.
+
+        Notes
+        -----
+        - This method distinguishes between class-level and instance-specific parameters.
+        - Instance-specific parameters are lazily created; they are not initialized unless
+        explicitly requested or accessed.
+        - When `instance='existing'`, only parameters already initialized at the instance level
+        will be returned, while class-level parameters remain unaffected.
```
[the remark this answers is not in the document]
maximlt: Removed.

param/parameterized.py
```diff
@@ -2611,26 +2641,63 @@ def _cls_parameters(self_):
...
+            - `False`: Return class-level parameters without creating instance-specific copies.
+            - `'existing'`: Return only the instance parameters that already exist, avoiding
+            creation of new instance-specific parameters.
+
+        Returns
+        -------
+        dict[str, Parameter]
+            A dictionary mapping parameter names to their corresponding `Parameter` objects.
+
+        Raises
```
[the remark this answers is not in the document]
maximlt: Removed.

param/parameterized.py
```diff
@@ -2643,13 +2710,39 @@ def objects(self_, instance=True):
...
+        of type `Event`, the parameter value will be temporarily set to `True` to
+        indicate that the event has been triggered.
+
+        Parameters
+        ----------
+        *param_names : str
+            Names of the parameters to trigger. Each name must correspond to a
+            parameter defined on this `Parameterized` instance.
+
+        Raises
```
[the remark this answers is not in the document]
maximlt: Removed.

param/parameterized.py
```diff
@@ -2643,13 +2710,39 @@ def objects(self_, instance=True):
...
 
-    def trigger(self_, *param_names):
+    def trigger(self_, *param_names: str) -> None:
         """
-        Trigger watchers for the given set of parameter names. Watchers
-        will be triggered whether or not the parameter values have
-        actually changed. As a special case, the value will actually be
-        changed for a Parameter of type Event, setting it to True so
-        that it is clear which Event parameter has been triggered.
+        Trigger event handlers for the specified parameters.
```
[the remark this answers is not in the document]
maximlt: Improved.

param/parameterized.py
```diff
@@ -2643,13 +2710,39 @@ def objects(self_, instance=True):
...
+        This method activates all watchers associated with the given parameter names,
+        regardless of whether the parameter values have actually changed. For parameters
+        of type `Event`, the parameter value will be temporarily set to `True` to
+        indicate that the event has been triggered.
+
+        Parameters
+        ----------
+        *param_names : str
+            Names of the parameters to trigger. Each name must correspond to a
+            parameter defined on this `Parameterized` instance.
```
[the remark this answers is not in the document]
maximlt: Fixed.

param/parameterized.py
```diff
@@ -2643,13 +2710,39 @@ def objects(self_, instance=True):
...
+        Raises
+        ------
+        RuntimeError
+            If the method is called on a `Parameterized` instance that has not been
+            fully initialized. Ensure `super().__init__(**params)` is called in the
+            constructor before triggering watchers.
+
+        Examples
+        --------
+        >>> import param
```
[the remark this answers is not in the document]
maximlt: Updated for an example that triggers a watcher after appending to a list.

param/parameterized.py
```diff
@@ -2643,13 +2710,39 @@ def objects(self_, instance=True):
...
-        will be triggered whether or not the parameter values have
-        actually changed. As a special case, the value will actually be
-        changed for a Parameter of type Event, setting it to True so
-        that it is clear which Event parameter has been triggered.
+        Trigger event handlers for the specified parameters.
+
+        This method activates all watchers associated with the given parameter names,
+        regardless of whether the parameter values have actually changed. For parameters
+        of type `Event`, the parameter value will be temporarily set to `True` to
+        indicate that the event has been triggered.
```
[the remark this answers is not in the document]
maximlt: Preserved.

param/parameterized.py
```diff
@@ -2781,24 +2877,30 @@ class or instance that contains an iterable collection of
...
-    def serialize_parameters(self_, subset=None, mode='json'):
+    def serialize_parameters(self_, subset: Union[Iterable[str], None]=None, mode='json'):
         """
         Return the serialized parameters of the Parameterized object.
 
         Parameters
         ----------
-        subset : list, optional
-            A list of parameter names to serialize. If None, all parameters will be serialized. Defaults to None.
+        subset : iterable, optional
```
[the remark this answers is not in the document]
maximlt: Updated for `iterable of str`.

param/parameterized.py
```diff
@@ -2865,27 +3015,126 @@ def deserialize_parameters(self_, serialization, subset=None, mode='json') -> di
...
+        >>> schema
+        {'name': {'anyOf': [{'type': 'string'}, {'type': 'null'}],
+         'description': "String identifier for this object. Default is the object's class name plus a unique integer",
+         'title': 'Name'},
+        'a': {'type': 'number',...
+        }
+        """
         if mode not in Parameter._serializers:
             raise ValueError(f'Mode {mode!r} not in available serialization formats {list(Parameter._serializers.keys())!r}')
+        self_or_cls = self_.self_or_cls
```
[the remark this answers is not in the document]
maximlt: Not sure why but keeping it as it doesn't affect the code.

param/parameterized.py
```diff
@@ -2957,14 +3234,44 @@ def force_new_dynamic_value(self_, name): # pylint: disable-msg=E0213
         else:
             return param_obj._force(slf, cls)
 
-    def get_value_generator(self_,name): # pylint: disable-msg=E0213
+    def get_value_generator(self_,name: str)->Any: # pylint: disable-msg=E0213
```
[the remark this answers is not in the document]
maximlt: Done

param/parameterized.py
```diff
@@ -2957,14 +3234,44 @@ def force_new_dynamic_value(self_, name): # pylint: disable-msg=E0213
...
 
-        For most parameters, this is simply the parameter's value
-        (i.e. the same as getattr()), but Dynamic parameters have
-        their value-generating object returned.
+        Returns
+        -------
+        Any
+            The current value of the parameter, a value-generating object for
+            `Dynamic` parameters, or a list of value-generating objects for
+            `CompositeParameter` parameters.
```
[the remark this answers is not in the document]
maximlt: Removed.

param/parameterized.py
```diff
@@ -2992,12 +3299,41 @@ def get_value_generator(self_,name): # pylint: disable-msg=E0213
 
         return value
 
-    def inspect_value(self_,name): # pylint: disable-msg=E0213
+    def inspect_value(self_,name: str)->Any: # pylint: disable-msg=E0213
```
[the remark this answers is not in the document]
maximlt: Done

param/parameterized.py
```diff
@@ -3016,14 +3352,41 @@ def inspect_value(self_,name): # pylint: disable-msg=E0213
 
         return value
 
-    def method_dependencies(self_, name, intermediate=False):
+    def method_dependencies(self_, name: str, intermediate: bool=False) -> list[PInfo]:
         """
-        Given the name of a method, returns a PInfo object for each dependency
-        of this method. See help(PInfo) for the contents of these objects.
+        Retrieve the parameter dependencies of a specified method.
```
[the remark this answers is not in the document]
maximlt: Re-added.

param/parameterized.py
```diff
@@ -2992,12 +3299,41 @@ def get_value_generator(self_,name): # pylint: disable-msg=E0213
...
+        ----------
+        name : str
+            The name of the parameter whose value is to be inspected.
+
+        Returns
+        -------
+        Any
+            The current value of the parameter, the last generated value for
+            `Dynamic` parameters, or a list of inspected values for composite
+            parameters.
```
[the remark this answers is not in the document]
maximlt: Done.

param/parameterized.py
```diff
@@ -3195,42 +3602,82 @@ def _register_watcher(self_, action, watcher, what='value'):
                     watchers[what] = []
                 getattr(watchers[what], action)(watcher)
 
-    def watch(self_, fn, parameter_names, what='value', onlychanged=True, queued=False, precedence=0):
+    def watch(
+        self_, fn, parameter_names: list[str], what: str='value', onlychanged: bool=True,
```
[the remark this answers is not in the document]
maximlt: Fixed.

param/parameterized.py
```diff
@@ -3195,42 +3602,82 @@ def _register_watcher(self_, action, watcher, what='value'):
...
+        ----------
+        fn : callable
+            The callback function to invoke when an event occurs. This function
+            will be provided with `Event` objects as positional arguments, allowing
+            it to determine the triggering events.
+        parameter_names : list[str]
+            A list of parameter names to watch for events.
+        what : str, optional
+            The type of change to watch for. By default, this is 'value', but it
+            can be set to other slots such as 'constant'. Default is 'value'.
```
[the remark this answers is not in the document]
maximlt: Fixed.

param/parameterized.py
```diff
@@ -3195,42 +3602,82 @@ def _register_watcher(self_, action, watcher, what='value'):
...
+        onlychanged : bool, optional
+            If True (default), the callback is only invoked when the watched
+            item changes. If False, the callback is invoked even when the `what`
+            item is set to its current value.
```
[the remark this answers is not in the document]
maximlt: Kept as is.

param/parameterized.py
```diff
@@ -3195,42 +3602,82 @@ def _register_watcher(self_, action, watcher, what='value'):
...
+        queued : bool, optional
+            If False (default), additional watcher events generated inside the
```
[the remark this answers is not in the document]
maximlt: Re-added.

param/parameterized.py
```diff
@@ -3195,42 +3602,82 @@ def _register_watcher(self_, action, watcher, what='value'):
...
+            executed earlier. User-defined watchers must use positive precedence
+            values. Negative precedences are reserved for internal watchers
+            (e.g., those set up by `param.depends`). Default is 0.
+
+        Returns
+        -------
+        Watcher
+            The `Watcher` object that encapsulates the registered callback.
+
+        Raises
```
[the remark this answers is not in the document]
maximlt: Removed.

param/parameterized.py
```diff
@@ -3246,19 +3693,157 @@ def _watch(self_, fn, parameter_names, what='value', onlychanged=True, queued=Fa
...
+
+        Parameters
+        ----------
+        watcher : Watcher
+            The `Watcher` object to remove. This should be an object returned
+            by a previous call to `watch` or `watch_values`.
+
+        Returns
+        -------
+        None
```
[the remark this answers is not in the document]
maximlt: Removed.

param/parameterized.py
```diff
@@ -3246,19 +3693,157 @@ def _watch(self_, fn, parameter_names, what='value', onlychanged=True, queued=Fa
...
+
+        Notes
+        -----
+        - If the watcher does not exist in the list of registered watchers,
+        the method logs a warning message instead of silently failing.
```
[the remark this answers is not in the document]
maximlt: Removed as this is being fixed separately.

param/parameterized.py
maximlt: Removed.

param/parameterized.py
maximlt: Undone.

param/parameterized.py
maximlt: Re-added.

param/parameterized.py
maximlt: Removed.

param/parameterized.py
maximlt: Done.

param/parameterized.py
maximlt: Removed.

param/parameterized.py
maximlt: Removed.

param/parameterized.py
maximlt: I tried it in VSCode and JupyterLab and was satisfied with the experience, the longer class docstring showing up on `Parameterized` and the shorter `__init__` one when instantiating a Parameterized object.

param/parameterized.py
maximlt: Updated.

param/parameterized.py
maximlt: Updated.

param/parameterized.py
maximlt: Removed.

param/parameterized.py
maximlt: Updated.

maximlt (approved):
Thank you so much @MarcSkovMadsen, massive improvement to Param!

PR #998, opened 2024-12-29
Other Docstring improvements

MarcSkovMadsen:
Starts from https://github.com/holoviz/param/pull/997. Please review that one first.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -29,7 +29,7 @@ from html import escape
 from itertools import chain
 from operator import itemgetter, attrgetter
 from types import FunctionType, MethodType
-from typing import Type, Union, Literal # When python 3.9 support is dropped replace Union with |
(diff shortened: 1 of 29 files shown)


param/parameterized.py
```diff
@@ -1168,97 +1214,106 @@ class Foo(Bar):
...
-        is passed in.
-
-        nested_refs: if True and allow_refs=True then even nested objects
-        such as dictionaries, lists, slices, tuples and sets will be
-        inspected for references and will be automatically resolved.
-
-        default, doc, and precedence all default to None, which allows
-        inheritance of Parameter slots (attributes) from the owning-class'
-        class hierarchy (see ParameterizedMetaclass).
+    def __init__( # pylint: disable-msg=R0913
```
MarcSkovMadsen, line 1224: I would have like to type annotate `Parameter.__init__` but because of the overloading and use of `Undefined` for example the type annotation of `doc` would become `Union[str, None, _Undefined]`. I'm not really sure that is helpful to users?

image

Issue #999, opened 2025-01-02
Deprecate `param/version.py`


maximlt:
We recently archived [autover](https://github.com/holoviz-dev/autover) but I completely forgot the implementation actually used across HoloViz lives is in `param/version.py`.

A few packages still depend on it but I don't see that as a blocker, those still active will need to migrate to an alternative solution (e.g. setuptools-scm) like we did recently for other packages. Other potential users of `param/version.py` are also highly encouraged to migrate immediately:

- [ ] jupyter-panel-proxy
- [x] pyct => migrated to hatch-vcs
- [x] earthsim => archived
- [x] nbsite => migrated to hatch-vcs
- [x] nbsmoke => archived

PR #1002, opened 2025-01-03
Deprecate version.py

maximlt:
Implements https://github.com/holoviz/param/issues/999

There are probably fancier ways to implement that but I think that will do the job.


jlstevens:
Looks good to me!

I think the main thing we need to check is that nothing in holoviz is using `param.Version` before the next version of param is released. That way, this warning should only be seen by non-holoviz users of `Version` of which I suspect there aren't that many (if any).

maximlt:
All the projects that used the version module have been either archived or updated to use another tool. Test failures are unrelated.

Issue #508, opened 2021-07-08
factory function for Parameter instead of default value + instantiate=True


tomas16:
#### Is your feature request related to a problem? Please describe.

I'm trying to use composition of 2 parameterized classes in which the top-level class has methods that depend on parameters in the subobjects. The subobjects have a `view` attribute which is essentially `pn.Param(self.param)` with some options for the widgets.

Below code doesn't work because when `pm.depends` is evaluated, `self.a` doesn't exist yet, so `'a.value'` can't get evaluated (it generates an AttributeError). This is still true when `self.a = A()` happens before `super().__init__()`.

```python
import param as pm


class A(pm.Parameterized):
    value = pm.Parameter(default=123)


class B(pm.Parameterized):
    a = pm.Parameter()

    def __init__(self):
        super().__init__()
        self.a = A()

    @pm.depends('a.value', watch=True)
    def _value_update(self):
        print(self.a.value)


if __name__ == '__main__':
    b = B()
    b.a.value = 456
```

I tried instantiating an `A` object as a default value instead, redefining `class B` like this:

```python
class B(pm.Parameterized):
    a = pm.Parameter(default=A(), instantiate=True)

    @pm.depends('a.value', watch=True)
    def _value_update(self):
        print(self.a.value)
```

This "works" in this toy example, but when A has a view attribute that contains panel GUI controls, the GUI controls don't work anymore. I suspect this is due to the way the default A instance is deepcopied when B objects are instantiated.

#### Describe the solution you'd like

Instead of having a default value that gets deepcopied, I propose supplying the `Parameter` constructor with a factory function to create a default value. It sidesteps all issues in cases where copy behavior gets complicated.

#### Describe alternatives you've considered

In my case, I was able to solve the problem by making A's `view` attribute a cached property. In python 3.7:

```python
@property
@functools.lru_cache(maxsize=None)
def view(self):
    ...
```

MarcSkovMadsen:
Hi @tomas16 

If you provide `A()` to `super.__init__` then it will work

```python
import param as pm

class A(pm.Parameterized):
    value = pm.Parameter(default=123)

class B(pm.Parameterized):
    a = pm.Parameter()

    def __init__(self):
        super().__init__(a=A())

    @pm.depends('a.value', watch=True)
    def _value_update(self):
        print(self.a.value)


if __name__ == '__main__':
    b = B()
    b.a.value = 456
```

```bash
$ python 'script.py'
456
```

**Does this solve your problem?**

[...]

PR #1012, opened 2025-01-17
Deprecate Selector's compute_default_fn slot and compute_default() method

maximlt:
As suggested in https://github.com/holoviz/param/issues/508#issuecomment-2593954649.

### DIFF
--- a/doc/user_guide/Parameter_Types.ipynb
+++ b/doc/user_guide/Parameter_Types.ipynb
@@ -826,8 +826,6 @@
     "\n",
     "Because `Selector` is usually used to allow selection from a list of existing (instantiated) objects, `instantiate` is False by default, but you can specify `instantiate=True` if you want each copy of this Parameter value to be independent of other instances and superclasses.\n",
     "\n",
-    "In cases where the objects in the list cannot be known when writing the Parameterized class but can be calculated at runtime, you can supply a callable (of no arguments) to `compute_default_fn`, and then ensure that at runtime you call `compute_default` on that Parameter to initialize the value.\n",
-    "\n",
     "A `param.ListSelector` works just the same as a regular Selector, but the value is a _list_ of valid objects from the available objects, rather than just one. Each item in the list is checked against the `objects`, and thus the current value is thus a _subset_ of the `objects`, rather than just one of the objects.\n",
     "\n",
     "A `param.FileSelector` works like a regular Selector with the value being a filename and the `objects` being computed from files on a file system. The files are specified as a `path` [glob](https://docs.python.org/3/library/glob.html), and all filenames matching the glob are valid `objects` for the parameter.\n",
--- a/param/parameters.py
+++ b/param/parameters.py
@@ -40,6 +40,7 @@ from .parameterized import (
 )
 from ._utils import (
     ParamFutureWarning as _ParamFutureWarning,
+    ParamDeprecationWarning as _ParamDeprecationWarning,
     _deprecate_positional_args,
     _deprecated,
     _validate_error_prefix,
@@ -1868,6 +1869,13 @@ class Selector(SelectorBase, _SignatureSelector):
                  compute_default_fn=Undefined, check_on_set=Undefined,
                  allow_None=Undefined, empty_default=False, **params):
 
+        if compute_default_fn is not Undefined:
+            warnings.warn(
+                'compute_default_fn has been deprecated and will be removed in a future version.',
+                _ParamDeprecationWarning,
+                stacklevel=3,
+            )
+
         autodefault = Undefined
         if objects is not Undefined and objects:
             if isinstance(objects, dict):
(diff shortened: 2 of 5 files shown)


maximlt:
@philippjfr @jbednar do you agree with this deprecation?

jbednar (approved):
I agree with this deprecation (and in fact I think I suggested it a while back :-) ). I see you've removed documentation of the old way to do it, so I assume that the new way to do it has already been documented appropriately in a separate PR?

maximlt:
> I agree with this deprecation (and in fact I think I suggested it a while back :-) ). I see you've removed documentation of the old way to do it, so I assume that the new way to do it has already been documented appropriately in a separate PR?

I'm working separately on `default_factory` but haven't opened a PR yet (you will know about it!). The main motivation is to deprecate `compute_default_fn/compute_defaut()` is that it is imo a weird API and it appears to be unused (see more details in https://github.com/holoviz/param/issues/508#issuecomment-2593954649, with only 3 occurrences found on Github in topographica). As such I don't see a need to document how to reproduce what this feature was achieving. For the users who might need this feature, they can easily reproduce it as the code is trivial:

```python
    def compute_default(self):
        """
        If this parameter's compute_default_fn is callable, call it
        and store the result in self.default.

        Also removes None from the list of objects (if the default is
        no longer None).
        """
        if self.default is None and callable(self.compute_default_fn):
            self.default = self.compute_default_fn()
            self._ensure_value_is_in_objects(self.default)

    def _ensure_value_is_in_objects(self, val):
        """
        Make sure that the provided value is present on the objects list.
        Subclasses can override if they support multiple items on a list,
        to check each item instead.
        """
        if val not in self.objects:
            self._objects.append(val)
```

maximlt:
Inserting below and updating the bit of text from [XXX](https://github.com/holoviz/param/issues/508#issuecomment-2593954649) that is relevant to this PR and supports the answer to:

>  I agree with this deprecation (and in fact I think I suggested it a while back :-) ). I see you've removed documentation of the old way to do it, so I assume that the new way to do it has already been documented appropriately in a separate PR?

The documentation removed is:

> In cases where the objects in the list cannot be known when writing the Parameterized class but can be calculated at runtime, you can supply a callable (of no arguments) to `compute_default_fn`, and then ensure that at runtime you call `compute_default` on that Parameter to initialize the value

I don't think compute_default_fn really does what was described in this paragraph. If I don't know in advance the list of objects:
- For a Selector, I would like to populate the list of all objects, which is not what this does, as it only sets objects as a list of one item (`[default]`).
- For a ListSelector, it does populate the list of all objects, but also sets the default to the whole list, which is quite arbitrary (in the apps I work on the best default for a ListSelector, driving a MultiSelect widget, is None).

compute_default_fn also only sets the Parameter `default` and `objects` slots, which does not affect the Parameterized instance attribute value, which is awkward.

So, I don't think there's any documentation to replace.

---

`compute_default_fn` is a slot of `Selector` (and all its subclasses). It accepts a callable that is **called only when the user calls the `compute_default()` method** on Selector. When the `default` attribute of a Parameter is not `None`, `compute_default()` sets `default` to the value returned by the callable.

This is unrelated to `compute_default_fn`, but note that `default` and the instance parameter value aren't linked once a Parameterized class is instantiated.

```python
class P(param.Parameterized):
    x = param.Parameter()

p = P()
p.param['x'].default = 'something'
# Setting `default` doesn't affect the current value of `x` on `p`.
assert p.x is None
```

So calling `compute_default()` on a Parameterized instance isn't so useful when it comes to updating `default`; it however does update `objects` if required, which is slightly more useful.

```python
class P(param.Parameterized):
    x = param.Selector(compute_default_fn=lambda: 3)

p = P()
p.param['x'].compute_default()
assert p.x != 3
assert p.x is None
assert p.param['x'].objects == [3]
```

So `compute_default()` has to be called before the instance is initialized.

Either in the init:
```python
class P(param.Parameterized):
    x = param.Selector(compute_default_fn=lambda: 3)

    def __init__(self, **params):
        self.param['x'].compute_default()
        super().__init__(**params)

p = P()
assert p.x == 3
assert p.param['x'].objects == [3]

# Bad idea as the parameter modified is the class-parameter!
assert P.param.x.default == 3
```

But obviously that's a bad idea. Or on the class object:

```python
class P(param.Parameterized):
    x = param.Selector(compute_default_fn=lambda: 3)

P.param['x'].compute_default()
p = P()
assert p.x == 3
assert p.param['x'].objects == [3]
```

Note that if the `Selector` had `objects` defined (excluding `None` as the first item), calling `compute_default()` would not do anything, as it only works if `default` is `None` (`Selector` auto sets `default` to the first item of `objects`).

On a list flavor Selector, `compute_default_fn` can extend the list of objects.

```python
class P(param.Parameterized):
    x = param.ListSelector(objects=[0, 1, 2], compute_default_fn=lambda: [3, 4])

P.param['x'].compute_default()
p = P()
assert p.x == [3, 4]
assert p.param['x'].objects == [0, 1, 2, 3, 4]
```

So basically, the functionality provided by `compute_default_fn` and the `compute_default` method can be implemented with this function that needs to be invoked with a parameter and a callable:
```python
def compute_default(parameter, fn):
    if parameter.default is None:
        parameter.default = default = fn()
        objects = parameter.objects
        if isinstance(parameter, param.ListSelector):
            for o in default:
                if o not in objects:
                    objects.append(o)
        else:
            if default not in objects:
                objects.append(default)
```

Searching for usage of `compute_default_fn` on Github, I only found 3 occurrences in *topographica*. Clearly, this isn't used much :)

https://github.com/holoviz/param/pull/1092 aims to add the `default_factory` slot to Param Parameters. This is very different from `compute_default`, as:
- the factory is always invoked by Param on instantiation, and can additionally be invoked on class creation (when the callable is wrapped in a `DefaultFactory` instance set `on_class=True`)
- the factory sets the instance attribute *value* and not the parameter *default* attribute

Issue #1014, opened 2025-02-10
Unexpected behaviour of param.edit_constant() context manager


dinya:
From [``param.edit_constant()``](https://param.holoviz.org/reference/param/generated/param.edit_constant.html#param-edit-constant) docstring:

> Temporarily set parameters on Parameterized object to constant=False to allow editing them.

Should "temporarily" mean "set editable inside context manager and unset outside"?

If so the following behaviour is unexpected.

```python
import param

class A(param.Parameterized):
    val = param.Integer(1, constant=True)

a = A()

#print(a.param.val.constant, a.param.val.readonly)

with param.edit_constant(a):
    a.val = 6

print(a.param.val.constant, a.param.val.readonly)

# "TypeError: Constant parameter 'val' cannot be modified" is expected to be raised:
a.val = 7
```

... but nothing is raised and it prints:

```python
False False
```

But if I change my code slightly (uncomment first print)

```python
import param

class A(param.Parameterized):
    val = param.Integer(1, constant=True)

a = A()

print(a.param.val.constant, a.param.val.readonly)
# or just call a.param.val

with param.edit_constant(a):
    a.val = 6

print(a.param.val.constant, a.param.val.readonly)

# "TypeError: Constant parameter 'val' cannot be modified" is expected to be raised:
a.val = 7
```

then it prints

```python
True False
True False
```

and ``TypeError: Constant parameter 'val' cannot be modified`` is raised as expected.

It seems that the ``param.edit_constant()`` context manager doesn't restore the state if ``a.param.val`` has not been called before.

------------

My environment:

1. python 3.11.9
2. param 2.2.0

maximlt:
Hi @dinya, it seems like when you copied your code over to this issue the indentation got wrong, for instance this is not valid Python:

```python
with param.edit_constant(a):
a.val = 6
```

Could you please edit your post to fix the indentation?

dinya:
@maximal Thanks, I fixed it.

maximal:
@dinya, you‘re welcome!

maximlt:
This is indeed a bug and a potential duplicate of https://github.com/holoviz/param/issues/931, thanks for reporting it @dinya !

Issue #931, opened 2024-04-04
Initialization changes constant attribute.


LinuxIsCool:
Thanks for contacting us! Please read and follow these instructions carefully, then delete this introductory text to keep your issue easy to read. Note that the issue tracker is NOT the place for usage questions and technical assistance; post those at [Discourse](https://discourse.holoviz.org) instead. Issues without the required information below may be closed immediately.

#### ALL software version info
panel==1.4.0
param==2.1.0
Python 3.10.12

#### Description of expected behavior and the observed behavior

Expecting normal initialization. Unfortunately, constant is being set to False on my parameter. 

I would expect `invariant_coef` to remain constant.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import param as pm
import panel as pn
pn.extension()

class BondingCurve(pm.Parameterized):
    supply = pm.Number(80)
    reserve = pm.Number(40)
    kappa = pm.Number(2)
    invariant_coef = pm.Number(constant=True)
    price = pm.Number(constant=True)
    
    # def __init__(self, **params):
    #     super().__init__(**params)
    #     self.set_invariant_coef()
    
    @pm.depends(on_init=True, watch=True)
    def set_invariant_coef(self):
        with pm.edit_constant(self):
            self.invariant_coef = self.supply ** self.kappa / self.reserve

bc = BondingCurve()

pn.panel(bc)

bc.param['invariant_coef'].constant

```

> False

![image](https://github.com/holoviz/param/assets/31582215/b0c0faa5-ae6a-436a-acd1-2fab5ce08f4f)

Hoxbro:
This context manager seems to work not sure about what 'existing' does

``` python
import param as pm
import panel as pn
from contextlib import contextmanager


@contextmanager
def edit_constant(parameterized):
    """
    Temporarily set parameters on Parameterized object to constant=False
    to allow editing them.
    """
    params = parameterized.param.objects().values()  # changed from: 
    # params = parameterized.param.objects('existing').values()
    constants = [p.constant for p in params]
    for p in params:
        p.constant = False
    try:
        yield
    finally:
        for p, const in zip(params, constants):
            p.constant = const


class BondingCurve(pm.Parameterized):
    price = pm.Number(constant=True)

    @pm.depends(on_init=True, watch=True)
    def set_ex(self):
        with edit_constant(self):
            self.price = 1


bc = BondingCurve()

bc.param.price.constant, bc.price
```

LinuxIsCool:
[...]

PR #1015, opened 2025-02-11
Ensure edit_constant resets class and instance level parameters

maximlt:
Fixes https://github.com/holoviz/param/issues/1014
Fixes https://github.com/holoviz/param/issues/931

Regression introduced in Param 2.0.0 in https://github.com/holoviz/param/pull/826 that introduced installing an instance parameter (copied from the class) on `__set__` (https://github.com/holoviz/param/commit/6ce1ef3f3ecc01d0c3eca88b0ec50d690dfb40e1). The context manager `edit_constant` was first getting a handle on all the existing parameters (mix of class and instance level parameters), yielding, and then resetting `constant` on the parameters already handled, missing that the parameterized object could have been equipped with instance-level parameters that were created with `constant=False`.

This is fixed in this MR, which also only touches the constant parameters.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -338,15 +338,23 @@ def edit_constant(parameterized):
     Temporarily set parameters on Parameterized object to constant=False
     to allow editing them.
     """
-    params = parameterized.param.objects('existing').values()
-    constants = [p.constant for p in params]
-    for p in params:
-        p.constant = False
+    kls_params = parameterized.param.objects(instance=False)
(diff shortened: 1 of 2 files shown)


jbednar (approved):
Seems correct to me. Thanks!

maximlt:
The version with this fix hasn't yet been released. I'll be working soon on preparing a release that will include it, release which will happen before my holidays start in August.

dansawula:
Ah ok, thank you!

Issue #1001, opened 2025-01-02
constant Parameter attribute not inherited


maximlt:
The `constant` Parameter attribute isn't correctly inherited in Parameterized subclasses. This was the behavior in Param 1 and somehow wasn't changed in Param 2.0. Seems like an oversight?

```python
import param

class A(param.Parameterized):
    x = param.Parameter(1, constant=True)

class B(A):
    x = param.Parameter()

print(B.param['x'].constant)
# False

b = B()
b.x = 2
```

A sort of fortunate bug as that would considerably break Panel when it overrides `name` and doesn't set `constant=False` explicitly.

https://github.com/holoviz/panel/blob/d48c45f0606d407c833c02fb85c8fc03c9d9cde4/panel/widgets/base.py#L86

maximlt:
Or mine! Let's blame the test suite :)

PR #1017, opened 2025-02-11
Ensure constant attribute is inherited

maximlt:
Fixes https://github.com/holoviz/param/issues/1001

Let's have a good look at the downstream tests before releasing 2.3.0, this is a nasty bug.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1301,7 +1301,10 @@ class Parameter(_ParameterBase):
         self.precedence = precedence
         self.default = default
         self.doc = doc
-        self.constant = constant is True or readonly is True # readonly => constant
+        if constant is True or readonly is True:  # readonly => constant
+            self.constant = True
+        else:
+            self.constant = constant
         self.readonly = readonly
         self._label = label
         self._set_instantiate(instantiate)
--- a/tests/testparameterizedobject.py
+++ b/tests/testparameterizedobject.py
@@ -1276,7 +1276,7 @@ def test_inheritance_instantiate_behavior():
     assert b.param.p.instantiate is True
 
 
-def test_inheritance_constant_behavior():
+def test_inheritance_readonly_behavior():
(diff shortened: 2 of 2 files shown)


param/parameterized.py
```diff
@@ -1301,7 +1301,10 @@ class hierarchy (see ParameterizedMetaclass).
         self.precedence = precedence
         self.default = default
         self.doc = doc
-        self.constant = constant is True or readonly is True # readonly => constant
+        if constant is True or readonly is True:  # readonly => constant
+            self.constant = True
+        else:
+            self.constant = constant
```
jbednar, line 1307: Ooh; good catch! I wonder if there were any other cases where it wouldn't inherit properly due to computation like this.

jbednar (commented):
Glad to see this one caught! I think that was my oversight orginally. Sorry about that!

param/parameterized.py
maximlt: > I wonder if there were any other cases where it wouldn't inherit properly due to computation like this.

Good question 🙃 !

maximlt:
> Glad to see this one caught! I think that was my oversight orginally. Sorry about that!

Any dependency between Parameter attributes (looking at you `Selector`) makes attribute inheritance more complex. In this case, one could argue there isn't really a need to automatically set `constant=True` when `readonly=True`!

maximlt:
> But then we'd have to wonder what it means for constant=False and readonly=True

I think I could live with that as `readonly` takes precedence over `constant` in my mind. Though it's easy for me as I've internalized what `readonly` and `constant` mean but for most users, their effect isn't obvious, `readonly` meaning `can_never_be_updated` and `constant` meaning `cannot_be_updated_after_instantiation`.

> Maybe it would have been best to handle that on attribute access, not via defaults?

I don't know and don't want to think about it too much :D

Issue #1003, opened 2025-01-03
unwatch should not log a warning when it fails


maximlt:
Reviewing Marc's docstring update work, I saw that `unwatch` apparently logs a warning if it fails at removing a watcher, there has to be a better way, e.g. raising a `LookupError`.

```python
    def unwatch(self_, watcher):
        """Remove the given Watcher object (from `watch` or `watch_values`) from this object's list."""
        try:
            self_._register_watcher('remove', watcher, what=watcher.what)
        except Exception:
            self_.warning(f'No such watcher {str(watcher)} to remove.')
```

jbednar:
I don't know of any reason it wasn't always an exception, but maybe someone else remembers.

PR #1018, opened 2025-02-11
`unwatch` no longer logs a warning and idempotent behavior clarified

maximlt:
Resolves https://github.com/holoviz/param/issues/1003

By raising `ValueError` when trying to remove an already removed/never registered watcher.

### DIFF
--- a/doc/user_guide/Dependencies_and_Watchers.ipynb
+++ b/doc/user_guide/Dependencies_and_Watchers.ipynb
@@ -480,7 +480,7 @@
     "\n",
     "- `obj.param.watch_values(fn, parameter_names, what='value', onlychanged=True, queued=False, precedence=0)`: <br>Easier-to-use version of `obj.param.watch` specific to watching for changes in parameter values. Same as `watch`, but hard-codes `what='value'` and invokes the callback `fn` using keyword arguments _param_name_=_new_value_ rather than with a positional-argument list of `Event` objects.\n",
     "\n",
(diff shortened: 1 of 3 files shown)


jlstevens:
After thinking about this, I think raising a `ValueError` at this stage is too abrupt a change. Instead I propose the following:

1. Improve the warning to use the official warnings module instead of `self_.warning`
2. If we agree the ValueError is a better approach, update this warning to tell users that in future the warning will become an exception.
3. Change it to a `ValueError`.

I believe the original intent was to make unwatching idempotent: if you ask for something to be unwatched and it exits, it gets unwatched. If it is not being watched already, then there is nothing to do.

I see arguments either way and I don't mind this behavior being deprecated, but not this suddenly.

maximlt:
Thanks for the review @jlstevens, beneficial insights. After giving it some more thought, I decided it was best to stay closer to the current implementation that aims to be idempotent. I say "aim" as it is not truly idempotent imo as it logs a warning when the watcher has already been removed. I agree there are situations in an app, even more so when it's async based, where unwatching can happen multiple times or in a non-specific order. `unwatch` being idempotent makes it so users don't have to wrap all their unwatch calls in a try/except.

FWIW what also pushed me to change things was that the try/except was just catching `Exception` which can swallow the wrong error if there's a bug in the code.

With https://github.com/holoviz/param/pull/1018/commits/016b675e1b9b5eb8a8999da011e3c3faeabc0d56:
- `unwatch` no longer raises an error or logs a warning; internally in `_register_watcher` a `ValueError` is caught and swallowed when a 'remove' action fails
- the docs were updated to declare this behavior (I preferred using no-op instead of idempotent)

@jlstevens could you please review again?

Issue #1007, opened 2025-01-14
Deprecate the Parameter attribute pickle_default_value?


maximlt:
`pickle_default_value` is defined as follows:

```
        pickle_default_value: whether the default value should be
        pickled. Usually, you would want the default value to be pickled,
        but there are rare cases where that would not be the case (e.g.
        for file search paths that are specific to a certain system).
```

and set to `False` (`True` by default) in `resolve_path.search_paths` and `resolve_path.path_to_file`. It has been in Param since at least its first commit on Github.

However, Param doesn't make any direct use of it internally, i.e. pickling a Parameterized object will not take into account its value, users have to write custom. The only usage I could find of it on Github is in topographica (https://github.com/search?q=repo%3Aioam%2Ftopographica%20pickle_default_value&type=code). There are no references to this Parameter attribute in Param's own issues or PRs (https://github.com/search?q=repo%3Aholoviz%2Fparam+pickle_default_value&type=issues).

I would suggest deprecating this Parameter attribute.

Issue #898, opened 2024-01-04
Add a `metadata` slot


maximlt:
Libraries like Pydantic, attrs, Traitlets and the built-in dataclasses all offer a way to attach metadata to their field/attributes with some more or less convenient API to retrieve them. `Parameter`s in Param doesn't have a similar slot.

From experience I know I've needed that a few times and I've definitely abused the `precedence` slot to attach some data to Parameters.

Another motivation to add a `metadata` slot is related to Panel. The `Param` Pane offered by Panel is very practical to turn a Parameterized class into a viewable component, Panel maintaining a mapping between Parameter types and its widget types. However, sometimes one needs to configure a widget just a bit more and in this case Panel offers two approaches shown below: (1) pass additional widget configs to `pn.Param` or (2) use `<WidgetClass>.from_param(<parameter>, ...)` to instantiate a widget manually.

![image](https://github.com/holoviz/param/assets/35924738/ca5a0224-f72b-4dfe-9ee9-ad8279d83c17)

```python
import param
import panel as pn

pn.extension()

class User(param.Parameterized):
    fullname = param.String()

user = User()

pn.Param(user.param, widgets={'fullname': {'width': 100}})  # 1

pn.widgets.TextInput.from_param(user.param.fullname, width=100)  # 2
```

I tend to use option (2) as there's often something custom I need to do (e.g. adding some text in between widgets, laying them out in a special way) that isn't made possible by (1), or at least not in a straightforward way. This makes the code quite verbose. Instead Panel could leverage the `metadata` slot to automatically pass widget kwargs collected from a special namespace (e.g. `panel_kwargs`) to `.from_param()` (that I believe is used in the implementation of `pn.Param`):

```python
class User(param.Parameterized):
    fullname = param.String(metadata={'panel_kwargs': {'width': 100}})

user = User()
pn.Param(user.param)
```

This was recently discussed in https://github.com/holoviz/panel/issues/5856.

---

*metadata* attribute in dataclasses, attrs, pydantic and traitlets:

![image](https://github.com/holoviz/param/assets/35924738/5a742a0b-d55d-40ea-a65f-216144619d0b)

<details><summary>Code</summary>
<p>

```python
metadata = {'namespace': {'key1': 'value1'}}

## `dataclasses`

import dataclasses

@dataclasses.dataclass
class User:
    name: str = dataclasses.field(metadata=metadata)

user = User(name='bob')
dataclasses.fields(user)[0].metadata

## `attrs`

import attrs

[...]

PR #1019, opened 2025-02-11
Deprecate the Parameter attribute pickle_default_value

maximlt:
Closes https://github.com/holoviz/param/issues/1007


maximlt:
Draft for now as I'd like:
- https://github.com/holoviz/param/pull/1016 to be merged
- potentially the docstrings MR from @MarcSkovMadsen to be merged to (I'd like to update the docstring of `Parameter` to mark pickle_default_value as deprecated

jbednar:
I think the problem that `pickle_default_value` was meant to address would still apply in 2025, in that pickling a value that is dependent on a particular run on a particular system can cause problems when unpickling, but if it's not been supported for some time, may as well worry about this issue next time it comes up in actual practice.

maximlt:
Yes, I agree that it sounds like it might be useful. But in practice, it seems like it was never implemented in Param, and no one reported setting `pickle_default_value=False` has no effect. So I agree that we can worry about it when this comes up again. In fact, deprecating it might lead someone to come here and ask for that feature to be implemented.

In the meantime, I'd like to push adding a `metadata` slot to Parameters (https://github.com/holoviz/param/issues/898). I think that would allow users to implement their own special Parameter handling without having to create their own Parameter class (I've had the need a few times to label Parameters differently in a class and abused `precedence`).

maximlt:
Unrelated test failures.

PR #1021, opened 2025-02-13
fix: Add dummy `__init__` to ParameterizedFunction

hoxbro:
See https://github.com/holoviz/panel/issues/7690

image

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -4740,6 +4740,9 @@ class ParameterizedFunction(Parameterized):
         else:                 inst.__name__ = self_or_cls.name
         return inst
 
+    def __init__(self, *args, **params):
+        super().__init__(*args, **params)
+
     def __new__(class_,*args,**params):
         # Create and __call__() an instance of this class.
         inst = class_.instance()
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -4741,6 +4741,8 @@ class ParameterizedFunction(Parameterized):
         return inst
 
     def __init__(self, *args, **params):
+        # This is needed so type checkers don't complain about positional arguments
+        # See https://github.com/holoviz/param/pull/1021
         super().__init__(*args, **params)
 
     def __new__(class_,*args,**params):


param/parameterized.py
```diff
@@ -4740,6 +4740,9 @@ def instance(self_or_cls,**params):
         else:                 inst.__name__ = self_or_cls.name
         return inst
 
+    def __init__(self, *args, **params):
+        super().__init__(*args, **params)
```
hoxbro, line 4744:
```suggestion
    def __init__(self, *args, **params):
        # This is needed so type checkers don't complain about positional arguments
        # See https://github.com/holoviz/param/pull/1021
        super().__init__(*args, **params)
```

maximlt:
I don't understand :) Is there some reproducible code I could run to validate the issue and that it's fixed with this change? I haven't seen the error reported on Panel.

hoxbro:
I created this environment: `mamba env create --name "tmp_param" python="3.12" "holoviews pyright" --offline --yes`

``` console
❯ pyright example.py
/home/shh/Downloads/example.py
  /home/shh/Downloads/example.py:3:14 - error: Expected 0 positional arguments (reportCallIssue)
1 error, 0 warnings, 0 informations

~/Downloads via 🅒 tmp_param
❯ cat example.py
import holoviews as hv

hv.extension("bokeh")

```

maximlt:
Ok so here's a reproducer, type checking it with `pyright` shows that error, `mypy` doesn't complain.

```python
class Base:
    def __init__(self, **params):
        pass

class FunctionBase(Base):

    # def __init__(self, *args, **kwargs):
    #     super().__init__(*args, **kwargs)

    def __new__(cls, *args, **kwargs) -> Any:
        inst = Base.__new__(cls)
        result = inst.__call__(*args, **kwargs)
        return result

    def __call__(self, *args, **kwargs):
        raise NotImplementedError

class SomeFunction(FunctionBase):
    def __call__(self, *args, **kwargs):
        return 1

SomeFunction('test')
```

Indeed, adding the dummy `__init__` makes `pyright` happier (and doesn't affect `mypy`). Not super satisfied with a run time solution to a type checking problem, I asked on `pyright`'s forum if there's another solution. Wait a little bit before merging please.

maximlt:
Yes that's true, still weird no? Pyright's developer answered saying that annotating `__new__` with a return type of `Any` (which I tried but saw no change) should work and was buggy at the moment in Pyright but will be fixed in the next release https://github.com/microsoft/pyright/discussions/9907#discussioncomment-12191355.

Issue #869, opened 2023-10-05
Raise an error when setting attributes on a reactive object


maximlt:
When experimenting with the reactive API, I've already done enough the mistake of setting `value` on the reactive object itself instead of using `.rx.value` to ask for this raising an error, e.g.:

```python
namex = rx('bob')
namex.value = 'bill'  # wrong but now error, I wanted to type `namex.rx.value`
```

Actually, I can set any attribute on a reactive object, `namex.foo = 'foo'` doesn't raise an error.

PR #1022, opened 2025-02-14
Raise AttributeError when attempting to set value on an rx obj

maximlt:
Resolves https://github.com/holoviz/param/issues/869

### DIFF
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -1961,6 +1961,11 @@ class rx:
             obj = fn(obj, *resolved_args, **resolved_kwargs)
         return obj
 
+    def __setattr__(self, name, value):
+        if name == "value":
+            raise AttributeError("'rx' has no attribute 'value', try '<reactive_expr>.rx.value = <val>'.")
+        super().__setattr__(name, value)
+
 
 def _rx_transform(obj):
     if not isinstance(obj, rx):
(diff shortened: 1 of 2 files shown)


param/reactive.py
```diff
@@ -1961,6 +1961,11 @@ def _eval_operation(self, obj, operation):
             obj = fn(obj, *resolved_args, **resolved_kwargs)
         return obj
 
+    def __setattr__(self, name, value):
+        if name == "value":
```
hoxbro, line 1965: Why only value?

And what if the underlying object itself has a value attribute?

maximlt: > And what if the underlying object itself has a value attribute?

I don't think this has any consequence? Currently, for an object that has a `value` attribute, `rx(obj).value` returns a reactive expression. Setting `rx(obj).value` sets `value` on the reactive expression, not on the underlying object, which can be done with `rx(obj).rx.value = <value>`. Let me know if I got this wrong.

> Why only value?

Because that's the most error-prone attribute in my opinion and I don't want to restrict `__setattr__` too much beyond that.

hoxbro: > Because that's the most error-prone attribute in my opinion and I don't want to restrict __setattr__ too much beyond that.

A bit opinionated, but I think I agree. Can you add a comment in the code about this?

maximlt: Comment added in https://github.com/holoviz/param/pull/1022/commits/ae47262d6738740f816b59efe8d5062172479892

Issue #661, opened 2022-11-18
Upgrade gmpy support to gmpy2


maximlt:
[gmpy](https://code.google.com/archive/p/gmpy/) is on longer maintained, in favor of [gmpy2](https://github.com/aleaxit/gmpy):

> GMPY and GMPY2 are C-coded Python extension modules that support fast multiple-precision arithmetic.
> 
> GMPY only supports the GMP library and provides fast multiple-precision integer and rational arithmetic. The limited mpf type from GMP is also supported. GMPY is no longer being maintained.
> 
> GMPY2 supports the GMP library for integer and rational arithmetic but GMPY2 adds support for multiple-precision real and complex arithmetic as provided by the MPFR and MPC libraries. GMPY2 is being actively developed.

In https://github.com/holoviz/param/pull/658 (https://github.com/holoviz/param/pull/658/commits/c08ea2640a7e7ca2bb6bb34def1e99e80564e7e7) I attempted to simply replace all the `gmpy` imports by `gmpy2` imports. On Linux there was a failing test and some warnings:

```
   __________ TestTimeDependentDynamic.test_time_hashing_rationals_gmpy2 __________
  
  self = <tests.API1.testtimedependent.TestTimeDependentDynamic testMethod=test_time_hashing_rationals_gmpy2>
  
      @pytest.mark.skipif(gmpy2 is None, reason="gmpy2 is not installed")
      def test_time_hashing_rationals_gmpy2(self):
          """
          Check that hashes of fractions and gmpy2 mpqs match for some
          reasonable rational numbers.
          """
          pi = "3.141592"
          hashfn = numbergen.Hash("test", input_count=1)
          self.assertEqual(hashfn(0.5), hashfn(gmpy2.mpq(0.5)))
  >       self.assertEqual(hashfn(pi), hashfn(gmpy2.mpq(3.141592)))
  E       AssertionError: 238769760 != 243050811
  
  tests/API1/testtimedependent.py:317: AssertionError
  ----------------------------- Captured stderr call -----------------------------
  WARNING:param.main: Casting type 'float' to Fraction.fraction
  WARNING:param.main: Casting type 'mpq' to Fraction.fraction
  WARNING:param.main: Casting type 'str' to Fraction.fraction
  WARNING:param.main: Casting type 'mpq' to Fraction.fraction
  ------------------------------ Captured log call -------------------------------
  WARNING  param.main:parameterized.py:2529 Casting type 'float' to Fraction.fraction
[...]

PR #1026, opened 2025-02-15
Drop gmpy support in favor of gmpy2

maximlt:
Closes https://github.com/holoviz/param/issues/661

https://github.com/holoviz/param/pull/971 added `gmpy2` in the `pixi.toml` file instead of `gmpy`, which are two different projects (the former being maintained while the latter hasn't received an update since 2013). The test suite wasn't updated accordingly so the gmpy-related tests were all skipped (attempting to import `gmpy` instead of `gmpy2`).

This MR lets us "officially" drop support of the `gmpy` Python library (which probably no one used with Param?) in favor of `gmpy2`.

I'm not entirely sure Param should keep supporting that but ensuring the test suite passes with gmpy2 wasn't too much work. I'd be okay if someone pushes to deprecate it.

### DIFF
--- a/numbergen/__init__.py
+++ b/numbergen/__init__.py
@@ -247,7 +247,11 @@ class Hash:
             numer, denom = val, 1
         elif isinstance(val, fractions.Fraction):
             numer, denom = val.numerator, val.denominator
+        elif hasattr(val, 'numerator') and hasattr(val, 'denominator'):
+            # gmpy2 mpq objects have these attributes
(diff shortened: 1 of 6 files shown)


jbednar (approved):
Looks good to me. Supporting gmpy2 is important for anyone using param for a simulator, where exact arithmetic is necessary to avoid subtle bugs, so I'm glad this is being updated.

Issue #651, opened 2022-09-08
concrete_descendents clobbers descendents


sdrobert:
#### ALL software version info
param v1.12.2, Colab notebook

#### Description of expected behavior and the observed behavior

Simply put, classes with the same name get clobbered in get_concrete_descendents().

I came across this bug when trying to run some unit tests with duplicated code (and class definitions) were in different files.

#### Complete, minimal, self-contained example code that reproduces the issue

``` python
import sys
import param

from param.parameterized import descendents

print(param.__version__)

class A(object):
  pass

class B(A):
  pass

b = B

print(b in param.concrete_descendents(A).values())

class B(A):
  pass

print(b in param.concrete_descendents(A).values(), b in descendents(A))
print(B in param.concrete_descendents(A).values(), B in descendents(A))
print(issubclass(b, A), issubclass(b, A))
```

[Colab](https://colab.research.google.com/drive/1k50CV77MmEPxFhTTthxYKTVSlVL1hKeQ?usp=sharing)

Hoxbro:
I can reproduce the problem. And looking at the code, it is easy to see why it happens, as there is no uniqueness to the name of a class. The correct fix would properly be to use the class as key and name as value, though this is a breaking change. 

Maybe the best thing is to test if there are duplicate class names in `param.concrete_descendents` and raise an exception if there are. This could raise exceptions in existing code, but at least it will indicate that something is gone wrong.

https://github.com/holoviz/param/blob/76c7026346b73951dcf4308b6302f0742b0e83e7/param/__init__.py#L1090-L1101

maximlt:
Hm yes the data structure used should probably not have been a dictionary but a list of tuple/namedtuple?

@philippjfr `param.concrete_descendents` is used quite a lot in Panel. What change would you be willing to accept, if any, to fix the reported issue?

Hoxbro:
Deprecating `concrete_descendents` with a warning and having a new function with a better data structure could be a solution. 

It would allow people to shift to the new function and not break anyone using the current implementation.

maximlt:
Yep that's another valid way to approach that 👍

maximlt:
@philippjfr I see plenty of usage of `concrete_descendents` in HoloViews/Panel/Lumen/... and assume you're the number one contributor of this code :) This function has a couple of related issues:

- https://github.com/holoviz/param/issues/519
- https://github.com/holoviz/param/issues/84
- https://github.com/holoviz/param/issues/1013
[...]

PR #1027, opened 2025-02-16
Add concrete keyword to descendents

maximlt:
Somewhat related to https://github.com/holoviz/param/issues/651, the idea being to steer users towards `descendents` instead of `concrete_descendents` (to deprecate it ultimately but it'll take a little while as it's used in a lot of holoviz code bases).

### DIFF
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -518,19 +518,42 @@ def _is_number(obj):
     else: return False
 
 
-def _is_abstract(class_):
+def _is_abstract(class_: type) -> bool:
     try:
         return class_.abstract
     except AttributeError:
         return False
 
 
-def descendents(class_, concrete=False):
+def descendents(class_: type, concrete: bool = False) -> list[type]:
     """
-    Return a list of the class hierarchy below (and including) the given class.
+    Return a list of all descendant classes of a given class.
 
-    The list is ordered from least- to most-specific.  Can be useful for
-    printing the contents of an entire class hierarchy.
+    This function performs a breadth-first traversal of the class hierarchy,
+    collecting all subclasses of `class_`. The result includes `class_` itself
+    and all of its subclasses. If `concrete=True`, abstract base classes
+    are excluded from the result.
+
+    Parameters
+    ----------
+    class_ : type
+        The base class whose descendants should be found.
+    concrete : bool, optional
+        If `True`, exclude abstract classes from the result. Default is `False`.
+
+    Returns
+    -------
+    list of type
+        A list of descendant classes, ordered from the most base to the most derived.
+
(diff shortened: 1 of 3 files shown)


param/_utils.py
```diff
@@ -518,30 +518,54 @@ def _is_number(obj):
     else: return False
 
 
-def _is_abstract(class_):
+def _is_abstract(class_: type) -> bool:
     try:
         return class_.abstract
     except AttributeError:
         return False
```
jbednar, line 525: Is this the best current way to determine if a class is abstract? I would have thought it needed multiple tests to handle the (older) Param way of declaring that and the (newer) Python 3 way of declaring that.

maximlt: This is Param's own way of determining if a class is abstract. I added support to the standard way in https://github.com/holoviz/param/pull/1031, I'd be fine updating `_is_abstract` here first/too.

maximlt: Ok I'll cherry-pick that change from https://github.com/holoviz/param/pull/1031.

maximlt: Added in https://github.com/holoviz/param/pull/1027/commits/bfe9c63abcdf774e6aad049455258726c09eb1c0

jlstevens:
The concept here makes sense to me.

Discussing this with Maxime, we decided there are two paths we can take:

1. Have `concrete_descendents` warn the user when a key is clobbered, directing the user to using `descendents` with `concrete=True`
2. Work the usual deprecation cycle where we eventually get rid of `concrete_descendents` entirely (i.e. start warning users of .`concrete_descendents` to use `descendents` with the keyword instead)

These things are orthogonal and I think the first suggestion is a good idea regardless. I do see the argument for deprecating `concrete_descendents` (i.e. cleaning up the API a bit) , especially given that most usage seems to call `.values()` to get a list-like iterator anyway.

This PR is necessary whether we decide to deprecate `concrete_descendents` or not so I'll go ahead and approve it now.

jlstevens (approved):
Looks good to me!

param/_utils.py
```diff
@@ -518,26 +518,52 @@ def _is_number(obj):
     else: return False
 
 
-def _is_abstract(class_):
+def _is_abstract(class_: type) -> bool:
+    if inspect.isabstract(class_):
```
hoxbro, line 522: (I haven't looked where this is used; if it is not in many places, ignore me).

Is there a penalty related to this check? E.g. if this is in a hotpath will this add significant overhead?

maximlt: Timed it and it's 2-3x slower than the try/except that follows. That's also not a particularly hot path I think.

Issue #84, opened 2015-01-25
Consider replacing our abstract class mechanism with Python's


ceball:
Introduced in python 2.6 (https://docs.python.org/2/whatsnew/2.6.html).

sdc50:
I often create parameterized classes that act as abstract base classes. Having them be "abstract" by convention is fine, but when I tried to officially make them abstract I get a strange error:

```python
import param
from abc import ABC

class Test(ABC, param.Parameterized): pass
```
```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-12-fe93c38dead1> in <module>
----> 1 class Test(ABC, param.Parameterized): pass

TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

While I don't fully understand the error message, I was able to get something working like this:

```python
from abc import ABCMeta

class ParameterizedABCMeta(param.parameterized.ParameterizedMetaclass, metaclass=ABCMeta): pass

class Test(param.Parameterized, metaclass=ParameterizedABCMeta): pass
```

I wonder if it would make sense to include a `ParameterizedABCMeta` class in the param library so that one might do this:

```python
class ParameterizedABC(param.Parameterized, metaclass=param.ParameterizedABCMeta): pass
```

For convenience param could also include a `ParameterizedABC` class:

```python
class ParameterizedABC(param.Parameterized, metaclass=ParameterizedABCMeta): pass
```

So that one could just do:

```python
class Test(param.ParameterizedABC): pass
```

And for completeness maybe these additional classes could also be included:

```python
class ParameterABCMeta(param.parameterized.ParameterMetaclass, metaclass=ABCMeta): pass

class ParameterABC(param.Parameter, metaclass=ParameterABCMeta): pass
```

sdc50:
@jbednar perhaps my comments above should have been in a separate issue since it seems like the original issue is dealing with internal code and my suggestions are dealing with the API. Let me know if you want me to create a new issue. I'd also be happy to create a PR if you think this is a reasonable approach.

jbednar:
@sdc50, does it work to [register your class as an ABC](https://docs.python.org/3/library/abc.html#abc.ABCMeta.register), to avoid having to create those dummy classes?

```
import param
from abc import ABC

@ABC.register
class Test(param.Parameterized): pass
```

PR #1031, opened 2025-02-16
Support declaring a Parameterized class as an ABC

maximlt:
Param has its own way of declaring "abstract" classes by annotating them with `__abstract = True`. Parameterized classes all have an `abstract` property inherited from the metaclass that returns whether this attribute (mangled) was set. The `concrete_descendents` function uses `_is_abstract` to build a collection of the non-abstract descendents only. https://github.com/holoviz/param/issues/84 suggests replacing this approach by [Abstract Bases Classes](https://docs.python.org/3/library/abc.html). This MR doesn't replace the current mechanism (we'd need to deprecate it first) but instead attempts to add support to ABCs, allowing users to declare a Parameterized ABC.

To avoid metaclass conflicts, it introduces the `ParamaterizedABC`class that users must inherit from when they want to declare an ABC.

```python
import abc
import param

class ModelABC(param.parameterized.ParameterizedABC):

    x = param.Number()
    y = param.Number()

    @abc.abstractmethod
    def run(self):
        """This must be implemented by subclasses."""

class BadModel(ModelABC): pass

class GoodModel(ModelABC):
    def run(self):
        return self.x * self.y

try:
    BadModel()
except Exception as e:
    print(repr(e))
    # TypeError("Can't instantiate abstract class BadModel without an implementation for abstract method 'run'")

gm = GoodModel(x=10, y=2)
print(gm.run())
# 20

print(param.concrete_descendents(ModelABC))
# {'GoodModel': <class '__main__.GoodModel'>}
```

---

@sdc50 I know your comment on https://github.com/holoviz/param/issues/84 dates a little (5 years :) ). If you're still interested in this feature, let me know what you think about it. On the HoloViz code bases side, I think we'll need to see whether we can effectively replace `__abstract = True`.

- [ ] Documentation
- [ ] Add more tests

### DIFF
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -519,6 +519,8 @@ def _is_number(obj):
 
 
 def _is_abstract(class_):
+    if inspect.isabstract(class_):
+        return True
     try:
         return class_.abstract
     except AttributeError:
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -8,6 +8,7 @@ either alone (providing basic Parameter support) or with param's
 __init__.py (providing specialized Parameter types).
 """
 
+import abc
 import asyncio
 import copy
 import datetime as dt
@@ -4569,6 +4570,22 @@ class Parameterized(metaclass=ParameterizedMetaclass):
     def __str__(self):
         """Return a short representation of the name and class of this object."""
         return f"<{self.__class__.__name__} {self.name}>"
+
+
+class ParameterizedABCMeta(abc.ABCMeta, ParameterizedMetaclass):
+    """Metaclass for abstract base classes using Parameterized.
+
+    Ensures compatibility between ABCMeta and ParameterizedMetaclass.
+    """
+
+
+class ParameterizedABC(Parameterized, metaclass=ParameterizedABCMeta):
+    """Base class for user-defined ABCs that extends Parameterized."""
+
+    def __init_subclass__(cls, **kwargs):
+        if cls.__bases__ and cls.__bases__[0] is ParameterizedABC:
+            setattr(cls, f'_{cls.__name__}__abstract', True)
+        super().__init_subclass__(**kwargs)
 
 
 def print_all_param_defaults():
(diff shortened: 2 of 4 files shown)


jlstevens (approved):
Looks good to me!

The only thing I want to see before this is merged are some tests making sure the decorators used to declare abstract methods, properties etc. work as expected.

In an ideal world we would run the whole test suite again using `ParameterizedABC` to check everything else works but that seems overkill, especially as this base class is opt-in.

maximlt:
@jlstevens I have added some tests and updated the documentation. I also added `ParameterizedABC` to the top-level module.

doc/user_guide/Parameters.ipynb
```diff
@@ -786,6 +786,191 @@
...
+   "metadata": {},
+   "source": [
+    "## Parameterized Abstract Base Class\n",
+    "\n",
+    "Param supports two ways of declaring a class as an abstract base class (ABC), which is a good approach to define an interface subclasses should implement:\n",
+    "\n",
+    "- Added in version 2.3.0, an abstract Parameterized class can be created by inheriting from `ParameterizedABC`, which is equivalent as inheriting from `ABC` from the Python [abc](https://docs.python.org/3/library/abc.html) module.\n",
+    "- A Parameterized class can be annotated with the class attribute `__abstract` set to `True` to declare it as abstract.\n",
+    "\n",
+    "We recommend adopting the first approach that is more generic and powerful. The second approach is specific to Param (`inspect.isabstract(class)` won't return `True` for example) and preserved for compatibility reasons.\n",
```
jlstevens, line 801: Maybe it is worth adding a follow on sentence to strongly recommend *not* mixing approaches? Having two inconsistent ways of checking if a class is abstract sounds like a recipe for confusion!

That said, maybe param could have a wrapper utility that tries `inspect.isabstract(class)` first and if that returns `False`, checks again the param way?

maximlt: Comment added in https://github.com/holoviz/param/pull/1031/commits/ce27b13a529a98a40b5c5892cae15675e539d92a

> That said, maybe param could have a wrapper utility that tries inspect.isabstract(class) first and if that returns False, checks again the param way?

Internally, we have the `_is_abstract` function that does exactly that. I don't really want to make it public though:
- Parameterized classes have an `abstract` property (that's documented too, I checked) that accounts for both approaches
- The less public API the better
- Ultimately, the goal would be to deprecate and remove the old abstract way

If people feel strongly about it, we can add it in another release anyway.

PR #1035, opened 2025-02-20
Warn when concrete_descendents clobbers classes

maximlt:


jbednar:
I don't know the cleanest way to get out of this mess, but presumably it's also worth (briefly) considering a multidict, which would preserve the interface but allow duplicate keys? That might well cause other problems later, of course, if someone tries to use the multidict as a regular dict.

maximlt:
True, I didn't consider multidict until now. I'll chat with Jean-Luc this week more about all this!

maximlt:
Reading this again https://github.com/holoviz/param/pull/1035#issuecomment-2674486583, I no longer think that's the right approach as I am not confident that there's a sane way to change the type of what `Selector.get_range()` returns.

Instead, I'd suggest either:
1. Status quo for `Selector.get_range()` (e.g. by just swallowing the warnings emitted by the updated `concrete_descendents` or by using an implementation that doesn't emit a warning)
2. Update `concrete_descendents` to no longer clobber the keys but instead append the duplicate ones with an increment (MyClass, MyClass1, MyClass2)

I might just go for 1) now to unblock this.

PR #1038, opened 2025-02-27
perf: Only update Signatures when needed

hoxbro:
e.g. In IPython environment or in docs build (by setting a system environment variable)


pyproject.toml
```diff
@@ -120,7 +120,7 @@ asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope="function"
 
 [tool.coverage.report]
-omit = ["param/version.py"]
+omit = ["param/version.py", "tests/testimports.py"]
```
hoxbro, line 123: This belongs to the other PR, but I don't think moving it over to it is important.

param/parameterized.py
```diff
@@ -981,6 +982,13 @@ def __getattribute__(mcs,name):
             return type.__getattribute__(mcs,name)
 
 
+@lru_cache()
+def _update_signature():
```
hoxbro, line 986: I can see this is the same name as used by the class. Open to better suggestions for naming.

hoxbro: This could be a constant as this will be called during initialization.

pixi.toml
```diff
@@ -90,6 +90,9 @@ nbval = "*"
 [feature.doc]
 channels = ["pyviz"]
 
+[feature.doc.activation.env]
+PARAM_SIGNATURE = "1"
```
hoxbro, line 94: We would need this to be enabled in our other packages, maybe just upstream it directly to nbsite.

hoxbro: So they can also take advantage of the updated signatures in their documentation. 

Could also be that I'm confused.

param/parameterized.py
```diff
@@ -981,6 +982,9 @@ def __getattribute__(mcs,name):
             return type.__getattribute__(mcs,name)
 
 
+_UDPATE_PARAM_SIGNATURE = _in_ipython() or (os.getenv("PARAM_PARAMETER_SIGNATURE", "false").lower() in ("1" , "true"))
```
hoxbro, line 985:
```suggestion
_UDPATE_PARAMETER_SIGNATURE = _in_ipython() or (os.getenv("PARAM_PARAMETER_SIGNATURE", "false").lower() in ("1" , "true"))
```

pixi.toml
maximlt: I can only think of Panel for a library that has custom Parameters, like `Margin`. It doesn't look like it's documented in Panel's API ref. Not worth the trouble of changing anything in these packages/nbsite until the need comes up (the code to dynamically update the signature has only been tested on Param's parameters, I don't know if it works for other Parameters).

maximlt (approved):

Issue #1047, opened 2025-03-27
Incorrect sentence in Roadmap doc


Coderambling:
The first paragraph currently states the below. That "...not expect..." should be "...expect...", correct?

Please confirm and I will update the doc. Also, is the info / status of items in the rest of the page still up to date?

New:

Thus Param users should expect only slow progress on these roadmap items

Current:

Thus Param users should not expect only slow progress on these roadmap items, but they are listed here in the hopes that they will be useful.

https://github.com/holoviz/param/blob/main/doc/roadmap.md

jbednar:
Yes! Would be happy to accept a PR on this.

Coderambling:
Done. Simply removed the word 'not'.

Looks like the roadmap was last updated in 2021. Any updates since then? Happy to update the doc further if you can indicate what has changed / implemented from the current list in the mean time.

Has 1.1. been completed? Also, having 2.1 would be great. I feel it would really improve readability, and make param and Panel code easier to understand for users that are new to the libraries.

Below are the first 3 items currently in the doc:

Currently scheduled plans:

1.1 Remove deprecated API and tag release 2.0. Deprecated API is noted in the source code and except in very rare cases has never been documented on the website, so removing these methods should only affect users of Param from 2020 or earlier.

1.2 More powerful serialization (to JSON, YAML, and URLs) to make it simpler to persist the state of a Parameterized object. Some support already merged as https://github.com/holoviz/param/pull/414 , but still to be further developed as support for using Parameterized objects to build REST APIS (see https://github.com/holoviz/lumen for example usage).

Other items that are not yet scheduled but would be great to have:

2.1 Integrate more fully with Python 3 language features like [type annotations](https://www.python.org/dev/peps/pep-0526) and [data classes](https://docs.python.org/3/library/dataclasses.html), e.g. to respect and validate against declared types without requiring an explicit param.Parameter declaration and potentially without inheriting from param.Parameterized, and to better support IDE type-checking features.

Coderambling:
Thanks!

maximlt:
1.1 is underway and could be removed. But to be honest, I'd be fine removing that page entirely as it doesn't feel like a roadmap but more like a wishlist.

jbednar:
Feel free to rename it to wishlist.md!

maximlt:
I'm fine :) Thanks @Coderambling !

PR #1049, opened 2025-03-28
Update roadmap.md to correct text in first paragraph

Coderambling:
See https://github.com/holoviz/param/issues/1047 for details.


jbednar (approved):
Oops! Good catch.

Issue #1057, opened 2025-05-05
Unexpected behaviour of param.values


esolberg-nr:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc should be added within the dropdown below.)

<details>
    <summary>Python</summary>

```plaintext
3.12.9
```
</details>
<details>
    <summary>param</summary>

```plaintext
2.2.0
```
</details>

#### Description of expected behavior and the observed behavior

Changing the default value of an already created object, changes
what is returned by param.values function. This was surprising to me.

#### Complete, minimal, self-contained example code that reproduces the issue

```python

import param


class TestParams(param.Parameterized):
    b = param.Integer(default=None)


print(param.__version__)  # Prints 2.2.0

p = TestParams()
print(p.param.values()["b"])  # Prints None
p.param.b.default = 1
print(p.param.values()["b"])  # Prints 1 (Would expect None)

```

#### Screenshots or screencasts of the bug in action

- [ ] I may be interested in making a pull request to address this

PR #1058, opened 2025-05-05
Fix `.values()/get_value_generator()` when

maximlt:
Fixes https://github.com/holoviz/param/issues/1057

`.param.values()` uses `.param.get_value_generator()` under the hood (this looks all overly complicated by that's another issue) to obtain the value of a Parameter. `get_value_generator()` seems to be designed to deal specifically with Dynamic Parameters (like `Integer`, `Number`, etc.), to return the generator callable. So in cases where there's no generator, we simply return the value.


philippjfr:
Makes sense.

Issue #376, opened 2020-02-05
Please support static typing


MarcSkovMadsen:
#### My Pain

I'm developing my applications in the VS Code editor. Furthermore I use pylint and mypy to help me create code of high quality.

Param unfortunately does not work that well with that. 

I've tried something like that.

```python
class Scatter2dWithSelectionComponent(param.Parameterized):
    """The Scatter2dWithSelectionComponent enables a user to

    - provide a DataFrame of data
    - View a 2D Scatter plot of the data
    - Select what to show on the x and y axis
    - Select what to show in the tooltips

    Parameters
    ----------
    data : pd.DataFrame
        A dataframe of data to be shown and selected
    """
    data: pd.DataFrame = param.DataFrame(precedence=0.1)
```

But 

- VS Code cannot provide tab completion for `instance.data.`  or help text on hover of `instance.data`.
- Pylint complains for example that `E1101: Instance of 'DataFrame' has no 'iloc' member (no-member)` if I have a line like `instance.data.iloc`.

#### Solution

Support static type checking and enable all the help you can get in modern editors.

jbednar:
We have been focusing on simultaneously supporting Python 2 and Python 3 in the same codebases for the past 5 years, so we haven't been able to make use of any Python3-only syntax like type hinting in our own code. Once we are finally able to leave Python 2 behind, we can become experts in Python 3-only stuff, but for the moment I'm not actually sure what the above type annotation is meant to indicate.  At the class level "data" is _not_ a pd.DataFrame; it's a param.DataFrame object, so I'm not sure that's the correct type annotation to supply. But in the instance it will be a pd.DataFrame, which is what matters for `instance.data.iloc`.  

I really don't know if Python 3's type hints are expressive enough to distinguish between type differences between the class and instance levels.  So I can't say if this is something that is supported now but with a different syntax, if it can be supported but would require changes to the ParameterizedMetaclass, or if it is simply not possible given the type differences between Parameterized classes and instances (which are inherent to Param's design).  Personally, I've found Parameter's type declarations to be vastly more powerful than static types (as they allow specifying ranges and allowed values and not simply type name), but it sounds like the fact that the type is static is something your editor can exploit, and of course the editor doesn't know about Param. Anyway, I wouldn't personally be ready to tackle this issue until we've completely left Python2 behind for good, but if someone else can see a way, be my guest!

MarcSkovMadsen:
Thanks for the discussion. 

[...]

PR #1065, opened 2025-07-11
Add static typing

MarcSkovMadsen:
Closing #376 

This PR adds static typing to param in a way that will not fundamentally change the current implementation and keep the api.

A more radical approach would be to leave descriptor pattern and engage with static typing stakeholders (including mypy) and get param supported in same way as dataclasses, attrs and pydantic.

## Recommendation (from analysis.md)

- Adopt PEP 681’s @dataclass_transform for Param’s metaclass or base class.
- Provide and maintain high-quality stubs for Param descriptors.
- Recommend using class-specific stubs for full static typing, including __init__ signatures.
- Avoid custom plugins unless absolutely necessary.
- Do not recommend wrapper functions with @overload as a general solution for Param static typing.

## Update - Plugin

Ahh. I tried experimenting with a plugin to generate the `__init__` signature - but this was too complicated for me and AI. See https://docs.pydantic.dev/latest/integrations/mypy/#enabling-the-plugin and https://github.com/pydantic/pydantic/blob/main/pydantic/mypy.py.

## Todo

- [x] Define problem (see problem.md)
- [x] Analyze and experiment (see experiment.md and experiment.py. Run with `mypy static-typing/experiment.py --no-incremental`.)
- [x] Propose solution (see analysis.md)
- [ ] Collect feedback
- [ ] Get more contributors onboard. Technically I might be able to succeed. But there will be so many opinions about this that more people need to get deeply involved. We will have to take decissions.
- [ ] Implement, test and review.


param/parameterized.pyi
```diff
@@ -0,0 +1,9 @@
+from typing import Any, Optional, overload
+
+class String:
+    default: Optional[str]
```
maximlt, line 4: How do you handle `allow_None=False`?

MarcSkovMadsen: I don't so far. This just explores the principles via real code. If `allow_None` can be changed dynamically it would be the right type annotation anyways.

static-typing/experiment.py
```diff
@@ -0,0 +1,66 @@
+from typing import Any
```
MarcSkovMadsen, line 1: This is where the value is :-) Showing how `dataclass_transform` can be applied etc.

param/parameterized.pyi
maximlt: Ok. I'd be interested in learning how we might need to adapt Param to make it work better with typing. For instance, yes I think you can modify `allow_None` dynamically? (you can do pretty much you want in Python dynamically anyway). In practice though, I don't think users change it dynamically. But `allow_None` is used quite a lot so it's important to support it, otherwise the hints will be wrong in a lot of times.

MarcSkovMadsen: What I understand is that the problem is that we are using descriptors (classes) and they cannot be overloaded - thus `Optional[str]` is the best type annotation we can provide.

In experiment.py I also shown how you can make overloaded `param_string` functions instead - might solve some issues. But I don't think we would like to go down that route?

maximlt:
@MarcSkovMadsen I'm not really sure what I'm supposed to review here and what the actual proposal is.

param/parameterized.pyi
philippjfr: >But allow_None is used quite a lot so it's important to support it, otherwise the hints will be wrong in a lot of times.

Though not ideal I think it's fine if the type is more permissive than the actual validation, so I'd expect we have to make all types `Optional` to account for `allow_None`.

MarcSkovMadsen:
Review the recommendation

image

Would you accept a PR like that?

Use the .md documents and .py experiment documents to learn about proposal if you need it.

param/parameterized.pyi
MarcSkovMadsen: As far as I can see there is no way around it unless we want to make breaking changes to param and leave the "descriptor" pattern behind.

We could split `String` into `String` and `OptionalString` parameters to follow Pythons type system.

static-typing/problem.md
```diff
@@ -0,0 +1,52 @@
+# Param - Static Typing - Problem
+
+[HoloViz Param](https://param.holoviz.org/) is a powerful library for defining parameters on `param.Parameterized` classes:
```
maximlt, line 3: This is a strange definition of Param?

static-typing/problem.md
```diff
@@ -0,0 +1,52 @@
...
+class MyClass(param.Parameterized):
+    my_parameter = param.String(default="some value", doc="A custom parameter")
+```
+
+## The Core Problem
+
+**Param does not play well with Python's static typing ecosystem.**
+
+Modern Python development expects type annotations to:
+- Enable code completion and inline documentation in editors (like VS Code, PyCharm)
```
maximlt, line 17: Type annotations enable inline documentation?

static-typing/problem.md
```diff
@@ -0,0 +1,52 @@
...
+- Allow static analysis and error checking with tools like mypy and pylint
```
maximlt, line 18: Do you mean pylance? I haven't used pylint in a while but don't really expect it to use type annotations much, at least not as much as mypy/pylance/ty and others.

static-typing/problem.md
```diff
@@ -0,0 +1,52 @@
...
+- Make code easier to understand, refactor, and maintain
+
+But with Param, these benefits are lost:
+
+- **No Automatic Type Inference:** When you declare a parameter (e.g. `my_parameter = param.String(...)`), neither the class nor its instances have a type annotation for `my_parameter`. Editors and type checkers cannot know its type.
+- **Poor Editor Experience:** You don't get tab completion, hover help, or docstrings for parameter attributes on instances. For example, `instance.my_parameter` is invisible to your IDE.
```
maximlt, line 24: I get tab completion in VS Code on the class or on an already created instance.
image

I don't get tab completion when creating an instance, i.e. `P(` doesn't show `test`.

static-typing/problem.md
```diff
@@ -0,0 +1,52 @@
...
+- **Advanced tooling (refactoring, auto-completion, documentation) is much less effective.**
+
+## Example
+
+```python
+class Scatter2dWithSelectionComponent(param.Parameterized):
+    data: pd.DataFrame = param.DataFrame(precedence=0.1)
+```
+
+Even with a type annotation, editors and type checkers do not recognize `data` as a `pd.DataFrame` on the instance. This leads to poor auto-completion and false warnings.
```
maximlt, line 42: That's not what I see.
image
image

static-typing/problem.md
```diff
@@ -0,0 +1,52 @@
...
+
+## What Is Needed
+
+- **Automatic type annotations for parameters.** Parameter attributes should be typed so editors and type checkers know their types.
+- **Constructor signatures that reflect available parameters.**
+- **Better editor integration for auto-completion and documentation.**
+- **No breaking changes unless absolutely necessary.**
+- **Bonus:** If the solution speeds up Param or makes it compatible with dataclasses or Pydantic, that's a huge win.
```
maximlt, line 50: Not sure what it means to be compatible with dataclasses or Pydantic in this context.

static-typing/analysis.md
```diff
@@ -0,0 +1,44 @@
+# Param - Static Typing - Analysis
+
+This document fixes the problem described in [problem.md](problem.md).
+
+## Updated Findings (2025)
+
+### Param Descriptors and Typing
+- Param descriptors (e.g., param.String, param.Number) can provide good attribute type inference in mypy if their stubs define correct __get__ return types (e.g., str, Optional[str], float).
```
maximlt, line 8: Finding a bit curious the usage of `Optional[str]` instead of `str | None` which I thought was the modern way to type that.

static-typing/analysis.md
```diff
@@ -0,0 +1,44 @@
...
+- Type annotations on descriptor assignments cause mypy errors unless suppressed with # type: ignore[assignment].
+- Wrapper functions using @overload (e.g., param_string) can help mypy infer the type of the value returned by the function, but the attribute is still a descriptor, so __init__ signatures are not synthesized.
+- Mypy does not automatically add Param attributes to the __init__ signature, even with @dataclass_transform or wrapper functions. Only dataclasses, attrs, and pydantic are special-cased for this behavior.
```
maximlt, line 11: Searched for `pydantic` in mypy's code base and found no hit https://github.com/search?q=repo%3Apython%2Fmypy%20pydantic&type=code. How does it special-case it?

static-typing/experiment.py
```diff
@@ -0,0 +1,66 @@
...
+from typing_extensions import dataclass_transform, reveal_type
+import param
+from typing import overload, Optional
+
+@dataclass_transform(field_specifiers=(param.String, param.Number))
```
maximlt, line 6: I don't understand. This doesn't seem to work. Is it just there to demonstrate it doesn't work? And if it doesn't work, why? What would it take to make this work?

image

static-typing/experiment.py
```diff
@@ -0,0 +1,66 @@
...
+    my_pyram_string_none = param_string(default=None, doc="A custom string parameter with None as default")
+    my_param_string_no_default = param_string(doc="A custom string parameter with no default")
+    my_param_string_annotated: str = param_string(default="annotated value", doc="A custom annotated string parameter")
+
+    my_test_string = test_string("test", doc="A test string parameter")
+    my_number = param.Number(default=42.0, doc="A number parameter")
+    my_value_annotated: int = 0  # Example of a simple parameter
+
+def test_myclass() -> None:
+    reveal_type(MyClass.my_parameter)  # Should be "builtins.str"
```
maximlt, line 37: My VS code infers `str | None`.

static-typing/experiment.py
```diff
@@ -0,0 +1,66 @@
...
+    
+    reveal_type(MyClass.my_param_string)  # Should be "builtins.str"
+    reveal_type(MyClass.my_pyram_string_none)  # Should be "typing.Optional[str]"
+    reveal_type(MyClass.my_param_string_no_default)  # Should be "typing.Optional[str
+
+    reveal_type(MyClass.my_test_string)  # Should be "builtins.str"
+    reveal_type(MyClass.my_number)     # Should be "builtins.float"
+    reveal_type(MyClass.my_value_annotated)  # Should be "builtins.int"
+    
+    reveal_type(MyClass.__init__)  # Should show correct signature
```
maximlt, line 49: I see no auto-completion options in VS Code. Running `mypy experiment.py` shows:

```
def (self: experiment.MyClass, *generated_args: Any, my_param_string_annotated: builtins.str =, my_value_annotated: builtins.int =, **generated_kwargs: Any)
```

I don't think I would call it the "correct signature"?

static-typing/experiment.py
```diff
@@ -0,0 +1,66 @@
...
+    
+    obj = MyClass(my_parameter="hello", my_number=3.14)
+    
+    reveal_type(obj.my_parameter)  # Should be "builtins.str"
```
maximlt, line 53: VS Code shows `str | None`.

param/parameterized.pyi
maximlt: If made more permissive than in reality I expect a type checker to complain at some point or another, see the code below. How would we let people know that the library provides inexact type hints? I haven't worked so much with type hints so far but know that when type checking is enabled in a code base dealing with wrong types gets quickly pretty frustrating.

```python
def algo(input: str): pass

class P(param.Parameterized):
    value = param.String(allow_None=False)

    def process(self):
        # Wouldn't a type checker be complaining there
        # if the type of `self.value` is `str | None`? Given
        # that `algo` only accepts `str`, I'd expect it to complain.
        return algo(self.value)
```

> We could split String into String and OptionalString parameters to follow Pythons type system.

That's one approach. You can integrate it in your proposal if you want.

param/parameters.pyi
```diff
@@ -0,0 +1,13 @@
+from typing import Any, Optional
+
+class Number:
+    default: float
+    def __get__(self, instance: Any, owner: Any) -> float: ...
+    def __set__(self, instance: Any, value: float) -> None: ...
+    def __init__(self, default: Optional[float] = ..., doc: Optional[str] = ...) -> None: ...
```
maximlt, line 7: Is the `__init__` required? I got more info before in VS Code:

Before:
image

Now:
image

param/parameters.pyi
```diff
@@ -0,0 +1,13 @@
+from typing import Any, Optional
+
+class Number:
+    default: float
```
maximlt, line 4: Why is `default` required here?

static-typing/analysis.md
```diff
@@ -0,0 +1,44 @@
+# Param - Static Typing - Analysis
+
+This document fixes the problem described in [problem.md](problem.md).
+
+## Updated Findings (2025)
+
+### Param Descriptors and Typing
+- Param descriptors (e.g., param.String, param.Number) can provide good attribute type inference in mypy if their stubs define correct __get__ return types (e.g., str, Optional[str], float).
+- Type annotations on descriptor assignments cause mypy errors unless suppressed with # type: ignore[assignment].
+- Wrapper functions using @overload (e.g., param_string) can help mypy infer the type of the value returned by the function, but the attribute is still a descriptor, so __init__ signatures are not synthesized.
```
maximlt, line 10: Why is the `__init__` signature not correct when the attribute is a selector?

maximlt (commented):
Ok so I spent some time reviewing the Markdown files and exploring `experiment.py`. Some points I wanted to raise in addition to the comments:

- I found some inaccuracies in the problem statement in `problem.md`
- I'm not sure what's the difference between `analysis.md` and `experiment.md`, they contain somewhat similar content
- The `dataclass_transform` is used but I haven't found it to work? From the documents, I don't see a description of why it doesn't work, and what it would take for it to work.
- Only `param.String` and `param.Number` are used as examples. I would have expected an analysis of this kind to experiment with a wider variety of Parameters (selectors, mappings, sequences, callable, class selector, etc.), to find potential limitations and suggest approaches how to deal with them.
- Have you looked into how SQLAlchemy implemented their typing support in v2? Some of their requirements look similar to ours https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#orm-declarative-models.

My intuition with making Param work better with type hints is that:
- it would need quite some work
- to make it a good experience some changes in Param would be required (and personally I would be in favor of whatever changes needed, even if it means losing some of the runtime capabilities of Param).

This analysis and the changes made both look too shallow. I also don't see what changes we would need to make to Param so it works well with type hints. Overall, I struggle to see from this analysis what are *our options*? Surely, there are many ways to improve the current situation.

Maybe it is because the goal is to get some more type hints in an IDE? If so, is that a goal we can reach without bringing some frustrations because the type hints we'd provide are not correct?

> Would you accept a PR like that? If yes then I will start implementing.

I don't think *I* would accept a PR like that as this analysis didn't give me the confidence this is a valid approach. Which shouldn't prevent you from exploring further that approach or another one! Also, I see myself as a release manager of Param; @jbednar, @philippjfr and @jlstevens are its official maintainers and can have a different opinion than mine; Philipp seems to be pretty interested in how this work started.

static-typing/problem.md
MarcSkovMadsen: Yes. Tooltips with __init__ signatures and in general more information.

static-typing/problem.md
MarcSkovMadsen: But does it matter?

static-typing/problem.md
MarcSkovMadsen: mypy and pylance use it more. But for example pylint/ ruff have rules for working with type annotations - for example enforcing them.

static-typing/problem.md
MarcSkovMadsen: Its when `data = param.DataFrame(precedence=0.1)`. Then its not recognized.

static-typing/problem.md
MarcSkovMadsen: Can be in many ways. One would be to build on dataclasses or param. Another one would be to be interchangable. A third would be to have the same api. When written it was meant in a broad, yet to be discovered sense. Up to what is possible.

static-typing/analysis.md
MarcSkovMadsen: Yes. But it does not matter for the analysis or conclusion as it is the same.

static-typing/analysis.md
MarcSkovMadsen: Via a plugin

static-typing/experiment.py
MarcSkovMadsen: image

static-typing/experiment.py
MarcSkovMadsen: Yes. But it should optimally be `str`. But that is not possible to get with descriptor pattern. You would need the @overload function for that. But just having `str `None` is an improvement.

static-typing/experiment.py
MarcSkovMadsen: And that is exactly the point. You can't get correct signature unless you either

- annotate attributes or
- provide overloaded function or
- provide plugin

static-typing/experiment.py
MarcSkovMadsen: Yes. It cannot infer more. But ideally it should be `str`. Cannot be achieved without 

- type annotation or 
- overloaded function or
- plugin

param/parameters.pyi
MarcSkovMadsen: Regarding first question. Yes static type checkers would complain that you would have to check for `None`. But now you only have two possibilities `str | None`. To some that is a big help. The ones that do no appreciate types can just let them be.

param/parameters.pyi
MarcSkovMadsen: Good question. I thought it defined the return type of `some_number = param.Number(...)`. But probably its just a type annotation of the `default` attribute should be `float | None` to be correct.

static-typing/analysis.md
MarcSkovMadsen: Sorry. Don't understand the question?

static-typing/problem.md
maximlt: You wrote on Discord that you used AI to work on this PR. I wouldn't have pointed this out otherwise but maybe in that case it matters.

static-typing/problem.md
maximlt: Then this example and the text need to be updated as it clearly states "even with a type annotation, ...":

> Even with a type annotation, editors and type checkers do not recognize `data` as a `pd.DataFrame` on the instance. This leads to poor auto-completion and false warnings.

static-typing/analysis.md
maximlt: I raised this as you wrote on Discord that you used AI to help with this PR. If the AI chose to use `Optional[str]`, then it might be a sign that it is suggesting older options. Python typing progresses very fast, in fact faster than Python releases, as new features can be quickly implemented in type checkers and in `typing-extensions`. So I had to keep that in mind when reviewing the PR, as a possibility for the AI not to be aware of newer features and typing knowledge that we could leverage.

static-typing/analysis.md
maximlt: Which plugin?

static-typing/experiment.py
maximlt: You're pointing at `mypy` but what I show with the screenshot is the output of VS Code. Do I need to configure VS Code differently for this to work?

static-typing/experiment.py
maximlt: I noted this one in particular as most of the other types mentioned in `# Should be ...` were correct, so I thought all should be correct. Maybe update the comment to reflect when the type checker reveals the expected type, and when it doesn't

static-typing/experiment.py
maximlt: Same comment as https://github.com/holoviz/param/pull/1065#discussion_r2203158831

static-typing/experiment.py
maximlt: Same comment as https://github.com/holoviz/param/pull/1065#discussion_r2203158831

param/parameters.pyi
maximlt: I agree that having type annotations would be very useful. But I also think that if we have a library that provides type annotations that are not always correct, then we should provide a solution for people to deal with that. It might be it would just be a documentation thing, stating that these annotations should only be used in an IDE and in a special mode. I don't really know to be honest but I don't find that answer to be complete and clear enough:

> The ones that do no appreciate types can just let them be.

static-typing/analysis.md
maximlt: Sorry I used the wrong word and meant descriptor instead of selector.

Why is the `__init__` signature not correct when the attribute is a descriptor?

param/parameterized.pyi
maximlt: One solution to make a type checker happy here is to add `assert isinstance(self.value, str)` before calling `algo`. Not sure that's a very satisfying solution but that is the sort of thing I'd expect us to have to document if we release a version of Param that doesn't always provide correct type hints.

static-typing/analysis.md
MarcSkovMadsen: The pydantic plugin https://docs.pydantic.dev/2.8/integrations/mypy/

static-typing/experiment.py
MarcSkovMadsen: Are you not referring to reveal_type?

static-typing/analysis.md
maximlt: This page starts with:

> Pydantic works well with [mypy](http://mypy-lang.org/) right out of the box.

The item I commented on states that mypy special-cases pydantic (and dataclasses and attrs) to automatically add the fields to the `__init__` signature. Say I run this code with `mypy t.py`:

```python
# t.py
from pydantic import BaseModel

class Foo(BaseModel):
    x: int

Foo()
```

I get:
```
❯ mypy t.py
t.py:156: error: Missing named argument "x" for "Foo"  [call-arg]
Found 1 error in 1 file (checked 1 source file)
```

I don't think I have the plugin enabled, at least I haven't done what is described in https://docs.pydantic.dev/2.8/integrations/mypy/#enabling-the-plugin. So it looks like mypy doesn't need any special treatment to understand that `Foo` needs to be called with the argument `x`.

It may well be that I'm wrong and if so please show me how so I can better understand the context.

static-typing/experiment.py
maximlt: No, this is the tooltip I get after typing `MyClass(`. Here's a clearer screenshot.

image

PR #1066, opened 2025-07-15
Implement descriptor factory typing system

philippjfr:
Supersedes https://github.com/holoviz/param/pull/1065 and https://github.com/holoviz/param/pull/985

After reviewing the other typing PRs and reading in a bit more depth on what is possible, including traitlets source code (which @maximlt pointed me to), this PR aims to provide an implementation of typing using `Generic`s plus factory style typing overloads that handle polymorphic types that depend on a dynamic value (like `allow_None`).

This approach seems to play well with `pylance` so correctly infers types inside VSCode, without having the user have to manually type things, even if multiple type variants exist for a particular Parameter type. I've not gotten very far down the road of typing each parameter type but at least for the String type this works well, i.e. this works:

```python
import param
from typing import reveal_type

    string = param.String(default='foo')

    string_or_none = param.String(default=None, allow_None=True)

    implicit_string_or_none = param.String(default=None)

t = Test()

reveal_type(t.string) # type: str
reveal_type(t.string_or_none) # type: str | None
reveal_type(t.implicit_string_or_none) # type: str | None
```

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -264,7 +264,8 @@ def resolve_ref(reference: t.Any, recursive: bool = False) -> list[Parameter]:
             if isinstance(arg, str):
                 owner = get_method_owner(reference)
                 if owner is None:
-                    raise
+                    msg = f'Cannot resolve reference {arg!r} on {reference!r} since it is not a method.'
+                    raise ValueError(msg)
                 if arg in owner.param:
                     arg = owner.param[arg]
                 elif '.' in arg:
@@ -923,12 +924,14 @@ def extract_dependencies(function: t.Callable[..., t.Any]) -> list[Parameter]:
         if isinstance(p, str):
             owner = get_method_owner(function)
             if owner is None:
-                raise ValueError()
+                msg = f'Cannot resolve reference {p!r} on {function!r} since it is not a method.'
+                raise ValueError(msg)
             *subps, pname = p.split('.')
             for subp in subps:
                 subowner = getattr(owner, subp, None)
                 if subowner is None:
-                    raise ValueError(f'Cannot depend on undefined sub-parameter {pname!r}.')
+                    msg = f'Cannot depend on undefined sub-parameter {pname!r}.'
+                    raise ValueError(msg)
                 owner = subowner
             if pname in owner.param:
                 pobj = owner.param[pname]


param/parameters.py
```diff
@@ -676,19 +672,55 @@ class Number(Dynamic):
...
+        nested_refs: bool = False,
+    ) -> None:
         ...
 
-    @_deprecate_positional_args
-    def __init__(self, default=Undefined, *, bounds=Undefined, softbounds=Undefined,
-                 inclusive_bounds=Undefined, step=Undefined, set_hook=Undefined, **params):
+    @t.overload
+    def __init__(
+        self: Number[float | int | None, float | int | None],
```
maximlt: IIRC type checkers assume `float` means `float | int` so `float` alone is enough.

jbednar:
Reviewing this, I guess it looks fine. The `if t.TYPE_CHECKING:` bits are gnarly, but if that's what it takes to get type checking, then sure!  

What's the status of this? Are you looking for specific feedback? Does it supersede the previous attempts entirely? What does it take to get it in?

MarcSkovMadsen:
Parameterized __init__ signatures.  That was the starting point for my  PR analysis.

MarcSkovMadsen:
Another source of inspiration is mssgspec. It's quite readable.  They use @dataclass_transform too to get __init__ right. See https://github.com/jcrist/msgspec/blob/main/msgspec/__init__.pyi.

maximlt:
> For example I started on ObjectSelector and am dubious that I can correctly type the value as `Literal[objects]`.

Yep also dubious this will be possible. I'd probably be fine with adding an `item_type` slot to Selector and have that reflected as a typing annotation.

> Parameterized init signatures. That was the starting point for my PR analysis.

I'm fine with that being done in a separate PR. So I guess that means this PR doesn't strictly supersede https://github.com/holoviz/param/pull/1065 (though as far as I could see https://github.com/holoviz/param/pull/1065 was not implementing `@dataclass_transform`).

MarcSkovMadsen:
#1065 is using @dataclass_transform. Its the whole point that we need an approach that supports both attribute types and __init__ signatures. They need to be compatible.

philippjfr:
Yeah also tried #1065 and had no luck getting it to work for `__init__` signatures.

flying-sheep:
what’s the `if t.TYPE_CHECKING` for?

if it’s there in order to keep Sphinx from picking up all the alternative signatures, I think it might be possible to use this: https://github.com/tox-dev/sphinx-autodoc-typehints/blob/e9db2aaee73c6dfe466cde83ea502dc4c33c2fb6/src/sphinx_autodoc_typehints/patches.py#L20-L35

maximlt:
I've tried this PR a bit more and did some testing. Found out that type checkers (mypy, pyright, ty, pyrefly) aren't consistent, even when running against one of the official conformance tests [dataclasses_descriptors.py](https://github.com/python/typing/blob/main/conformance/tests/dataclasses_descriptors.py). It seems like, at the moment, optimizing for one type checker may be at the cost of the experience provided by another one. So, for whatever approach ends up being adopted, it'd be nice to document what kind of experience it provides with a few selected type checkers.

I have this script that I tested with Pyright/Pylance in VSCode. The experience seems decent. Without providing any additional type hint, you already get type inference for attributes. With a bit more work and additional type hints, you can get `__init__` to work quite nicely (displaying the full Parameter definitions in a tooltip).

```python
import typing as t
from param import Parameterized, String, Parameter as P

class MyClass1(Parameterized):
   sa = String()
   sb = String(allow_None=True)

t.assert_type(MyClass1.sa, str)
t.assert_type(MyClass1.sb, str | None)

# No helpful auto complete in VSCode
mc1 = MyClass1()

t.assert_type(mc1.sa, str)
t.assert_type(mc1.sb, str | None)


class MyClass2(Parameterized):
   sa: P[str] = String(doc="Some docs for sb")
   sb: P[str | None] = String(allow_None=True, doc="Some docs for sb")

t.assert_type(MyClass2.sa, str)
t.assert_type(MyClass2.sb, str | None)

MyClass2.x = 10
MyClass2.sa = 10  # Pyright unhappy: good
MyClass2.sa = None  # Pyright unhappy: good
MyClass2.sa = "test"
MyClass2.sb = None

# With auto complete in VSCode
mc2 = MyClass2()
MyClass2(x=10)  # Pyright unhappy: good
MyClass2(sa=10)  # Pyright unhappy: good
MyClass2(sa=None)  # Pyright unhappy: good
MyClass2(sa="test")
MyClass2(sb=None)

t.assert_type(mc2.sa, str)
t.assert_type(mc2.sb, str | None)

class MyClass3(Parameterized):
   sa: str = String(doc="Some docs for sb")  # Pyright unhappy: unfortunate for users who have already added type hints
   sb: str | None = String(allow_None=True, doc="Some docs for sb")  # Pyright unhappy: unfortunate for users who have already added type hints

# With auto complete in VSCode
mc3 = MyClass3()
MyClass3(x=10)  # Pyright unhappy: good
MyClass3(sa=10)  # Pyright unhappy: good
MyClass3(sa=None)  # Pyright unhappy: good

t.assert_type(mc3.sa, str)
t.assert_type(mc3.sb, str | None)
```

Because of the conflicts, breaking down this PR into attribute inference + `__init__` + code base type hints might get too complicated.

param/__init__.py
```diff
@@ -127,7 +128,7 @@
     try:
         # __version__ was added in _version in setuptools-scm 7.0.0, we rely on
         # the hopefully stable version variable.
-        from ._version import version as __version__
```
hoxbro: This is not correct

hoxbro, line 130: I think we should keep and just type ignore it.

philippjfr:
Okay, I've kept 3.10 support and made `typing_extensions` a python3.10 only dependency. Will have to figure that out for conda too.

param/_utils.py
hoxbro: I will underscore these. To avoid them showing up in auto-completion. Just commenting here as this is the first place I saw it, and I know `_utils` is private. 

Image

param/_utils.py
```diff
@@ -1,23 +1,23 @@
...
 import traceback
+import typing as t
 import warnings
 from collections import OrderedDict, abc, defaultdict
 from contextlib import contextmanager
-from numbers import Real
 from textwrap import dedent
 from threading import get_ident
-from typing import TYPE_CHECKING, Callable, TypeVar
+from typing import TYPE_CHECKING, Callable, Protocol, TypeVar, runtime_checkable
```
hoxbro: I think we should either do `from typing import ...` or `import typing as t` not both.

param/_utils.py
```diff
@@ -656,8 +677,12 @@ def _find_stack_level() -> int:
     import numbergen
     import param
 
-    ng_dir = os.path.dirname(numbergen.__file__)
-    param_dir = os.path.dirname(param.__file__)
+    numbergen_file = getattr(numbergen, "__file__", None)
+    param_file = getattr(param, "__file__", None)
+    if numbergen_file is None or param_file is None:
+        return 0
```
hoxbro:
```suggestion
        return 2
```

Don't think this will ever be hit.

param/depends.py
```diff
@@ -66,23 +75,30 @@ def depends(
...
 
     if inspect.isgeneratorfunction(func):
+        func_gen = t.cast(Callable[P, Generator[Any, Any, Any]], func)
         @wraps(func)
-        def _depends(*args, **kw):
-            for val in func(*args, **kw):
+        def _depends_gen(*args, **kw):
+            for val in func_gen(*args, **kw):
                 yield val
+        _depends = t.cast(Callable[P, R], _depends_gen)
```
hoxbro:
```suggestion
        _depends = t.cast("Callable[P, R]", _depends_gen)
```

This should be done for all code. Could also mean we can hide some of the `TypeVars`/imports behind `TYPE_CHECKING`. 

Can be done automatically with `ruff check --select=TYP`, we should likely also add this to ruff rules.

param/depends.py
```diff
@@ -1,46 +1,54 @@
...
+    on_init: bool
+
+class DependsFunc(t.Protocol[P, R]):
+    _dinfo: DependencyInfo
+    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> R: ...
+
+
+@t.overload
 def depends(
     *dependencies: str, watch: bool = ..., on_init: bool = ...
```
hoxbro: Even though `...` is correct. Maybe we should specify them directly to make it easier to see this is how it looks for me in pylance. Hard to see what the default is:

Image

Image

param/display.py
```diff
@@ -1,7 +1,7 @@
 import weakref
 
-_display_accessors = {}
-_reactive_display_objs = weakref.WeakSet()
+_display_accessors: dict[str, object] = {}
+_reactive_display_objs: weakref.WeakSet[object] = weakref.WeakSet()
```
hoxbro, line 4: Why `object` and not `Any` here?

pixi.toml
```diff
@@ -195,3 +200,18 @@ pre-commit = "*"
...
+
+# =============================================
+# =============== TYPECHECK ===================
+# =============================================
+[feature.typecheck.dependencies]
+ty = "*"
+mypy = "*"
+numpy = "*"
+pandas = "*"
+pyrefly = "*"
```
hoxbro, line 212:
```suggestion
ty = "*"
mypy = "*"
numpy = "*"
pandas = "*"
pyrefly = "*"
IPython = "*"
setupstools_scm = "*"
typing_extensions = "*"
```

I think we should add all packages which we depend upon in the codebase.

param/ipython.py
```diff
@@ -21,12 +21,24 @@
...
+_get_ipython: t.Callable[[], t.Any] | None = None
+_display: t.Callable[..., t.Any] | None = None
+try:
+    from IPython import get_ipython as _imported_get_ipython  # type: ignore[unresolved-import]
+    from IPython.display import display as _imported_display  # type: ignore[unresolved-import]
+except Exception:
+    pass
+else:
+    _get_ipython = _imported_get_ipython
+    _display = _imported_display
```
hoxbro:
```suggestion
_get_ipython: t.Callable[[], t.Any] | None = None
_display: t.Callable[..., t.Any] | None = None
try:
    from IPython import get_ipython as _get_ipython  # type: ignore[unresolved-import]
    from IPython.display import display as _display  # type: ignore[unresolved-import]
except ImportError:
    pass
```

Not entirely sure why something like this would not work.

hoxbro: Also this forces IPython import, when importing HoloViews.

This branch:
Image

Main:
Image

`[k for k in sys.modules.keys() if k.startswith("param") or "ipython" in k.lower()]`

param/parameterized.py
```diff
@@ -18,29 +19,22 @@
...
-CRITICAL = 50
-ERROR = 40
-WARNING = 30
-INFO = 20
-DEBUG = 10
-VERBOSE = INFO - 1
+if sys.version_info < (3, 11):
+    from typing_extensions import Self, Unpack
+else:
+    from typing import Self, Unpack
```
hoxbro: We should be able to hide this behind `TYPE_CHECKING`

param/parameterized.py
```diff
@@ -247,7 +262,9 @@ def resolve_ref(reference, recursive=False):
         for arg in (args + kwargs):
             if isinstance(arg, str):
                 owner = get_method_owner(reference)
-                if arg in owner.param:
+                if not isinstance(owner, Parameterized):
```
hoxbro: Couldn't this be a lot of instance checks?

param/parameterized.py
```diff
@@ -441,7 +460,7 @@ def edit_constant(parameterized: 'Parameterized') -> Generator[None, None, None]
             # Some operations trigger a parameter instantiation (copy),
             # we ensure both the class and instance parameters are reset.
             if pname in kls_params and pname not in init_inst_params:
-                type(parameterized).param[pname].constant=True
+                t.cast(t.Any, type(parameterized).param)[pname].constant = True
```
```suggestion
                t.cast("Parameters", type(parameterized).param)[pname].constant = True
```
This will not work?

param/parameterized.py
```diff
@@ -852,8 +878,7 @@ def _resolve_mcs_deps(obj, resolved, dynamic, intermediate=True):
             dependencies.append(dep)
             continue
         inst = obj if dep.inst is None else dep.inst
-        dep = PInfo(inst=inst, cls=dep.cls, name=dep.name,
-                    pobj=inst.param[dep.name], what=dep.what)
+        dep = PInfo(inst, dep.cls, dep.name, inst.param[dep.name], dep.what)
```
hoxbro: I prefer having keywords arguments here.

param/parameterized.py
```diff
@@ -889,13 +914,16 @@ def _skip_event(*events, **kwargs):
...
+def extract_dependencies(function: t.Callable[..., t.Any]) -> list[Parameter]:
     """Extract references from a method or function that declares the references."""
-    subparameters = list(function._dinfo['dependencies'])+list(function._dinfo['kw'].values())
+    dinfo = t.cast(dict[str, t.Any], getattr(function, "_dinfo", {}))
+    subparameters = list(dinfo.get("dependencies", ())) + list(t.cast(dict[str, t.Any], dinfo.get("kw", {})).values())
     params = []
     for p in subparameters:
         if isinstance(p, str):
             owner = get_method_owner(function)
+            if not isinstance(owner, Parameterized):
```
hoxbro: Again could this be a lot of instances check?

param/parameterized.py
```diff
@@ -1344,21 +1411,48 @@ class Parameter(_ParameterBase):
...
         ...
 
-    def __init__( # pylint: disable-msg=R0913
+    def __init__(
         self,
-        default=Undefined,
+        default: t.Any = Undefined,
         *,
-        doc=Undefined,
+        doc=Undefined, # pylint: disable-msg=R0913
```
```suggestion
        doc=Undefined,
```

We don't use pylint afaik.

param/parameterized.py
```diff
@@ -1748,7 +1846,7 @@ def _update_state(self):
         values, after the slot values have been set in the inheritance procedure.
         """
 
-    def __get__(self, obj, objtype): # pylint: disable-msg=W0613
+    def __get__(self, obj: Parameterized | None, objtype: type[Parameterized] | None = None) -> T: # pylint: disable-msg=W0613
```
```suggestion
    def __get__(self, obj: Parameterized | None, objtype: type[Parameterized] | None = None) -> T:
```

param/parameterized.py
```diff
@@ -1708,15 +1805,16 @@ def __setattr__(self, attribute, value):
         if has_watcher and old is not NotImplemented:
             self._trigger_event(attribute, old, value)
 
-    def _trigger_event(self, attribute, old, new):
-        event = Event(what=attribute, name=self.name, obj=None, cls=self.owner,
-                      old=old, new=new, type=None)
+    def _trigger_event(self, attribute: str, old: t.Any, new: t.Any):
+        if self.owner is None:
+            return
```
hoxbro: Should we error here if this happens or is it acceptable?

param/parameterized.py
```diff
@@ -1798,13 +1899,16 @@ def __set__(self, obj, val):
...
         name = self.name
+
         if obj is not None and self.allow_refs and obj._param__private.initialized:
             syncing = name in obj._param__private.syncing
             ref, deps, val, is_async = obj.param._resolve_ref(self, val)
             refs = obj._param__private.refs
             if ref is not None:
-                self.owner.param._update_ref(name, ref)
+                obj.param._update_ref(name, ref)
```
hoxbro, line 1937: Was this a bug?

param/parameterized.py
```diff
@@ -1344,21 +1411,48 @@ class Parameter(_ParameterBase):
...
     def __init__(
         self,
-        default=None, *,
-        doc=None, label=None, precedence=None, instantiate=False, constant=False,
-        readonly=False, pickle_default_value=True, allow_None=False, per_instance=True,
-        allow_refs=False, nested_refs=False, default_factory=None, metadata=None,
-    ):
+        default: t.Any = "",
+        *,
+        doc: str | None = None,
```
hoxbro: Is it because of default values we don't use `Unpack[ParameterKwargs]` here?

param/parameterized.py
```diff
@@ -1798,13 +1899,16 @@ def __set__(self, obj, val):
         object stored in a constant or read-only Parameter (e.g. one
         item in a list).
         """
+        if self.name is None:
+            return
```
hoxbro: Should we error here too?

param/parameterized.py
```diff
@@ -2115,17 +2273,18 @@ class Comparator:
...
 
         for eq_type, eq in equals.items():
-            try:
-                are_instances = isinstance(obj1, eq_type) and isinstance(obj2, eq_type)
-            except TypeError:
-                pass
-            else:
-                if are_instances:
-                    return eq(obj1, obj2)
+            if isinstance(eq_type, type) or (isinstance(eq_type, tuple) and all(isinstance(t, type) for t in eq_type)):
```
hoxbro, line 2300: A lot of instances checks.

param/parameterized.py
```diff
@@ -2287,16 +2447,17 @@ def __dir__(self_):
         list[str]:
             A combined list of standard attributes and parameter names.
         """
-        return super().__dir__() + list(self_._cls_parameters)
+        return list(super().__dir__()) + list(self_._cls_parameters)
 
-    def __iter__(self_):
+    def __iter__(self_) -> Generator[str, None, None]:
```
```suggestion
    def __iter__(self_) -> Iterator[str]:
```

would this work here?

param/parameterized.py
```diff
@@ -2523,16 +2698,28 @@ def _changed(cls, event):
...
         # deepcopy or store a reference to reference param_obj.default into
         # self._param__private.values (or dict_ if supplied) under the
         # parameter's name (or key if supplied)
-        instantiator = copy.deepcopy if deepcopy else lambda o: o
-        self = self_.self
+        if deepcopy:
+            instantiator: Callable[[t.Any], t.Any] = copy.deepcopy
+        else:
+            def instantiator(o: t.Any) -> t.Any:
+                return o
```
```suggestion
        instantiator: Callable[[t.Any], t.Any] = copy.deepcopy if deepcopy else lambda o: o
```
would this work?

param/parameterized.py
```diff
@@ -2600,27 +2794,32 @@ def _resolve_dynamic_deps(self, obj, dynamic_dep, param_dep, attribute):
...
         callback = None
         if depth > 0:
-            def callback(*events):
+            def cb(*events):
                 """
                 If a subobject changes, we need to notify the main
                 object to update the dependencies.
                 """
                 obj.param._update_deps(attribute)
+            callback = cb
```
hoxbro, line 2843:
```
        callback: callable | None = None
        if depth > 0:
            def callback(*events):
                """
                If a subobject changes, we need to notify the main
                object to update the dependencies.
                """
                obj.param._update_deps(attribute)
```

I think this should also work.

param/parameterized.py
```diff
@@ -2600,27 +2794,32 @@ def _resolve_dynamic_deps(self, obj, dynamic_dep, param_dep, attribute):
...
-        for subpath in dynamic_dep.spec.split('.')[:-1]:
+        for subpath in spec_parts[:-1]:
             subobj = getattr(subobj, subpath.split(':')[0], None)
-            subobjs.append(subobj)
+            if subobj is not None:
+                subobjs.append(subobj)
 
         dep_obj = param_dep.cls if param_dep.inst is None else param_dep.inst
-        if dep_obj not in subobjs[:-1]:
+        if dep_obj not in subobjs:
```
hoxbro, line 2829: Was this a bug?

param/parameterized.py
```diff
@@ -2777,36 +2990,46 @@ def update(self_, arg=Undefined, /, **kwargs):
         refs = {}
         if self_.self is not None:
             private = self_.self._param__private
-            params = list(kwargs if arg is Undefined else dict(arg, **kwargs))
-            for pname in params:
+            base: dict[str, t.Any] = {}
+            if not isinstance(arg, UndefinedType):
```
hoxbro: I think this `arg is not  Undefined` was cleaner and likely also more performance.

param/parameterized.py
```diff
@@ -2777,36 +2990,46 @@ def update(self_, arg=Undefined, /, **kwargs):
...
+        **kwargs: t.Any
+    ):
         BATCH_WATCH = self_._BATCH_WATCH
         self_._BATCH_WATCH = True
         self_or_cls = self_.self_or_cls
-        if arg is not Undefined:
-            kwargs = dict(arg, **kwargs)
+
+        base: dict[str, t.Any] = {}
+        if not isinstance(arg, UndefinedType):
```
hoxbro: Same as above. 
```suggestion
        if arg is not UndefinedType:
```

param/parameterized.py
```diff
@@ -3393,7 +3614,7 @@ def values(self_, onlychanged: bool = False) -> dict[str, Any]:
     # Please update the docstring with better description and examples
     # I've (MarcSkovMadsen) not been able to understand this. Its probably because I lack context.
     # Its not mentioned in the documentation or pytests
-    def force_new_dynamic_value(self_, name): # pylint: disable-msg=E0213
+    def force_new_dynamic_value(self_, name: str): # pylint: disable-msg=E0213
```
```suggestion
    def force_new_dynamic_value(self_, name: str):
```

param/parameterized.py
```diff
@@ -3581,8 +3805,7 @@ def method_dependencies(self_, name: str, intermediate: bool = False) -> list[PI
         [PInfo(inst=MyClass(a=None, b=None, name='MyClass...]
         """
         method = getattr(self_.self_or_cls, name)
-        minfo = MInfo(cls=self_.cls, inst=self_.self, name=name,
-                      method=method)
+        minfo = MInfo(self_.self, self_.cls, name, method)
```
hoxbro, line 3829: I still like keyword arguments for these. So easy to give wrong order here.

param/parameterized.py
```diff
@@ -3719,24 +3943,28 @@ def _spec_to_obj(self_, spec, dynamic=True, intermediate=True):
                         deps += subdeps
                 return deps, [] if intermediate == 'only' else [DInfo(spec=spec)]
 
+        if not (isinstance(src, type) and issubclass(src, Parameterized) or isinstance(src, Parameterized)):
+            raise ValueError(f"Dependency {spec!r} is not a Parameterized class or instance.")
         cls, inst = (src, None) if isinstance(src, type) else (type(src), src)
```
hoxbro, line 3969:
```suggestion

        is_type = isinstance(src, type)
        if not (is_type and issubclass(src, Parameterized) or isinstance(src, Parameterized)):
            raise ValueError(f"Dependency {spec!r} is not a Parameterized class or instance.")
        cls, inst = (src, None) if is_type else (type(src), src)
```

Not sure how hot path this is, but would there ever be a case where src is not Parameterized or an instance of Parameterized.

param/parameterized.py
```diff
@@ -4339,9 +4594,7 @@ def _pprint(self, imports=None, prefix=" ", unknown_value='<?>',
             if k in posargs:
                 # value will be unknown_value unless k is a parameter
                 arglist.append(value)
-            elif (k in kwargs or
-                  (hasattr(spec, 'varkw') and (spec.varkw is not None)) or
-                  (hasattr(spec, 'keywords') and (spec.keywords is not None))):
+            elif k in kwargs or (spec.varkw is not None):
```
hoxbro, line 4618: Nice

param/parameterized.py
```diff
@@ -4979,9 +5252,9 @@ def container_script_repr(container,imports,prefix,settings):
 def _no_script_repr():
     # Suppress scriptrepr for objects not yet having a useful string representation
     if random := sys.modules.get("random"):
-        yield random.Random
+        yield t.cast(t.Any, random).Random
```
``` python
if "random" in sys.modules:
    import random
    yield random.Random
```

A bit cleaner IMO.

param/parameterized.py
```diff
@@ -5244,7 +5537,7 @@ def __init__(
         self.ref_watchers = []
         self.async_refs = {}
         self.parameters_state = parameters_state
-        self.dynamic_watchers = defaultdict(list) if dynamic_watchers is None else dynamic_watchers
+        self.dynamic_watchers = defaultdict(list) if dynamic_watchers is None else defaultdict(list, dynamic_watchers)
```
```suggestion
        self.dynamic_watchers = defaultdict(list, dynamic_watchers or ())
```

i think this should work.

param/parameterized.py
```diff
@@ -5258,6 +5551,53 @@ def __setstate__(self, state):
             setattr(self, k, v)
 
 
+C = t.TypeVar("C", bound="Parameterized")
+
+
+class NS:
```
hoxbro: These twos should likey be `_NS` and `_PrivateNS`.

param/parameterized.py
```diff
@@ -5907,7 +6245,7 @@ def instance(self_or_cls,**params):
         >>> instance(5)
         10
         """
-        if isinstance (self_or_cls,ParameterizedMetaclass):
+        if isinstance(self_or_cls, type):
```
hoxbro: Not sure this is correct.

param/parameters.py
```diff
@@ -15,27 +15,34 @@
...
 from collections import OrderedDict
-from collections.abc import Iterable
+from collections.abc import Iterable, Mapping, Sequence
 from contextlib import contextmanager
+from os import PathLike
+
+if sys.version_info < (3, 11):
+    from typing_extensions import Unpack
+else:
+    from typing import Unpack
```
hoxbro: I think we should have these behind `TYPE_CHECKING`

param/parameters.py
```diff
@@ -1271,10 +1900,12 @@ def _validate_step(self, val, step):
     def _validate_order(self, val, step, allow_None):
         if val is None and allow_None:
             return
+        if val is None:
+            return
```
hoxbro: This looks wrong to me. We should either remove one or the if statement or raise a ValueError.

param/parameters.py
```diff
@@ -2370,7 +3127,7 @@ def serialize(cls, value):
     def deserialize(cls, value):
         if value == 'null' or value is None:
             return None
-        import pandas
+        pandas = t.cast(t.Any, importlib.import_module('pandas'))
```
hoxbro:
```suggestion
        import pandas
```

Same with other places, this should work.

param/parameters.py
```diff
@@ -2598,11 +3409,11 @@ class resolve_path(ParameterizedFunction):
     than just os.getcwd() can be used, and the file must exist.
     """
 
-    search_paths = List(default=[os.getcwd()], pickle_default_value=None, doc="""
+    search_paths = t.cast(t.Any, List)(default=[os.getcwd()], pickle_default_value=None, doc="""
         Prepended to a non-relative path, in order, until a file is
         found.""")
 
-    path_to_file = Boolean(default=True, pickle_default_value=None,
+    path_to_file = Boolean(default=True, pickle_default_value=t.cast(bool, None),
```
hoxbro: This looks weird.

param/parameters.py
```diff
@@ -2598,11 +3409,11 @@ class resolve_path(ParameterizedFunction):
     than just os.getcwd() can be used, and the file must exist.
     """
 
-    search_paths = List(default=[os.getcwd()], pickle_default_value=None, doc="""
+    search_paths = t.cast(t.Any, List)(default=[os.getcwd()], pickle_default_value=None, doc="""
```
hoxbro: This looks weird.

param/reactive.py
```diff
@@ -1822,7 +1845,7 @@ def __array_ufunc__(self, ufunc, method, *args, **kwargs):
         }
         return new._clone(operation)
 
-    def _apply_operator(self, operator, *args, reverse=False, **kwargs) -> 'rx':
+    def _apply_operator(self, operator, *args, reverse=False, **kwargs) -> t.Any:
```
hoxbro: The change of output type likely break what was originally intended.

pixi.toml
```diff
@@ -195,3 +200,18 @@ pre-commit = "*"
...
+
+[feature.typecheck.tasks]
+ty-typecheck = 'ty check param --exclude param/version.py --ignore unused-type-ignore-comment'
+mypy-typecheck = 'mypy param --follow-imports=skip --disable-error-code import-untyped --disable-error-code import-not-found'
+pyrefly-typecheck = 'pyrefly check --output-format min-text --count-errors=0'
```
hoxbro: At least for ty and pyrefly we should likely set these in `pyproject.toml` so their respective LSP can see the configurations.

pyproject.toml
```diff
@@ -149,3 +151,18 @@ ignore = [
...
 [tool.ruff.lint.pydocstyle]
 convention = "numpy"
+
+[tool.pyrefly]
+project-includes = ["param/"]
+project-excludes = [
+    "param/version.py",
+    "**/version.py",
+    "**/tests",
+    "**/test_*",
```
hoxbro, line 175: A good way to test if type checking works is to take a look at how it lint the tests.

param/_utils.py
philippjfr: Yep, seem to have missed _utils

param/parameterized.py
philippjfr: Type checkers prefer positional args for namedtuple. Would need an ignore otherwise.

param/parameters.py
philippjfr: Fixed

param/parameters.py
philippjfr: Fixed

param/reactive.py
philippjfr: Fixed

param/display.py
philippjfr: `object` typing here effectively says:

> “This holds arbitrary objects, and you are not allowed to assume anything about them.”

while `Any` would allow any operation on the values.

param/_utils.py
philippjfr: I think underscoring things in public modules make sense, but if a module is already private it seems redundant to then underscore things in the private namespace, but I guess it'd be consistent with what's already done here.

param/parameterized.py
philippjfr: Honestly don't think it can happen.

param/parameterized.py
philippjfr: No, `obj` and `self.owner` should be interchangeable here I think but seems cleaner for consistency in this case.

param/parameterized.py
hoxbro: Have pushed [75d99cd](https://github.com/holoviz/param/pull/1066/commits/75d99cd3cdd0f7402b6e7253656e8de8c7d11e6d) which changes Undefined to an Enum, which type checkers understand.

param/parameterized.py
hoxbro: Have pushed [75d99cd](https://github.com/holoviz/param/pull/1066/commits/75d99cd3cdd0f7402b6e7253656e8de8c7d11e6d) which changes Undefined to an Enum, which type checkers understand.

param/parameterized.py
philippjfr: Resolved by using a cast instead.

param/parameterized.py
philippjfr: Done

param/parameterized.py
philippjfr: Removed all pyling ignores.

param/parameterized.py
philippjfr: Reverting this

philippjfr: Nevermind, yes, this was a bug and we added tests to fix it.

param/parameterized.py
philippjfr: Both are fine, but I'll revert.

param/parameterized.py
philippjfr: Doing this.

param/parameterized.py
philippjfr: Yes, I have now expanded the first signature everywhere so the VSCode UI shows the default args..

param/parameterized.py
philippjfr: Will fix with a cast.

param/parameterized.py
philippjfr: Not a hot path and not usually a lot of items here.

param/parameterized.py
```diff
@@ -925,7 +925,7 @@ def extract_dependencies(function: t.Callable[..., t.Any]) -> list[Parameter]:
             owner = t.cast("Parameterized", get_method_owner(function))
             *subps, p = p.split('.')
             for subp in subps:
-                owner = getattr(owner, subp, None)
+                owner = t.cast("Parameterized", getattr(owner, subp, None))
```
hoxbro: Something in this function looks wrong to me...

param/reactive.py
```diff
@@ -109,12 +109,7 @@
 from ._utils import _to_async_gen, iscoroutinefunction, full_groupby
 
 if t.TYPE_CHECKING:
-    import sys
-
-    if sys.version_info < (3, 11):
-        from typing_extensions import Self
-    else:
-        from typing import Self
+    from typing_extensions import Self
```
philippjfr, line 115: Why the change? The point was to not add dependencies.

hoxbro: Have also removed it as a dependency. It is behind a TYPE_CHECKING blog, so no need to have it as a runtime dependency.

philippjfr: But it becomes a downstream dependency if someone wants to get type checking during development, no?

param/parameterized.py
philippjfr: Why?

param/reactive.py
philippjfr: Looks like all type checkers depend on it in some form, so it's not an issue.

param/parameterized.py
```diff
@@ -2062,7 +2059,7 @@ def _set_names(self, attrib_name: str):
                 'ensure that you create a new parameter '
                 'instance for each new class.'
             )
-        self.name = t.cast("t.Any", attrib_name)
+        self.name = attrib_name  # TODO: Is the casting need to support other names?
```
philippjfr: Don't think the cast was needed.

hoxbro:
```suggestion
        self.name = attrib_name
```

Didn't think so too, but just wanted to be sure that it was not something explicitly added.

param/parameterized.py
hoxbro: From a type perspective.

philippjfr: It's fine I think, the assumption is that dotted depends specs, i.e. `@param.depends('subobj.subsubobj.parameter')` always point to Parameterized sub-objects (or None).

philippjfr: Guess, the cast should be updated to be Parameterized | None.

PR #1079, opened 2025-10-23
Version the docs

maximlt:


jbednar (approved):
Thanks. Looks good to me. Usual concern about disruption to search engine results and old links applies!

PR #1089, opened 2025-10-29
Do not override refs when parameter is triggered

philippjfr:
When a parameter value is updated any refs that are tied to the parameter are removed. This makes sense when a user actually sets an external value since it causes a de-sync of the reference and the parameter, however when merely calling `.param.trigger(<param_name>)` there's no reason to de-sync them.

- [x] Fixes https://github.com/holoviz/param/issues/1076
- [x] Add test

### DIFF
--- a/tests/testrefs.py
+++ b/tests/testrefs.py
@@ -96,7 +96,8 @@ def test_parameter_ref():
     p.string = 'new_string'
     assert p2.string == 'new_string'
 
-def testa_param_ref_no_desync_on_trigger():
+def test_param_ref_no_desync_on_trigger():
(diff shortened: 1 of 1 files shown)


tests/testrefs.py
```diff
@@ -96,6 +96,18 @@ def test_parameter_ref():
     p.string = 'new_string'
     assert p2.string == 'new_string'
 
+def testa_param_ref_no_desync_on_trigger():
+    p = Parameters(string='foo')
```
maximlt:
```suggestion
def test_param_ref_no_desync_on_trigger():
    # See https://github.com/holoviz/param/issues/1076
    p = Parameters(string='foo')
```

PR #1092, opened 2025-10-30
Add a default_factory attribute to Parameter

maximlt:
Resolves https://github.com/holoviz/param/issues/508

In the simple and likely most common case, `default_factory` is set with a callable that is called internally on instance creation without any argument. This is useful for example to set a UUID on each new instance, or a created_at field:

```python
import param
import datetime
import functools
import uuid

def uuid4():
    return str(uuid.uuid4())

class Object(param.Parameterized):
    uuid = param.String(default_factory=uuid4)
    created_at = param.Date(default_factory=functools.partial(datetime.datetime.now, tz=datetime.UTC))

obj1 = Object()
print(obj1)
# Object(created_at=datetime.datetime(2025, 10, 30, 10, 33, 16, 470076, tzinfo=datetime.timezone.utc), name='Object00002', uuid='a466d2e9-5dca-4729-af5b-17cbcab49823')
```

My original motivation for `default_factory` was to find a way to re-implement the behavior of the `name` parameter all Parameterized classes have. To that end, the factory callable can be wrapped in a `DefaultFactory` instance, which will let Param know that it needs to call it with the three arguments `cls`, `self`, and `parameter`. Instantiating `DefaultFactory` with `on_class=True` enables the factory to also be called on class creation (in which case `self` is passed None).

```python
debug_id_counter = 0

def debug_id(cls, self, parameter):
    global debug_id_counter
    if self:
        # When the class value is overriden.
        if getattr(cls, parameter.name) != cls.__name__:
            return getattr(cls, parameter.name)
        else:
            name = f'{cls.__name__}{debug_id_counter:05d}'
            debug_id_counter += 1
            return name
    else:
        return cls.__name__

class Object(Parameterized):
    debug_id = param.String(
        default_factory=param.parameterized.DefaultFactory(
            debug_id,
            on_class=True,
        )
    )

class SubObject(Object):
    pass
```
image

On instance creation, the factory is called after the instance has been internally marked as initialized. I made that choice as:
- The parameter (obtained directly from the `parameter` argument, or via `self.param[<pname>]`) is an instance-level parameter, and not class-level (only class-level parameters can be obtained before the class is initialized), which is less confusing and error-prone, I think
- The factory has access to the values of all the other parameters

Reviewing `testdefaultfactory.py` is recommended to understand how this all works.


doc/user_guide/Parameters.ipynb
```diff
@@ -786,6 +787,148 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "c9af5e76-29c5-4a78-a2c3-07fef892c832",
+   "metadata": {},
+   "source": [
+    "## Default Factory\n",
+    "\n",
+    "It is sometimes useful to dynamically set the default value of a Parameter, e.g. based on the system state, or just randomly. A typical example would be assigning a unique identifier to each new instance of a class. The `default_factory` keyword and attribute (added in version 2.3.0) can be set to a factory callable to dynamically compute the parameter value.\n",
+    "\n",
+    "The callable can be any callable that can be executed without positional arguments (e.g. `def foo(): ...`, `def foo(*args): ...`, `def foo(**kwargs): ...`, etc.). In addition, the callable can be wrapped as an instance of `param.parameterized.DefaultFactory`, in which case it is called with the three arguments `cls`, `self`, and `parameter`:\n",
```
philippjfr:
```suggestion
    "The `default_factory` can be any callable that can be invoked without requiring positional arguments (e.g. `def foo(): ...`, `def foo(*args): ...`, `def foo(**kwargs): ...`, etc.). In addition, the callable can be wrapped as an instance of `param.parameterized.DefaultFactory`, in which case it is called with the three arguments `cls`, `self`, and `parameter`:\n",
```

doc/user_guide/Parameters.ipynb
```diff
@@ -786,6 +787,148 @@
...
+    "\n",
+    "- On instance creation, the callable is passed the Parameterized class to `cls`, the Parameterized instance to `self`, and the instance-level Parameter object to `parameter`.\n",
+    "- On class creation (typically when importing a module that defines Parameterized classes), and only if `DefaultFactory` is created with `on_class=True` (default is False), the callable is passed the Parameterized class to `cls`, `None` to `self` (as the instance doesn't exist yet), and the class-level Parameter object to `parameter`. This case is the only one that allows to control the *class-level* attribute value.\n",
```
philippjfr:
```suggestion
    "- During class creation (typically when a module defining `Parameterized` classes is imported), and only if the `DefaultFactory` is initialized with `on_class=True` (default is `False`), the callable is instead passed the `Parameterized` class as `cls`, `None` for `self` (since no instance exists yet), and the class-level `Parameter` object as `parameter`. This is the only case where the callable can influence the *class-level* attribute value.\n",
```

* When an instance is created, the callable receives the `Parameterized` class as `cls`, the instance itself as `self`, and the instance-level `Parameter` object as `parameter`.
*

doc/user_guide/Parameters.ipynb
```diff
@@ -786,6 +787,148 @@
...
+   "id": "c9af5e76-29c5-4a78-a2c3-07fef892c832",
+   "metadata": {},
+   "source": [
+    "## Default Factory\n",
+    "\n",
+    "It is sometimes useful to dynamically set the default value of a Parameter, e.g. based on the system state, or just randomly. A typical example would be assigning a unique identifier to each new instance of a class. The `default_factory` keyword and attribute (added in version 2.3.0) can be set to a factory callable to dynamically compute the parameter value.\n",
+    "\n",
+    "The callable can be any callable that can be executed without positional arguments (e.g. `def foo(): ...`, `def foo(*args): ...`, `def foo(**kwargs): ...`, etc.). In addition, the callable can be wrapped as an instance of `param.parameterized.DefaultFactory`, in which case it is called with the three arguments `cls`, `self`, and `parameter`:\n",
+    "\n",
+    "- On instance creation, the callable is passed the Parameterized class to `cls`, the Parameterized instance to `self`, and the instance-level Parameter object to `parameter`.\n",
```
philippjfr:
```suggestion
    "- When an instance is created, the callable receives the `Parameterized` class as `cls`, the instance itself as `self`, and the instance-level `Parameter` object as `parameter.\n",
```

doc/user_guide/Parameters.ipynb
```diff
@@ -786,6 +787,148 @@
...
+    "- On class creation (typically when importing a module that defines Parameterized classes), and only if `DefaultFactory` is created with `on_class=True` (default is False), the callable is passed the Parameterized class to `cls`, `None` to `self` (as the instance doesn't exist yet), and the class-level Parameter object to `parameter`. This case is the only one that allows to control the *class-level* attribute value.\n",
+    "\n",
+    ":::{important}\n",
+    "Parameters all have a `default` value. When `default_factory` is set, it *takes precedence over* `default`. For example, updating a Parameter's attribute at the class level has no effect on the current and future instances of this class if the Parameter underlying the attribute has a `default_factory`.\n",
+    ":::\n",
+    "\n",
+    "Let's see how it works. We'll start with the most common use-case: simple factory functions that generate a parameter value used at the instance level. In this example, every new instance of the `Object` class will have its `uuid` and `created_at` attributes dynamically computed on creation."
```
philippjfr:
```suggestion
    "Let's see how this works. We'll start with the most common use-case: simple factory functions that generate a parameter value used at the instance level. In this example, every new instance of the `Object` class will have its `uuid` and `created_at` attributes dynamically computed on instantiation."
```

param/parameterized.py
```diff
@@ -1376,6 +1377,16 @@ def __init__( # pylint: disable-msg=R0913
...
             parameter, which can be overridden in an instance.
             Default is ``None``.
+        default_factory : callable, optional
+            A callable to generate the attribute value. The callable can either
+            be passed directly (in which case it is called with 0 arguments),
+            or can be wrapped in a :class:`DefaultFactory` for advanced use cases
+            (in which case the factory is called with the arguments``cls``,
+            ``self`` and ``parameter``). ``default_factory`` takes precedence
+            over ``default`` when set. On instance creation, the factory is
+            called once the :class:`Parameterized` instance is initialized.
```
philippjfr:
```suggestion
            A callable used to generate the attribute value. It can be provided
            directly (in which case it is invoked with no arguments) or wrapped
            in a :class:`DefaultFactory` for more advanced behavior (in which
            case it receives ``cls``, ``self``, and ``parameter`` as arguments).
            When specified, ``default_factory`` takes precedence over ``default``.
            The factory is invoked once the :class:`Parameterized` instance has
            is initialized.
```

philippjfr:
My vote is to stick with `default_factory`. I think it's clearer than the alternative.

jbednar:
I'm fine with `default_factory`, since one could imagine having factory functions for other things than the default value.

doc/user_guide/Parameters.ipynb
```diff
@@ -28,6 +28,7 @@
     "Each Parameter type can define additional behavior and associated metadata, but the metadata supported for all Parameter types includes:\n",
     "\n",
     "- **default**: Default value for this parameter at the class level, which will also be the value at the Parameterized instance level if it hasn't been set separately on the instance.\n",
+    "- **default_factory**: Optional callable to generate the attribute value. The callable can either be passed directly (in which case it is called with 0 arguments), or can be wrapped in a `DefaultFactory` for advanced use cases (in which case the factory is called with the arguments `cls`, `self` and `parameter`). `default_factory` takes precedence over `default` when set. On instance creation, the factory is called once the Parameterized instance is initialized.\n",
```
jbednar: "the attribute value" seems confusing here, as it doesn't have any apparent linkage to the `default` metadata item defined just previously. Maybe:

```suggestion
    "- **default_factory**: Optional callable to generate the `default` value for this parameter when an instance of this class is created. The callable can either be passed directly (in which case it is called with 0 arguments), or can be wrapped in a `DefaultFactory` for advanced use cases (in which case the factory is called with the arguments `cls`, `self` and `parameter`). `default_factory` takes precedence over `default` when set. On instance creation, the factory is called once the Parameterized instance is initialized.\n",
```

doc/user_guide/Parameters.ipynb
```diff
@@ -786,6 +787,148 @@
...
+    "\n",
+    "The `default_factory` can be any callable that can be invoked without requiring positional arguments (e.g. `def foo(): ...`, `def foo(*args): ...`, `def foo(**kwargs): ...`, etc.). In addition, the callable can be wrapped as an instance of `param.parameterized.DefaultFactory`, in which case it is called with the three arguments `cls`, `self`, and `parameter`:\n",
+    "\n",
+    "- When an instance is created, the callable receives the `Parameterized` class as `cls`, the instance itself as `self`, and the instance-level `Parameter` object as `parameter`.\n",
+    "- During class creation (typically when a module defining `Parameterized` classes is imported), and only if the `DefaultFactory` is initialized with `on_class=True` (default is `False`), the callable is instead passed the `Parameterized` class as `cls`, `None` for `self` (since no instance exists yet), and the class-level `Parameter` object as `parameter`. This is the only case where the callable can influence the *class-level* attribute value.\n",
```
jbednar: Can you explicitly state what "This" is?

doc/user_guide/Parameters.ipynb
```diff
@@ -786,6 +787,148 @@
...
+    "\n",
+    ":::{important}\n",
+    "Parameters all have a `default` value. When `default_factory` is set, it *takes precedence over* `default`. For example, updating a Parameter's attribute at the class level has no effect on the current and future instances of this class if the Parameter underlying the attribute has a `default_factory`.\n",
```
jbednar:
```suggestion
    "Parameters all have a `default` value. When `default_factory` is set, it *takes precedence over* `default` as the source of the default value for a new instance. For example, updating a Parameter's attribute at the class level has no effect on the current and future instances of this class if the Parameter underlying the attribute has a `default_factory`.\n",
```

doc/user_guide/Parameters.ipynb
maximlt: That's not quite correct as the factory is not called when the Parameter instance is created, but when the Parameterized instance is created.

doc/user_guide/Parameters.ipynb
maximlt: Done in https://github.com/holoviz/param/pull/1092/commits/9a31616974549b451e47b57cc120c009d3c6f0e6

doc/user_guide/Parameters.ipynb
maximlt: Done in https://github.com/holoviz/param/pull/1092/commits/96fa2cd5679812fc9d77ce7dc2f78cf2f6bebad4

jbednar: Oops; that's what I meant!

PR #1094, opened 2025-11-06
Add the metadata Parameter attribute

maximlt:
Resolves https://github.com/holoviz/param/issues/898

`metadata` is None by default. At some point, I thought about using `{}`, which would be safe as Param automatically makes a copy of Parameter attributes with a value that is a mutable container (see https://github.com/holoviz/param/pull/826), but then I thought that this wasn't bringing much and wasn't worth the overhead. Note that attrs/dataclasses make a shallow copy of the dict and wrap it in a [`types.MappingProxyType`](https://docs.python.org/3/library/types.html#types.MappingProxyType) to make it readonly. I tend not to think this is needed in Param.

The docstring and docs say that `metadata` is a mapping. This PR doesn't implement a run-time check for that, which is aligned with the other attributes.

### DIFF
--- a/doc/user_guide/Parameters.ipynb
+++ b/doc/user_guide/Parameters.ipynb
@@ -1009,7 +1009,7 @@
    "id": "d0cfdb8c-29ac-4391-a0c8-353665e93e03",
    "metadata": {},
    "source": [
-    "To provide a more concrete example on how third-party libraries can benefit from `metadata`, we'll give an example with [Panel](https://panel.holoviz.org), a library to build web apps in Python only. Panel already integrates with Param, and makes it easy to display a Parameterized instance as an app component, each Parametery type being automatically mapped to a widget type (e.g., a `Number` Parameter is displayed as a float slider). Customizing the automatically created widget requires additional work (e.g., using the `Widget.from_param(<parameter>)` API). To simplify this process, Panel could inspect the `metadata` attribute for custom widget configuration, and apply it on widget creation:"
+    "To provide a more concrete example on how third-party libraries can benefit from the `metadata` attribute, we'll give an example with [Panel](https://panel.holoviz.org), a library to build web apps written in Python only. Panel already integrates with Param, and makes it easy to display a Parameterized instance as an app component, each Parameter type being automatically mapped to a widget type (e.g., a `Number` Parameter is displayed as a float slider). Customizing the automatically created widget requires additional work in Panel (e.g., using the `Widget.from_param(<parameter>)` API). To simplify this process, Panel could inspect the `metadata` attribute for custom widget configuration, and apply it on widget creation:"
    ]
   },
   {


doc/user_guide/Parameters.ipynb
```diff
@@ -929,6 +930,80 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "32cc7b01-55a2-43dd-b886-acf0c30cb5ba",
+   "metadata": {},
+   "source": [
+    "## Arbitrary Metadata\n",
+    "\n",
+    "Parameters may include arbitrary metadata as a mapping stored in the `metadata` attribute (default is `None`). This attribute is not used internally by Param, instead it is meant to enable rich functionality in third-party libraries, or to be leveraged in your own code base to extend Parameters with information that cannot be declared with the other attributes available. Note that a shallow copy of the dictionary you'll pass will be made on class creation and on instance-level parameter creation. This behavior is not specific to the `metadata` attribute, it is how Param handles attributes with value that are mutable containers (lists, dicts, sets)."
```
jbednar: Here it describes the value as a dictionary, whereas above it's described as a mapping. Should presumably be consistent, e.g. by mentioning a dictionary as one possible value rather than the only one expected.

doc/user_guide/Parameters.ipynb
```diff
@@ -929,6 +930,80 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d3cd5940-9b40-48a3-9fe1-4a14f2c6c803",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "p.param['s'].metadata is P.param['s'].metadata  # shallow copy made when the instance-level Parameter is created"
+   ]
```
jbednar, line 1005: I think it's useful to provide an example specific to Panel somewhere in Param's docs, emphasizing that it's just one example. That way the intent will be clearer.

jbednar (approved):
Nice!

doc/user_guide/Parameters.ipynb
maximlt: Addressed in https://github.com/holoviz/param/pull/1094/commits/6f9f4688b708d9a5b2bcdbdb3b77fc99fdb3a6cf, simply mentioning mapping everywhere.

doc/user_guide/Parameters.ipynb
maximlt: Added in https://github.com/holoviz/param/pull/1094/commits/6f9f4688b708d9a5b2bcdbdb3b77fc99fdb3a6cf

doc/user_guide/Parameters.ipynb
```diff
@@ -929,6 +930,105 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d0cfdb8c-29ac-4391-a0c8-353665e93e03",
+   "metadata": {},
+   "source": [
+    "To provide a more concrete example on how third-party libraries can benefit from `metadata`, we'll give an example with [Panel](https://panel.holoviz.org), a library to build web apps in Python only. Panel already integrates with Param, and makes it easy to display a Parameterized instance as an app component, each Parametery type being automatically mapped to a widget type (e.g., a `Number` Parameter is displayed as a float slider). Customizing the automatically created widget requires additional work (e.g., using the `Widget.from_param(<parameter>)` API). To simplify this process, Panel could inspect the `metadata` attribute for custom widget configuration, and apply it on widget creation:"
```
maximlt:
```suggestion
    "To provide a more concrete example on how third-party libraries can benefit from the `metadata` attribute, we'll give an example with [Panel](https://panel.holoviz.org), a library to build web apps written in Python only. Panel already integrates with Param, and makes it easy to display a Parameterized instance as an app component, each Parameter type being automatically mapped to a widget type (e.g., a `Number` Parameter is displayed as a float slider). Customizing the automatically created widget requires additional work in Panel (e.g., using the `Widget.from_param(<parameter>)` API). To simplify this process, Panel could inspect the `metadata` attribute for custom widget configuration, and apply it on widget creation:"
```

Issue #934, opened 2024-04-06
Discussion about the intro video


maximlt:
Fixing https://github.com/holoviz/param/issues/932, I noticed that the link to download the Promo notebook on which is based the intro video was broken (https://tinyurl.com/parampromo). The notebook links to an image in `docs/assets`, link that is likely to be broken too.

Beyond these issues, the video, even if it's nice, is at this stage a bit outdated (no bind, rx, HTML repr, etc.). I also have to say that as someone who has looked at many websites of Python packages, I feel that having an intro video for a package like Param is unexpected (I don't think I would watch it, to be honest). I would instead expect code snippets and a bullet list of features, that help me to quickly grasp what the library offers. With links to sections of interest to dig further. See this for instance https://docs.pydantic.dev/latest/. I would by the way expect the same thing on the `README`. So my preference would be to replace the video in favor of that kind of content. What do others think about it? I'm not super happy suggesting that we get rid of something that probably needed a lot of work to create, at the same time, it's also difficult to maintain over time.

MarcSkovMadsen:
Sounds good to me

PR #1095, opened 2025-11-12
Update the README and the landing page

maximlt:
Resolves https://github.com/holoviz/param/issues/934

I've wanted for a long time to extend the README with more examples, and update the landing page a little bit too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 This combination of **rich attributes** and **reactive APIs** makes Param a solid foundation for constructing user interfaces, graphical applications, and responsive systems where data integrity and automatic synchronization are paramount. In fact, Param serves as the backbone of HoloViz’s [Panel](https://panel.holoviz.org) and [HoloViews](https://holoviews.org) libraries, powering their rich interactivity and data-driven workflows.
 
-Here is a very simple example showing both features at play. We declare a `UserForm` class with three Parameters: `age` and `name` for user data, and the `submit` event to simulate a button in a user interface. We also declare that the `save_user_to_db` method should be called automatically when the `submit` attribute changes.
+Here is a very simple example showing both features at play. We declare a UserForm class with three parameters: `age` as an *Integer* parameter and and `name` as a *String* parameter for user data, and `submit` as an *Event* parameter to simulate a button in a user interface. We also declare that the `save_user_to_db` method should be called automatically when the value of the `submit` attribute changes.
 
 ```python
 import param
@@ -158,16 +158,17 @@ print(cprocessor.mode)  # => 'accurate'
 
 Param extends beyond rich class attributes with a suite of APIs for reactive programming. Let's do a quick tour!
 
-We'll start with APIs that trigger side-effects only, which either have the name watch in their name or are invoked with `watch=True`:
+We'll start with APIs that trigger side-effects only, which either have the noun watch in their name or are invoked with `watch=True`:
 
 1. `<parameterized_obj>.param.watch(fn, *parameters, ...)`: Low-level, imperative API to attach callbacks to parameter changes, the callback receives one or more rich `Event` objects.
 2. `@depends(*parameter_names, watch=True)`: In a Parameterized class, declare dependencies and automatically watch parameters for changes to call the decorated method.
 3. `bind(fn, *references, watch=True, **kwargs)`: Function binding with automatic references (parameters, bound functions, reactive expressions) watching and triggering on changes.
 
-Running the snippet below, you will see that updating the parameters `a`, `b`, and `c` will emit print statements.
-
 ```python
 import param
+
+def debug_event(event: param.parameterized.Event):
+    print(event)
 
 class SideEffectExample(param.Parameterized):
     a = param.String()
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,18 +1,352 @@
...
+        ...
+
+user = UserForm(name='Bob', age=25)
+
+user.submit = True  # => Saving user to db: name=Bob, age=25
+```
+
+---
+
+Enjoying Param? Show your support with a [Github star](https://github.com/holoviz/param) — it’s a simple click that means the world to us and helps others discover it too! ⭐️
```
jbednar: Hmm. I'm not sure it does mean the world to us, since it's such an odd metric! But I do agree it will help others discover it too, so I suggest:

```suggestion
Enjoying Param? Show your support with a [Github star](https://github.com/holoviz/param) to help others discover it too! ⭐️
```

README.md
```diff
@@ -1,18 +1,352 @@
...
+# to update multiple parameters at once, or `.values()` to obtain
+# a dict of parameter name to parameter value.
+processor.param.update(mode='accurate', verbose=False)
+print(processor.param.values())
+# => {'mode': 'accurate', 'name': 'Processor00042', 'retries': 3, 'verbose': False}
+```
+
+Runtime attribute validation is a great feature that helps build defendable code bases! Alternative libraries, like [Pydantic](https://docs.pydantic.dev) and others, excel at input validation, and if this is only what you need, you should probably look into them. Where Param shines is when you also need:
+
+- Attributes that are also available at the class-level, allowing to easily configure a hierarchy of classes and their instances.
```
jbednar:
```suggestion
- Attributes that are also available at the class level, allowing to easily configure a hierarchy of classes and their instances.
```

README.md
```diff
@@ -1,18 +1,352 @@
...
+
+# This namespace offers many more useful methods, such as `.update()`
+# to update multiple parameters at once, or `.values()` to obtain
+# a dict of parameter name to parameter value.
+processor.param.update(mode='accurate', verbose=False)
+print(processor.param.values())
+# => {'mode': 'accurate', 'name': 'Processor00042', 'retries': 3, 'verbose': False}
+```
+
+Runtime attribute validation is a great feature that helps build defendable code bases! Alternative libraries, like [Pydantic](https://docs.pydantic.dev) and others, excel at input validation, and if this is only what you need, you should probably look into them. Where Param shines is when you also need:
```
jbednar, line 115: I'd personally say that Pydantic focuses on sanitization or coercion rather than validation, coercing values rather than rejecting them (see https://github.com/pydantic/pydantic/issues/578#issue-452384685), but maybe that's too much detail to get into here.

README.md
```diff
@@ -1,18 +1,352 @@
...
+class SideEffectExample(param.Parameterized):
+    a = param.String()
+    b = param.String()
+    c = param.String()
+    d = param.String()
+
+    def __init__(self, **params):
+        super().__init__(**params)
+        self.param.watch(lambda event: print(event), 'a')  # 1.
+
```
jbednar, line 182: Is it truly necessary to have an `__init__` here? I worry that it conveys a level of complexity that param users can generally avoid, even when doing reactive programming.

README.md
```diff
@@ -1,18 +1,352 @@
...
+        print(f"print_b: {self.b=}")
+
+sfe = SideEffectExample()
+sfe.a = 'foo'
+# Event(what='value', name='a', obj=SideEffectExample(a='foo', b='', c='', d='', name='SideEffectExample00015'), cls=Example(a='foo', b='', c='', d='', name='SideEffectExample00015'), old='', new='foo', type='changed')
+
+sfe.b = 'bar'
+# print_b: self.b='bar'
+
+@param.depends(sfe.param.c, watch=True)  # 3.
```
jbednar: I would suggest not documenting using `depends()` outside of a class because of how the decorator syntax encourages early binding to a reactive object, which I think limits how the function can be used. I don't know of any advantage that usage has over `bind` and would always steer users to `bind` instead when using functions.

README.md
```diff
@@ -1,18 +1,352 @@
...
+
+sfe.d = 'biz'
+# print_d: d='biz'
+```
+
+Let's continue the tour with what we'll call "reactive APIs". Contrary to the APIs presented above, in this group parameter updates do not immediately trigger side effects. Instead, these APIs let you declare relationships, dependencies, and expressions, which can be introspected by other frameworks to set up their own reactive workflows.
+
+1. `@depends(*parameter_names)`: In a Parameterized class, declare parameter dependencies by decorating a method.
+2. `@depends(*parameters)`: Equivalent of 1 above to decorate functions.
+3. `bind(fn, *references, **kwargs)`: Create a bound function, that when called, will always use the current parameter/reference value. `bind` is in a way a reactive version of [`functools.partial`](https://docs.python.org/3/library/functools.html#functools.partial).
```
jbednar:
```suggestion
3. `bind(fn, *references, **kwargs)`: Create a bound function, that when called, will always use the current parameter/reference value. `bind` is essentially a reactive version of [`functools.partial`](https://docs.python.org/3/library/functools.html#functools.partial).
```

README.md
```diff
@@ -1,18 +1,352 @@
...
+4. `rx()`: Fluent API to create reactive expressions, that allows chaining and composing operations.
```
jbednar:
```suggestion
4. `rx()`: Fluent API to create reactive expressions, which allow chaining and composing operations.
```

README.md
```diff
@@ -1,18 +1,352 @@
...
+bound_mul()
+# 8
+```
+
+`rx()` is the highest-level reactive API Param offers. We'll show a simple example first, using literal values as input of three source reactive expressions, that we combine with simple arithmetic operations.
+
+```python
+from param import rx
+
+# Reactive expressions can be created by simply calling an object with `rx()`
```
jbednar: I don't think it's valid to describe `rx(y)` as "calling y". It's calling `rx` _with_ `y`, right?

```suggestion
# Reactive expressions can be created by simply wrapping an object with `rx()`
```

README.md
```diff
@@ -1,18 +1,352 @@
...
+
+class RXExample(param.Parameterized):
+    val1 = param.String('foo')
+    val2 = param.String('bar')
+
+example = RXExample()
+
+# Reactive expressions can be created directly from Parameters by
+# calling `.rx()`. The source of the reactive expression is not a
+# literal value, but the more dynamic Parameter object.
```
jbednar:
```suggestion
# literal value, but a fully dynamic Parameter object.
```

README.md
```diff
@@ -1,18 +1,352 @@
...
+print(example.param.val1.rx().title().rx.value)  # => 'Foo'
+
+# By updating the parameter value, we update the source value of
+# this expression.
+example.val1 = 'fab'
+print(example.param.val1.rx().title().rx.value)  # => 'Fab'
+
+# Reactive expressions aren't constrained to one data type, the source
```
jbednar:
```suggestion
# Reactive expressions aren't constrained to one data type; the source
```

jbednar (approved):
Nicely done! Thanks for doing this.

README.md
maximlt: Thanks suggestion accepted. FYI I just copied that from Panel's README.

README.md
maximlt: Yeah, I think they've gone back and forth on how they want to describe themselves between validation and parsing/coercion. In v2, they added a strict mode (configurable per class or per field) that prevents coercion, and decided to stick to "validation" (more info in the middle of this long blog post https://pydantic.dev/articles/pydantic-v2#im-dropping-the-word-parse-and-just-using-validate).

image

I'd prefer to stick to "validation" here as a general term.

README.md
maximlt: Well, I could remove it, instantiate the class, and then attach the watcher with `sfe.param.watch(...)` but it's not the same, as what I generally want is for all the instances to have the watcher set up. I believe that having to override `__init__` is very common when writing Parameterized classes (and is a reason why I want to add `post_init`).

README.md
maximlt: I'm curious about what others think (and believe it's time to reach a clear consensus) so I asked on Discord, quoting your comment. Not sure how long it'll take for people to answer, and if there will actually be a consensus (I guess yes but who knows). In case it takes too long, I think this could still be merged as is, and the README updated based on feedback later.

maximlt: I removed these examples.

maximlt:
Added a section with `allow_refs=True` to help users understand how they can directly use bound functions/reactive expressions in Param without having to rely on another framework to introspect them. Also made a few minor edits. I'm going to merge as the README can easily be edited later (its changes being directly reflected on Github).

PR #1104, opened 2025-12-22
Remove Parameterized `__init__` docstring as it shadows class user docstrings

maximlt:
A docstring was added to `Parameterized.__init__` in https://github.com/holoviz/param/pull/998/commits/b66a846ca330a52ebb3d6bce7ad1e8c7c275d024. This caused some issues in VSCode, which would now display that docstring instead of user-defined class docstrings (what Panel components typically define). @MarcSkovMadsen initially reported it in this issue against Pylance https://github.com/microsoft/pylance-release/issues/7793.

This PR simply removes the `Parameterized.__init__` docstring, adding some of its content to the class docstring.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -5348,6 +5348,15 @@ class Parameterized(metaclass=ParameterizedMetaclass):
     This makes it well-suited for robust, maintainable code bases and
     particularly useful in interactive applications requiring reactive behavior.
 
+    To construct an instance, optional Parameter values must be supplied as
+    keyword arguments (``param_name=value``), overriding their default values
+    for this one instance. Any parameters not explicitly set will retain their
+    defined default values.
+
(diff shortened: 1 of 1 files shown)


param/parameterized.py
```diff
@@ -5411,37 +5411,6 @@ class Parameterized(metaclass=ParameterizedMetaclass):
         plus a unique integer""")
 
     def __init__(self, **params):
-        """
```
jbednar, line 5414: Thanks! Is there somewhere else this info can be put, at least as comments along with a comment that says why it's not a docstring?

maximlt: Done in https://github.com/holoviz/param/pull/1104/commits/70db6da4bc6e2c0461f729e7f192bb56ee758098.

Issue #1105, opened 2025-12-28
Chaining map triggers evaluation


Dsolnik:
#### Description of expected behavior and the observed behavior

When chaining together multiple maps, we shouldn’t eagerly evaluate.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
In [1]: a = [1,2,3]

In [2]: def f(x):
   ...:     print(x)
   ...: 

In [3]: v = param.rx(a).rx.map(f)

In [4]: v = param.rx(a).rx.map(f).rx.map(f)
1
2
3
```

PR #1106, opened 2026-01-16
feat: add rx(…, lazy=)

camriddell:
Added an optional flag to `rx(..., lazy: bool = True | False)`. When `lazy=True` chaining operations does not evaluate previous operations. Whereas when `lazy=False` the chaining a new operation evaluates the previous operation (original behavior).

Addresses #1105 

This is the current/original behavior

```python
fcalls = 0
def debug_call_count(value):
    nonlocal fcalls
    fcalls += 1
    return value

base = rx(input, lazy=False).rx.pipe(debug_call_count)
title = base.title() # evaluates the previous `.pipe` step
assert fcalls == 1
assert title.rx.value == 'Bob'
assert fcalls == 1 # the entire chain is not re-evaluated
```

When using `lazy=True`

```python
fcalls = 0
def debug_call_count(value):
    nonlocal fcalls
    fcalls += 1
    return value

base = rx(input, lazy=True).rx.pipe(debug_call_count)
title = base.title() # does not evaluate the previous `.pipe` step
assert fcalls == 0
assert title.rx.value == 'Bob' # the entire chain is evaluated here
assert fcalls == 1
```

### DIFF
--- a/tests/testimports.py
+++ b/tests/testimports.py
@@ -6,16 +6,10 @@ from textwrap import dedent
 def test_no_blocklist_imports():
     # setuptools_scm is imported when importing param. It itself imports
     # the logging module.
-    blocklist = {"numpy", "IPython", "pandas", "asyncio", "html", "logging"}
-    try:
-        import setuptools_scm  # noqa
-        blocklist.remove("logging")
-    except ModuleNotFoundError:
-        pass
+    blocklist = {"numpy", "IPython", "pandas", "asyncio", "html"}
 
     check = f"""\
     import sys
-    sys.modules.pop("logging", None)  # Added because of coverage will add it
     import param
 
     blocklist = {blocklist}


camriddell:
@philippjfr all stable now. Let me know if this is the way you envisioned the API or if there's a different approach you'd like me to try out.

param/reactive.py
```diff
@@ -1648,31 +1657,59 @@ def _invalidate_obj(self, *events):
...
+                if stale():
+                    return
+                self._current_ = shared.rx.value
+                trigger.param.trigger('value')
+            elif inspect.isasyncgen(obj):
+                async for val in obj:
+                    if stale():
+                        try:
+                            await obj.aclose()
+                        except Exception:
```
```suggestion
                        except (StopAsyncIteration,  GeneratorExit):
```

I think we should avoid a bare `Exception` here, I _think_ this is what is needed.

hoxbro (approved):

param/reactive.py
philippjfr: Not sure about that, I'd be okay with logging other exceptions but errors in the generator cleanup logic should not bleed through here and potentially break a shared downstream task via:

```
    if shared._current_task:
         await shared._current_task
```

PR #1114, opened 2026-03-13
Accept array-like objects in param.Array via __array__ protocol

ghostiee-11:
## Summary

I was building a Panel dashboard that uses `param.Array` to store label arrays from a pandas DataFrame. With pandas 2.x using pyarrow as the default string backend, `df["label"].values` returns an `ArrowStringArray` instead of a numpy array. This causes `param.Array` to reject it with a `ValueError`, even though the object fully supports the numpy array protocol.

```python
import param, pandas as pd

class State(param.Parameterized):
    labels = param.Array()

s = State()
df = pd.DataFrame({"label": ["cat", "dog", "cat"]})
s.labels = df["label"].values
# ValueError: Array parameter 'State.labels' value must be an instance of ndarray, not <ArrowStringArray>
```

I had to work around this everywhere with `np.array(df[col].tolist())` which felt unnecessary since the object already implements `__array__`.

## Fix

I overrode `_validate_class_` in `Array` to also accept objects implementing `__array__` or `__array_interface__`, while still rejecting plain lists, strings, and other non-array types.

### After
```python
s.labels = df["label"].values  # works now (ArrowStringArray)
s.labels = pd.Categorical(["a", "b", "a"])  # works
s.labels = [1, 2, 3]  # still rejected (no __array__)
s.labels = "not an array"  # still rejected
```

## Changes

- `param/parameters.py`: Added `_is_array_like()` static method and `_validate_class_()` override in `Array` to accept objects with `__array__` or `__array_interface__` protocol
- `tests/testnumpy.py`: Added 5 new tests covering `__array__`, `__array_interface__`, plain list rejection, string rejection, and pandas ExtensionArray (Categorical + ArrowStringArray)

## Test plan

- [x] All 1449 existing param tests pass (0 failures)
- [x] 5 new tests covering acceptance and rejection cases
- [x] Verified `serialize()` / `tolist()` works with ArrowStringArray
- [x] Verified `np.asarray()` round-trip works on stored values

### DIFF
--- a/param/parameters.py
+++ b/param/parameters.py
@@ -2233,8 +2233,9 @@ class Array(ClassSelector):
     """Parameter whose value is a numpy array.
 
     Accepts numpy ``ndarray`` objects as well as array-like objects that
-    implement the ``__array__`` protocol (e.g. pandas ``ExtensionArray``
-    subclasses such as ``ArrowStringArray``).
+    implement the ``__array__`` or ``__array_interface__`` protocols
+    (e.g. pandas ``ExtensionArray`` subclasses such as
+    ``ArrowStringArray``).
     """
 
     @typing.overload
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks for this. I'm not yet 100% whether we should support it on `Array` directly, add a flag to enable the support or add a `ArrayLike` parameter type. See https://github.com/holoviz/param/pull/975 for the equivalent for DataFrame objects.

hoxbro:
I agree that we shouldn't loosen the strictness of `param.Array`. `ArrayLike` seems like a good for a new parameter. 

I do think this PR is still quite a way from being in a mergeable state, even with a name change.

jbednar:
I do like `ArrayLike`, as it clearly signals what type of object will be accepted.

PR #1129, opened 2026-04-10
 fix: `param.depends` method call in class

hoxbro:
## Description

Fixes the following error seen in HoloViews. Also updating the infrastructure a bit. 

``` python
from __future__ import annotations

import param


class MyClass(param.Parameterized):
    value = param.Parameter(default=None)

    @param.depends("value", watch=True)
    def _on_change(self) -> None:
        pass

    def trigger(self) -> None:
        self._on_change()  # error: missing-argument
```

Pyright gives the following warnings, but I don't see a way to remove them without making other type checkers fail:

```
/home/shh/projects/holoviz/repos/param/param/depends.py
  /home/shh/projects/holoviz/repos/param/param/depends.py:50:39 - warning: TypeVar "_S" appears only once in generic function signature
    Use "object" instead (reportInvalidTypeVarUse)
  /home/shh/projects/holoviz/repos/param/param/depends.py:56:39 - warning: TypeVar "_S" appears only once in generic function signature
    Use "object" instead (reportInvalidTypeVarUse)
```

## How Has This Been Tested?

CI + locally

## AI Disclosure

- [x] This PR contains AI-generated content.

Tools and Models:  Claude Code + Sonnet 4.6, To apply the type problem.

## Checklist

- [ ] Tests added and are passing
- [ ] Added documentation

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -15,7 +15,7 @@ default = [
     "example",
     "test-example",
     "lint",
-    "typecheck",
+    "type",
     "doc",
     "dev",
 ]
@@ -60,8 +60,8 @@ no-default-feature = true
 features = ["py313", "lint"]
 no-default-feature = true
 
-[environments.typecheck]
-features = ["py313", "required", "typecheck"]
+[environments.type]
+features = ["py313", "required", "type", "type-task"]
 no-default-feature = true
 
 [feature.required.dependencies]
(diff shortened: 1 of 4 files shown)


pixi.toml
```diff
@@ -204,20 +204,29 @@ lint-install = 'pre-commit install'
...
 typing_extensions = "*"
 tornado = "*"
 
-[feature.typecheck.tasks]
-type-ty = 'ty check'
-type-mypy = 'mypy'
-type-pyrefly = 'pyrefly check --output-format min-text --count-errors=0'
-type-pyright = 'pyright'
-type-all = { depends-on = ["type-ty", "type-mypy", "type-pyrefly", "type-pyright"] }
+[feature.type-task.tasks]
```
hoxbro, line 218: Making it `type-task` so when we run `pixi run type-all`. It uses the environment we actually test for in the CI to avoid running it in default environment.

Added arg to the code, so we can `pixi run type-all mre.py` and only run that file.

pixi.toml
```diff
@@ -204,20 +204,29 @@ lint-install = 'pre-commit install'
...
 # =============== TYPECHECK ===================
 # =============================================
-[feature.typecheck.dependencies]
+[feature.type.dependencies]
 ty = "*"
 mypy = "*"
 numpy = "*"
 pandas = "*"
-pyrefly = "!=0.60.0"  # https://github.com/facebook/pyrefly/issues/3042
+pyrefly = "*"
```
hoxbro, line 212: New release have been made. Ignored it in the code for now,

philippjfr (approved):

PR #1132, opened 2026-04-14
fix: Update generics for ClassSelector

hoxbro:
## Description

Started to see some type failures in HoloViews after #1131 was merged.

There is two changes. I'm not sure about the first one though.

1. Changed `self: ClassSelector[CT | None]` to `self: ClassSelector[CT]`. The `| None` caused ty to fail invariance checking. Using self: ClassSelector[CT] lets ty infer CT directly from class_: type[CT].
2. New overload added: `class_: tuple[type, ...]` with `allow_None=True`, which was missing.

``` python
import param


class MyElement:
    def clone(self, data):
        return self

    def iloc(self, index):
        return self


class Container(param.Parameterized):
    # Error: no-matching-overload
    # All overloads require either `default` or specific combinations.
    # This is valid at runtime but ty can't resolve it.
    element = param.ClassSelector(
        class_=MyElement,
        doc="The element.",
    )

    # Error: no-matching-overload when class_ is a tuple
    multi = param.ClassSelector(
        class_=(list, tuple),
        allow_None=True,
        doc="A list or tuple.",
    )


class User(param.Parameterized):
    container = param.ClassSelector(
        class_=Container,
    )

    def use(self):
        # Error: unresolved-attribute — ty infers `element` as
        # `Unknown | _T@ClassSelector` and doesn't know it has `.clone`
        return self.container.element.clone([])
```

## AI Disclosure

- [x] This PR contains AI-generated content.

Tools and Models: Claude + Sonnet 4.6. To create the MRE based on HoloViews examples and then for fixing them in Param.

### DIFF
--- a/param/parameters.py
+++ b/param/parameters.py
@@ -2939,7 +2939,7 @@ class ClassSelector(SelectorBase[_T]):
 
         @t.overload
         def __init__(
-            self: ClassSelector[CT | None],
+            self: ClassSelector[CT],
             *,
             default: None = None,
             class_: type[CT],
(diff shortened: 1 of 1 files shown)


param/parameters.py
```diff
@@ -2939,7 +2939,7 @@ def __init__(
 
         @t.overload
         def __init__(
-            self: ClassSelector[CT | None],
```
philippjfr, line 2942: Don't follow this one. `default=None` should resolve to `CT | None` surely.

philippjfr: Hmm, reading your description I get it but pyright did seem to correctly infer things and that's the main one we care about and now I assume it does not infer `param.ClassSelector(class_=int)` correctly as `int | None` (for example)?

hoxbro: Is that correct if `allow_None` is False?

hoxbro: It is. reverted this.

PR #1133, opened 2026-04-15
Implement param.Model: A Parameterized subclass based on type annotations

philippjfr:
Based on discussion in https://github.com/holoviz/param/pull/1066 I prototyped `ParamModel`.

**Summary**

Adds `ParamModel`, a new `Parameterized` subclass that synthesises `Parameter` instances automatically from PEP 526 type annotations. This gives users a dataclass-like authoring experience without requiring manual `Parameter` instantiation for common cases.

**Motivation**

After investigating `@dataclass_transform`, it became clear that the decorator cannot bind RHS `Parameter` descriptor types to LHS attribute types in a way that satisfies type checkers, you'd still need to repeat yourself (`string_param: str = String()`) and even then checkers would see `String[str]`, not `str`. Additionally, `dataclass_transform` requires every field specifier to be enumerated upfront, which is impossible at `ParameterizedMetaclass` definition time because not all `Parameter` subclasses exist yet.

`'ParamField',` sidesteps these constraints entirely: the metaclass handles synthesis at class creation time, and a dedicated `ParamField()` factory covers the cases where users need to supply explicit parameter configuration.

**Changes**

`ParamModel` (decorated with `@dataclass_transform(field_specifiers=(ParamField,))`) overrides `__new__` to walk the class `__annotations__` dict at class creation time. For each non-private, non-`ClassVar` annotation it calls `_annotation_parameter_factory` to map the annotation to an appropriate `Parameter` subclass and infer constructor kwargs, then calls `_build_parameter_from_field` to construct the final `Parameter` object. String annotations are `eval`'d against the module globals to support `from __future__ import annotations`.

`_annotation_parameter_factory` handles the following mappings:
- `bool` → `Boolean`
- `int` → `Integer`
- `float` → `Number`
- `str` → `String`
- `list[T]` → `List(item_type=T)`
- `tuple[...]` → `Tuple(length=N)` (fixed-length) or `Tuple` (variable)
- `dict[...]` → `Dict`
- `set` → `ClassSelector(class_=set)`
- `Literal[...]` → `Selector(objects=[...])`
- `T | None` / `Optional[T]` → wraps the inner type with `allow_None=True`
- `Annotated[T, {...}]` → passes the mapping as extra kwargs to the parameter constructor
- `Any` / `object` → bare `Parameter`

`ParamField()` is a typed factory (three overloads) that returns a `_FieldSpec` sentinel. It accepts `default`, `default_factory`, a `parameter` override (class, callable, or instance), and arbitrary kwargs forwarded to the parameter constructor. Type checker overloads ensure that `default: FT` and `default_factory: Callable[[], FT]` both propagate `FT` to the attribute type, matching dataclass semantics.

`_build_parameter_from_field` merges the inferred factory and kwargs with anything supplied via `Field`, giving explicit `Field` values precedence. If a `Parameter` *instance* is passed as the `parameter` argument it is shallow-copied and mutated rather than re-instantiated.

**Usage**

```python
class MyModel(ParamModel):
    # annotation-only → synthesised automatically
    name: str
    count: int = 0
    ratio: float | None = None
    tags: list[str]
    mode: Literal["fast", "slow"] = "fast"

    # explicit parameter config via Field
    label: str = ParamField(default="untitled", doc="Human-readable label")

    # explicit parameter type override
    path: str = ParamField(parameter=param.String, regex=r"^/")

    # Regular class attribute
    attr: ClassVar[str] = "foo"
```


philippjfr:
Some suggestions instead of `Field`:

- `ParamSpec` / `ParameterSpec`
- `ParamConfig` / `ParameterConfig`
- `ParameterHint`

jbednar:
`param.Hints`?

philippjfr:
Alternatively `param.Param`.

philippjfr:
And if we're already riffing on more succinct naming for the Parameter factory, then I'd also suggest `ParamModel` or `ParamClass` as names for `TypedParameterized`. `TypedParameterized` I no longer love because Parameterized is to some extent also typed.

philippjfr:
Okay, my favorite proposal now is `ParamModel` and `ParamField`.

jbednar:
> TypedParameterized I no longer love because Parameterized is to some extent also typed.

I guess it would literally be ParameterizedFromTypeDeclarations?  

> Okay, my favorite proposal now is ParamModel and ParamField.

I don't personally love Model since it is used to mean so many things that it's lost all meaning for me, but I don't object since Pydantic uses that term.

doc/getting_started.md
```diff
@@ -67,6 +67,24 @@ The types in Param may remind you of the static types found in some languages, b
...
+class Sum(param.Model):
+    title: str = "sum"
+    a: int = param.Field(default=2, bounds=(0, 10), doc="First addend")
+    b: int = param.Field(default=3, bounds=(0, 10), doc="Second addend")
+
+    def __call__(self):
+        return self.title + ": " + str(self.a + self.b)
+```
+
+Leave off the `=` and the annotation becomes required: omitting that parameter when constructing the class raises a clear `TypeError` naming the missing field, rather than silently filling in an empty string or zero. See the [Typing](user_guide/Typing) user guide for the full set of supported annotations (`Literal`, `Optional`, containers, and more) and for how `Model` interacts with static type checkers.
```
maximlt:
```suggestion
Leave off the `=` and the annotation becomes required: omitting that parameter when constructing the class raises a clear `TypeError` naming the missing field, rather than silently filling in a default value. See the [Typing](user_guide/Typing) user guide for the full set of supported annotations (`Literal`, `Optional`, containers, and more) and for how `Model` interacts with static type checkers.
```

param/typed.py
```diff
@@ -0,0 +1,327 @@
...
+import copy
+import enum
+import importlib
+import sys
+import types
+import typing as t
+
+from collections.abc import Callable, Mapping
+from typing import Any
+from typing_extensions import dataclass_transform
```
maximlt: Why not import it from `typing`? If required from typing_extensions, then it should be added as a required or optional dependency.

param/typed.py
```diff
@@ -0,0 +1,327 @@
...
+
+        required: set[str] = set()
+        for base in bases:
+            required |= getattr(base, "_model_required", frozenset())
+
+        for attr, annotation in annotations.items():
+            if isinstance(annotation, str):
+                annotation = eval(annotation, module_globals, namespace)
+            if attr.startswith("_"):
+                continue
```
maximlt: Can be moved up before potentially running `eval`.

maximlt: Actually, what is the concept here with handling attributes prefixed with `_`?

param/typed.py
```diff
@@ -0,0 +1,327 @@
...
+    )
+
+    kwargs: dict[str, Any] = {}
+    ann = annotation
+    # `Annotated` and `Optional` can occur in either order.
+    while True:
+        origin = t.get_origin(ann)
+        if origin is t.Annotated:
+            annotated_args = list(t.get_args(ann))
+            for meta in annotated_args[1:]:
```
maximlt: So it means I can do that? `n: Annotated[int, {"bounds": (0, 10)}]`

doc/user_guide/Typing.ipynb
```diff
@@ -180,16 +180,136 @@
...
+    "except TypeError as e:\n",
+    "    print(e)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f00d1e5e",
+   "metadata": {},
+   "source": [
+    "The one exception is `Literal`-derived `Selector` fields (`mode` in the `TrainingRun` example above): these default to the first listed value even without an explicit default, since that mirrors `Selector`'s existing runtime behavior of picking `objects[0]` when no default is given."
```
maximlt: Mirroring what other libraries like pydantic do would also be fine by me, i.e. requiring users to also set the default.

param/typed.py
```diff
@@ -0,0 +1,327 @@
...
+            elem = list_args[0]
+            if isinstance(elem, type):
+                kwargs["item_type"] = elem
+            elif t.get_origin(elem) in (t.Union, types.UnionType):
+                elem_types = tuple(a for a in t.get_args(elem) if isinstance(a, type))
+                if elem_types:
+                    kwargs["item_type"] = elem_types
+        return List, kwargs
+
+    if ann is tuple or origin in (tuple, t.Tuple):
```
maximlt: `t: tuple[int, ...]` and `P(t=(1, 1, 1))` raises this:
```
ValueError: Attribute 'length' of Tuple parameter 'P.t' is not of the correct length (3 instead of 2).
```

param/typed.py
```diff
@@ -0,0 +1,327 @@
...
+            from .parameters import Selector
+
+            # Selectors derive a default from `objects`; other fields are required.
+            factory_kwargs["default"] = Undefined
+            is_required = not (factory is Selector and factory_kwargs.get("objects"))
+
+    if isinstance(factory, Parameter):
+        pobj = copy.copy(factory)
+        for key, value in factory_kwargs.items():
+            setattr(pobj, key, value)
```
maximlt: There are likely some caveats with skipping the Parameter initialization.

maximlt (commented):
- New APIs need to be referenced somewhere in `/docs/reference/param`
- Inlined imports in typed.py don't seem necessary

param/typed.py
philippjfr: Good point, currently just skipped but probably shouldn't be.

PR #1135, opened 2026-04-16
Loosen ClassSelector default typing to handle subclass invariance

philippjfr:
Without this change type checkers would complain about defaults that are subclass of the `class_`.


philippjfr:
Not needed.

PR #1139, opened 2026-04-30
Add Typing user guide

philippjfr:
Adds a user guide to explain the basics behind typing in Param, current limitations, best practices and the future of typing.

### DIFF
--- a/doc/user_guide/Typing.ipynb
+++ b/doc/user_guide/Typing.ipynb
@@ -206,7 +206,7 @@
     "- Use specific Parameter classes (`Integer`, `String`, `ClassSelector`, etc.) instead of generic `Parameter` whenever possible.\n",
     "- Use Parameter keyword arguments (`allow_None`, `item_type`, `class_`, `bounds`) to improve both runtime checks and type precision.\n",
     "- For finite choices today, prefer `Literal[...]` annotations with `Selector`.\n",
-    "- Run your type checker (Pyright/mypy) alongside your tests to get both static and runtime safety."
+    "- Run your type checker alongside your tests to get both static and runtime safety."
    ]
   }
  ],
--- a/doc/user_guide/Typing.ipynb
+++ b/doc/user_guide/Typing.ipynb
@@ -107,12 +107,27 @@
   },
   {
    "cell_type": "markdown",
+   "id": "56c5e6bc-0d3e-4a43-834a-feecb0060ecd",
+   "metadata": {},
+   "source": [
+    "## Choice of Type Checker\n",
+    "\n",
+    "The Python ecosystem now has quite a number of different type checking tools. Param itself is type checked against all four of the most common popular type checkers, including:\n",
+    "\n",
+    "- `mypy`: The default type checker for Python. Unfortunately it is somewhat slow and has some limitations.\n",
+    "- `pyright`: The default type checker for VSCode.\n",
+    "- `pyrefly`: A faster (but incomplete) type checker by Meta.\n",
+    "- `ty`: A fast type checker written in Rust by the Astral team (the makers of `uv` and `ruff`). Still incomplete but improving every day.\n",
+    "\n",
+    "If you are developing a library built on Param, we recommend using pyright as your type checker. Param's type annotations are primarily optimized for pyright, and it is the checker most likely to benefit your users. VSCode (via Pylance) runs pyright automatically, so correct inference will surface in your users' editors without any extra setup on their part."
+   ]
+  },
+  {
+   "cell_type": "markdown",
    "id": "8cc4ed0d",
    "metadata": {},
    "source": [
-    "## Exceptions & Limitations\n",
-    "\n",
-    "### Type Narrowing\n",
+    "## Limitations\n",
     "\n",
     "Unfortunately type narrowing can only get you so far in Python's type system. Specifically `Literal` values, as in when a `Selector` parameter is bound on a `Parameterized` class cannot be automatically inferred from the `objects`, e.g.:"
    ]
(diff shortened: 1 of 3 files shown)


doc/user_guide/Typing.ipynb
```diff
@@ -0,0 +1,255 @@
...
+    "\n",
+    "Both of these issues will be addressed by future proposals."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "987c7a95",
+   "metadata": {},
+   "source": [
+    "## Future direction\n",
```
hoxbro, line 188: I don't think we should have this section yet. Two reasons for this:

1) The format haven't been 100% agreed upon. There could be some shortcomings, which could hinder the current proposes API and naming.
2) An LLM could try to implement this by looking at the documentation, which would then fail.

tests/testpickle.py
```diff
@@ -7,6 +7,8 @@
     import cloudpickle
 except ModuleNotFoundError:
     cloudpickle = None
+except AttributeError:
+    cloudpickle = None  # PyPy on Windows
```
hoxbro:
```suggestion
except (ModuleNotFoundError, AttributeError): # PyPy on Windows
    cloudpickle = None
```

doc/user_guide/Typing.ipynb
```diff
@@ -0,0 +1,259 @@
...
+   "source": [
+    "This is unfortunate and one reason why this is not the end to the typing story for Param."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "40fa161f-6282-42b9-a41f-e838bf74ff08",
+   "metadata": {},
+   "source": [
+    "### Signature\n",
```
hoxbro: I don't think this is relevant.

tests/testpickle.py
```diff
@@ -8,6 +8,8 @@
 except (AttributeError, ModuleNotFoundError):
     # Not installed or PyPy on Windows
     cloudpickle = None
+except AttributeError:
+    cloudpickle = None  # PyPy on Windows
```
philippjfr:
```suggestion
```

doc/user_guide/Typing.ipynb
philippjfr: It is relevant in the context of the future section but if we indeed decide to remove that then this doesn't make sense.

doc/user_guide/Typing.ipynb
philippjfr: I agree with 2., on 1. I'm pretty confident that the approach can work (I've got a near full conversion of Panel sitting in my stash).

doc/user_guide/Typing.ipynb
```diff
@@ -0,0 +1,259 @@
...
+    "\n",
+    "- how types are inferred from Parameter *types* and constructor *arguments*\n",
+    "- where typing support is heading next (annotation-first parameter declarations)\n",
+    "\n",
+    "Before we dive into the details it's important to clarify the difference between the type annotation and parameter declaration:\n",
+    "\n",
+    "- **type annotation** communicates intent to static tools (`Literal`, unions, etc.)\n",
+    "- **Parameter declaration** controls runtime behavior and validation\n",
+    "\n",
+    "Throughout this guide, examples use standard `param.Parameterized` classes, and type hints are meant for tools like Pyright and mypy."
```
maximlt, line 22: Would it be possible to let users know which type checkers have been tested for the adopted approach? Or you prefer to keep this undefined?

doc/user_guide/Typing.ipynb
```diff
@@ -0,0 +1,259 @@
...
+   "metadata": {},
+   "source": [
+    "## Practical recommendations\n",
+    "\n",
+    "For now these are our practical recommendations for typing in Param:\n",
+    "\n",
+    "- Use specific Parameter classes (`Integer`, `String`, `ClassSelector`, etc.) instead of generic `Parameter` whenever possible.\n",
+    "- Use Parameter keyword arguments (`allow_None`, `item_type`, `class_`, `bounds`) to improve both runtime checks and type precision.\n",
+    "- For finite choices today, prefer `Literal[...]` annotations with `Selector`.\n",
+    "- Run your type checker (Pyright/mypy) alongside your tests to get both static and runtime safety."
```
maximlt: Do type checkers need to be configured in a special way to ignore typing features that can't be implemented in Param?

maximlt (approved):
LGTM in general.

The only feedback I have is that I don't know if the guide conveys the effort required for a user to migrate their code base to leverage typing. The "Future Direction" section leaves me wondering whether, as a user, I should just wait for the next release until this more powerful approach is released.

philippjfr:
>The "Future Direction" section leaves me wondering whether, as a user, I should just wait for the next release until this more powerful approach is released.

Yeah, that was also my concern about releasing without introducing the new constructs. It leaves a 2-step upgrade path. To be quite honest though, after having done the migration for Panel, there's very little work that needs to be done and then undone to add proper types and then convert to `ParamModel` (or whatever we call the typed Parameterized). Basically it's only `Selector` and `List` types that need to be specially annotated now and could then be simplified in the future approach.

doc/user_guide/Typing.ipynb
philippjfr: No, seems fine to mention.

doc/user_guide/Typing.ipynb
```diff
@@ -0,0 +1,221 @@
...
+    "## Practical recommendations\n",
+    "\n",
+    "If you are updating an existing Param based codebase we recommend starting to implement typing across your codebase. While the future typing story will make certain things easier, the migration path from old-style `Parameterized` classes to the newer type annotated kind will be relatively straightforward.\n",
+    "\n",
+    "For now these are our practical recommendations for typing in Param:\n",
+    "\n",
+    "- Use specific Parameter classes (`Integer`, `String`, `ClassSelector`, etc.) instead of generic `Parameter` whenever possible.\n",
+    "- Use Parameter keyword arguments (`allow_None`, `item_type`, `class_`, `bounds`) to improve both runtime checks and type precision.\n",
+    "- For finite choices today, prefer `Literal[...]` annotations with `Selector`.\n",
+    "- Run your type checker (Pyright/mypy) alongside your tests to get both static and runtime safety."
```
hoxbro:
```suggestion
    "- Run your type checker alongside your tests to get both static and runtime safety."
```

Nit.

hoxbro (approved):

PR #1140, opened 2026-04-30
Resolve recent ty and pyright type errors

philippjfr:
Recent versions of `ty` caused some additional type errors to pop-up (largely capturing issues that other type checkers already flagged). This PR adds type ignores for these.

Additionally this fixes some pyright errors related to allow_None validation. The `allow_None` validation was repeated a few times leading to some issues.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2255,7 +2255,7 @@ class String(Parameter[_T]):
     def _validate_regex(self, val: t.Any, regex: str | re.Pattern[str] | None):
         if val is None or regex is None:
             return
-        if re.match(regex, val) is None:
+        if re.fullmatch(regex, val) is None:
             raise ValueError(
                 f'{_validate_error_prefix(self)} value {val!r} does not '
                 f'match regex {regex!r}.'
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2253,9 +2253,9 @@ class String(Parameter[_T]):
         self._validate(self.default)
 
     def _validate_regex(self, val: t.Any, regex: str | re.Pattern[str] | None):
-        if (val is None and self.allow_None):
+        if val is None or regex is None:
             return
-        if regex is not None and re.match(regex, val) is None:
+        if re.match(regex, val) is None:
             raise ValueError(
                 f'{_validate_error_prefix(self)} value {val!r} does not '
                 f'match regex {regex!r}.'
--- a/param/parameters.py
+++ b/param/parameters.py
@@ -4071,7 +4071,7 @@ class Color(Parameter[_T]):
     def _validate_allow_named(self, val, allow_named):
         if val is None:
             return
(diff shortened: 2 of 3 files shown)


param/parameters.py
```diff
@@ -1985,7 +1985,7 @@ def __init__(
         allow_None: bool = t.cast("bool", Undefined),  # pyrefly: ignore[bad-argument-type]
         **kwargs: Unpack[_DateInitKwargs]
     ) -> None:
-        super().__init__(default=default, allow_None=allow_None, **kwargs)  # type: ignore[misc, call-overload]
+        super().__init__(default=default, allow_None=allow_None, **kwargs)  # type: ignore[misc, call-overload]  # ty: ignore[invalid-argument-type]
```
hoxbro:
```suggestion
        super().__init__(default=default, allow_None=allow_None, **kwargs)  # type: ignore[misc, call-overload, ty:invalid-argument-type] 
```

I think this will work (or some variant of it.)

param/parameterized.py
```diff
@@ -2253,9 +2253,9 @@ def __init__(
         self._validate(self.default)
 
     def _validate_regex(self, val: t.Any, regex: str | re.Pattern[str] | None):
-        if (val is None and self.allow_None):
+        if val is None or regex is None:
             return
-        if regex is not None and re.match(regex, val) is None:
+        if re.match(regex, val) is None:
```
hoxbro: Unrelated but shouldn't this be `re.search`? This makes it so it start with the regex and not contain the regex. In Python 3.15, it has been soft-deprecated for `prefixmatch` as it confuses people. 

https://docs.python.org/3.15/library/re.html#re.Pattern.prefixmatch

param/parameters.py
```diff
@@ -2830,13 +2831,14 @@ def compute_default(self):
...
     def _validate(self, val):
-        if (val is None and self.allow_None):
+        if val is None and self.allow_None:
             return
         self._validate_type(val)
 
         if self.check_on_set:
             self._validate_value(val)
         else:
+            assert val is not None
```
hoxbro:
```suggestion
            if val is None: 
                raise ValueError(...)
```

hoxbro (approved):
Some minor comments.

param/parameters.py
philippjfr: No, this is already caught before this.

param/parameterized.py
philippjfr: Yes, that seems better. Though maybe re.fullmatch?

param/parameters.py
```diff
@@ -2830,13 +2831,14 @@ def compute_default(self):
...
             for v in val:
```
hoxbro:
```suggestion
            for v in val:  # pyright: ignore[reportOptionalIterable]
```

Then lets ignore

PR #1142, opened 2026-05-04
fix: Use re.search for regex check

hoxbro:
## Description

Fixes `re.fullmatch` as it is too strict.

## How Has This Been Tested?

CI

## Checklist

- [ ] Tests added and are passing

### DIFF
--- a/param/parameters.py
+++ b/param/parameters.py
@@ -4236,7 +4236,7 @@ class Color(Parameter[_T]):
     def _validate_allow_named(self, val, allow_named):
         if val is None:
             return
-        is_hex = re.fullmatch('^#?(([0-9a-fA-F]{2}){3}|([0-9a-fA-F]){3})$', val)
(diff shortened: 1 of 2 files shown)


param/parameters.py
```diff
@@ -4236,7 +4236,7 @@ def _validate_value(self, value, allow_None):
     def _validate_allow_named(self, val, allow_named):
         if val is None:
             return
-        is_hex = re.fullmatch('^#?(([0-9a-fA-F]{2}){3}|([0-9a-fA-F]){3})$', val)
+        is_hex = re.search('^#?(([0-9a-fA-F]{2}){3}|([0-9a-fA-F]){3})$', val)
```
hoxbro, line 4239: This is the same as the regex is wrapped with `^` and `$`.

philippjfr (approved):

PR #1144, opened 2026-05-17
feat: add DataFrameLike parameter for cross-backend dataframe inputs

ghostiee-11:
`param.DataFrame` only accepts `pandas.DataFrame`, so there is no way to declare a parameter that holds tabular data when the value might be Polars, PyArrow, or another backend. This adds a new `DataFrameLike` parameter that validates anything the [Narwhals](https://narwhals-dev.github.io) protocol recognises and passes the native object through unchanged. `param.DataFrame` is deliberately left untouched, so existing pandas-only code keeps its guarantee. This is the separate-class direction discussed in #975; serialization backend-preservation is intentionally left as an open question there.

Same `rows` / `columns` / `ordered` slots as `DataFrame` (driven through Narwhals so they work on every backend), plus `allow_lazy=True` for Polars `LazyFrame` / Dask / DuckDB with no implicit collect. Narwhals is an optional dependency, deferred like pandas is for `DataFrame`, with a clear install message if missing.

**Before** 
dflike-before
**After** 
dflike-after

Tested with pandas, Polars (eager + lazy), and PyArrow; full suite 1550 passed, `testpandas.py` unchanged. Validation goes entirely through Narwhals, so any other Narwhals-supported backend uses the identical code path.


ghostiee-11:
Hey @philippjfr !! Looking for your views over this.

Thanks!!

param/parameters.py
```diff
@@ -3487,6 +3502,239 @@ def deserialize(cls, value):
...
+        import narwhals
+    except ModuleNotFoundError as e:
+        raise ImportError(
+            "param.DataFrameLike requires the optional 'narwhals' package. "
+            "Install it with: pip install narwhals"
+        ) from e
+    return narwhals
+
+
+class DataFrameLike(ClassSelector[t.Any]):
```
hoxbro, line 3506: Not sure if we should call this `DataFrameLike`, whereas it make sense for something like `ArrayLike` as most package conforms to the numpy syntax. Narwhals is not a global standard and may confuse users who are not well-versed in the DataFrame ecosystem.

param/parameters.py
```diff
@@ -3487,6 +3502,239 @@ def deserialize(cls, value):
...
+            allow_None=allow_None,  # type: ignore[arg-type]
+            **params,
+        )
+        self._validate(self.default)
+
+    def _as_narwhals(self, val):
+        narwhals = _get_narwhals()
+        try:
+            return narwhals.from_native(
+                val, eager_only=not self.allow_lazy, pass_through=False
```
hoxbro: Maybe we should just call it `eager_only`? Though not sure about this.

param/parameters.py
```diff
@@ -3487,6 +3502,239 @@ def deserialize(cls, value):
...
+                f"{_validate_error_prefix(self)} value must be a dataframe-like "
+                f"object that Narwhals recognises, not None."
+            )
+
+        nwframe = self._as_narwhals(val)
+        narwhals = _get_narwhals()
+        is_lazy = isinstance(nwframe, narwhals.LazyFrame)
+        # ``collect_schema().names()`` is the Narwhals-recommended way to read
+        # column names uniformly across eager and lazy frames; ``.columns`` on
+        # a lazy frame triggers a backend schema-resolution warning.
```
hoxbro: Don't see the need for this comment.

param/parameters.py
```diff
@@ -3487,6 +3502,239 @@ def deserialize(cls, value):
...
+    are exercised in this project's test suite; any other Narwhals-supported
+    backend uses the identical code path. The value is passed through
+    unchanged, so reading the parameter returns the original native object
+    (no Narwhals wrapper). Authors who want a backend-agnostic API can call
+    ``narwhals.from_native`` on the value themselves.
+
+    Narwhals is an optional dependency, imported on instantiation; a clear
+    ``ImportError`` with the install command is raised if it is missing. The
+    structure of the frame can be constrained by the rows and columns
+    arguments:
```
hoxbro: I think this is heavy influenced by an LLM and should be rewritten to be more compact.

For example, the link is wrong.

param/parameters.py
```diff
@@ -3487,6 +3502,239 @@ def deserialize(cls, value):
...
+        if self.ordered and isinstance(self.columns, Iterable):
+            if cols != list(self.columns):
+                raise ValueError(
+                    f"{_validate_error_prefix(self)}: provided columns "
+                    f"{cols} must exactly match {self.columns}"
+                )
+        # Row count requires materialising a lazy frame; skip for lazy so
+        # the frame is never implicitly collected.
+        if self.rows is not None and not is_lazy:
+            _length_bounds_check(self, self.rows, nwframe.shape[0], 'row')
```
hoxbro: you can use `count()` to get the row count for a LazyFrame.

``` python
import polars as pl
import narwhals.stable.v2 as nw

N = 100
lf = nw.from_native(pl.LazyFrame(dict(A=["a"] * N, B=["b"] * N)))
cols = list(lf.collect_schema().names())
lf.select(nw.col(cols[0]).count()).collect().item()
```

Image

tests/testdataframelike.py
```diff
@@ -0,0 +1,258 @@
...
+import param
+import pytest
+
+from .utils import check_defaults
+
+try:
+    import narwhals  # noqa: F401
+except ModuleNotFoundError:
+    if os.getenv('PARAM_TEST_NARWHALS', '0') == '1':
+        raise ImportError("PARAM_TEST_NARWHALS=1 but narwhals not available.")
```
hoxbro: I think you should keep this simpler. No need to have an undocumented environment variable.

tests/testdataframelike.py
```diff
@@ -0,0 +1,258 @@
...
+    def test_set_with_ordered_raises(self):
+        with self.assertRaises(ValueError):
+            class P(param.Parameterized):
+                df = param.DataFrameLike(
+                    default=pd.DataFrame({'a': [1]}),
+                    columns={'a'}, ordered=True)
+
+
+@skip_no_pandas
+@skip_no_polars
```
hoxbro: I think we can rewrite this to be polars only.

param/parameters.py
```diff
@@ -3487,6 +3502,239 @@ def deserialize(cls, value):
...
+
+    Serialization is intentionally backend-neutral: ``serialize`` emits a
+    list of records via Narwhals (a lazy frame is collected at this point),
+    and ``deserialize`` reconstructs a ``pandas.DataFrame`` because JSON
+    carries no backend information. Round-tripping therefore does not
+    preserve a non-pandas backend; callers needing another backend can
+    rebuild from the records form.
+    """
+
+    __slots__ = ['rows', 'columns', 'ordered', 'allow_lazy']
```
hoxbro: Maybe we should have something like `stable_version`?

tests/testdataframelike.py
```diff
@@ -0,0 +1,258 @@
...
+    import pyarrow as pa
+except ModuleNotFoundError:
+    pa = None
+
+skip_no_pandas = pytest.mark.skipif(pd is None, reason="pandas not available")
+skip_no_polars = pytest.mark.skipif(pl is None, reason="polars not available")
+skip_no_pyarrow = pytest.mark.skipif(pa is None, reason="pyarrow not available")
+
+
+class TestDataFrameLikeDefaults(unittest.TestCase):
```
hoxbro: I would rewrite these to not inherit from `unittest.TestCase` and use `assert` instead of.

param/parameters.py
```diff
@@ -3311,6 +3317,25 @@ def deserialize(cls, value):
             return numpy.asarray(value)
 
 
+def _length_bounds_check(parameter, bounds, length, name):
```
hoxbro, line 3321: You could also make a `_DataFrameMixin` class which have the shared functionality between them.

hoxbro (commented):
Left some comments.

I'm not entirely sure if we should keep the underlying DataFrame or convert it to a narwhals DataFrame/LazyFrame. 

A user would need to do the conversion in every Parameterized class methods which uses it as they API is very different for the DataFrame APIs.

Also how much AI have you used for this PR?

param/parameters.py
```diff
@@ -3487,6 +3502,239 @@ def deserialize(cls, value):
             return pandas.DataFrame(value)
 
 
+def _get_narwhals():
```
philippjfr: Probably want to move this into `_utils.py`

param/parameters.py
```diff
@@ -3487,6 +3502,239 @@ def deserialize(cls, value):
...
+                f"object that Narwhals recognises, not None."
+            )
+
+        nwframe = self._as_narwhals(val)
+        narwhals = _get_narwhals()
+        is_lazy = isinstance(nwframe, narwhals.LazyFrame)
+        # ``collect_schema().names()`` is the Narwhals-recommended way to read
+        # column names uniformly across eager and lazy frames; ``.columns`` on
+        # a lazy frame triggers a backend schema-resolution warning.
+        cols = list(nwframe.collect_schema().names())
```
philippjfr: Let's avoid collecting the cols unless needed.

param/parameters.py
ghostiee-11: Yeah okay, but as per Andrew GSoC goal planning doc, the param named as DataFrameLike only, can make it to ArrayLike, will change this to ArrayLike only.

param/parameters.py
ghostiee-11: I picked `allow_lazy=False` because the default is "reject lazy", and a positive-named flag reads better at the call site (`allow_lazy=True` is easier to parse than `eager_only=False`). But `eager_only` does mirror the narwhals `from_native(eager_only=...)` kwarg we delegate to, which is a real consistency win. Will flip it to `eager_only=True` (default) unless you'd rather keep the positive-named form.

param/parameters.py
ghostiee-11: Okayy!!

param/parameters.py
ghostiee-11: Will rewrite. The correct URL is https://narwhals-dev.github.io/narwhals/. Compacting the whole docstring to roughly the same length as the existing DataFrame one, dropping the prose.

param/parameters.py
ghostiee-11: Thanks!! will look into this and fix it.

tests/testdataframelike.py
ghostiee-11: Agreed, dropping the env var.

tests/testdataframelike.py
ghostiee-11: Will do, `TestDataFrameLikeAllowLazy`only exercises lazy semantics and Polars LazyFrame is the only backend that lets us assert "no implicit collect" deterministically. Dropping the pandas skip from the class.

param/parameters.py
ghostiee-11: Yes, will switch the import to `import narwhals.stable.v2` as nw and use that throughout.

tests/testdataframelike.py
ghostiee-11: Yeah!! True will change this to assert thing only

param/parameters.py
ghostiee-11: Yeah, Thanks for this feedback, will try this!!

param/parameters.py
ghostiee-11: Sure!! Moving it

param/parameters.py
ghostiee-11: Okay suree

ghostiee-11:
> Left some comments.
> 
> I'm not entirely sure if we should keep the underlying DataFrame or convert it to a narwhals DataFrame/LazyFrame.
> 
> A user would need to do the conversion in every Parameterized class methods which uses it as they API is very different for the DataFrame APIs.
> 
> Also how much AI have you used for this PR?

Thanks for the review.

On native vs. narwhals return: pass-through is intentional. Returning a narwhals.DataFrame would break any consumer that calls `.iloc / .groupby / native` methods on `self.df`, which defeats the point of a drop-in parameter. Authors who want the unified API can do `nw.from_native(self.df)` themselves (one line, the pattern Narwhals recommends). Validation already goes through narwhals internally, so the backend-agnostic surface is used where it matters. Happy to add an opt-in `as_narwhals=True` slot later if users ask for it.

On AI usage: drafting and refactoring assistant. Design is mine and matches what I argued for in #975. Will trim the LLM-flavoured prose in the docstring this round.

param/parameters.py
ghostiee-11: Will take this to follow up pr!!

param/parameters.py
hoxbro: My point with ArrayLike was related to your other PR. I don't want this to be called `Arraylike`.

ghostiee-11: Ohh okay my bad..

ghostiee-11:
Hey @philippjfr, can you review the changes now.

Thankyou!!

philippjfr:
Apologies I missed your message. I will attempt to get this merged once we get the latest patch release out.

ghostiee-11:
No worries :) Thankyou @philippjfr

philippjfr:
> I'm not entirely sure if we should keep the underlying DataFrame or convert it to a narwhals DataFrame/LazyFrame.

I don't think converting by default is a good option, the standard assumption should be that `param` does not convert by default.

That said, a `convert` or `as_narwhals` option seems like a good idea.

philippjfr:
A length check on a `LazyFrame` is not a good idea imo, since it can potentially force a full scan of huge data. We'll have to document that the length checks are not enforced there.

philippjfr (approved):

ghostiee-11:
Thanks for the review @philippjfr and the changes :)

philippjfr (approved):

PR #1148, opened 2026-05-30
Add mypy plugin to handle descriptor __set__ issue

philippjfr:
## Summary

- Add a mypy plugin (`param.mypy_plugin`) that enables correct type checking for class-level parameter assignment (e.g. `MyClass.flag = True`)
- Register the plugin in pyproject.toml under [tool.mypy]

##  Motivation

After implementing descriptor-based typing for Parameters (#1141), mypy reports false positives on class-level assignment to parameters:

```
MyClass.flag = True
# error: Incompatible types in assignment
#   (expression has type "bool", variable has type "Boolean[bool]")
```

This is caused by [mypy #9758](https://github.com/python/mypy/issues/9758): mypy explicitly skips descriptor `__set__` handling for class-level attribute access (`checkmember.py: "__set__ is not called on class objects"`). This is technically correct for
vanilla Python, but param's `ParameterizedMetaclass.__setattr__` routes class-level assignment through the descriptor's `__set__`, so the type checker should accept value-typed assignments. Pyright already
handles this correctly by respecting the metaclass `__setattr__`.

##  Approach

The plugin hooks into `get_class_attribute_hook`. When a class attribute access is an lvalue and the attribute resolves to a `Parameter[_T]` subclass, the plugin returns `_T` (the first type argument) instead of
the raw descriptor type. This correctly reflects the runtime behavior where MyClass.param = value sets the default, not the descriptor.

The plugin correctly handles:
- Direct class assignment (MyClass.flag = True)
- Inherited parameters on subclasses (SubClass.flag = True)
- Type validation (rejects MyClass.flag = "wrong")
- No interference with instance-level access or class-level reads

##  Usage

Users add to their mypy config:

```
[tool.mypy]
plugins = ["param.mypy_plugin"]
```


philippjfr:
Okay, I've now also set up checking of the type asserts using mypy, ensuring that pyright and mypy behaviors match.

param/mypy_plugin.py
```diff
@@ -0,0 +1,99 @@
...
+    if not _is_parameter_type(default_type):
+        return ctx.default_attr_type
+
+    if default_type.args:
+        return default_type.args[0]
+
+    return AnyType(TypeOfAny.special_form)
+
+
+class ParamPlugin(Plugin):
```
hoxbro:
```suggestion
class _ParamPlugin(Plugin):
```

param/mypy_plugin.py
```diff
@@ -0,0 +1,99 @@
...
+        if var is None or var.type is None:
+            return None
+        if not _is_parameter_type(var.type):
+            return None
+
+        return _class_attribute_hook
+
+
+def plugin(version: str) -> type[Plugin]:
+    return ParamPlugin
```
hoxbro:
```suggestion
    return _ParamPlugin
```

param/mypy_plugin.py
```diff
@@ -0,0 +1,99 @@
...
+
+import typing as t
+
+from mypy.nodes import TypeInfo, Var
+from mypy.plugin import AttributeContext, Plugin
+from mypy.types import AnyType, Instance, Type, TypeOfAny, get_proper_type
+
+
+PARAMETER_BASE_FULLNAME = "param.parameterized.Parameter"
+PARAMETERIZED_FULLNAME = "param.parameterized.Parameterized"
```
hoxbro: I would just put these inside the functions.

param/mypy_plugin.py
```diff
@@ -0,0 +1,99 @@
...
+
+Works around mypy issue #9758: metaclass __setattr__ is not consulted
+for class-level assignment to descriptor attributes. Without this plugin,
+``MyClass.flag = True`` on a Boolean parameter raises:
+
+    Incompatible types in assignment
+    (expression has type "bool", variable has type "Boolean[bool]")
+
+The fix: intercept class-attribute access via get_class_attribute_hook and,
+when it's an lvalue, return the Parameter's value type (the first type arg
```
hoxbro, line 11: What is a `lvalue`? Left value of equal sign? I think we should be a bit more human readable without having to also read the code.

hoxbro (approved):
Some minor comments,

The only thing I want to note is this is a public API, so maybe we should make import and variable name private (except `plugin`).

param/mypy_plugin.py
```diff
@@ -0,0 +1,99 @@
...
+of Parameter[_T]) instead of the raw descriptor type.
+
+Usage: add ``plugins = ["param._mypy_plugin"]`` to your mypy configuration.
```
hoxbro:
```suggestion
Usage: add ``plugins = ["param.mypy_plugin"]`` to your mypy configuration.
```

param/mypy_plugin.py
philippjfr: lvalue is standard terminology I think.

maximlt (approved):

param/mypy_plugin.py
philippjfr:
```suggestion
```

param/mypy_plugin.py
```diff
@@ -0,0 +1,99 @@
...
+PARAMETER_BASE_FULLNAME = "param.parameterized.Parameter"
+PARAMETERIZED_FULLNAME = "param.parameterized.Parameterized"
+
+
+def _is_parameter_type(typ: Type) -> bool:
+    proper = get_proper_type(typ)
+    if not isinstance(proper, Instance):
+        return False
+    return any(
+        base.fullname == PARAMETER_BASE_FULLNAME
```
philippjfr:
```suggestion
        base.fullname == "param.parameterized.Parameter"
```

param/mypy_plugin.py
```diff
@@ -0,0 +1,99 @@
...
+        for base in proper.type.mro
+    )
+
+
+def _is_parameterized_class(info: TypeInfo) -> bool:
+    return any(
+        base.fullname == PARAMETERIZED_FULLNAME
```
philippjfr:
```suggestion
        base.fullname == "param.parameterized.Parameterized"
```

PR #1149, opened 2026-06-02
Reorder param.List overloads to allow default type inference

philippjfr:
The `param.List` typing overloads were defined in an order that meant that by default `mypy` would complain that an explicit type declaration was needed. This PR fixes this so that `param.List` resolves to `list[Any]` or `list[Any] | None` by default.

For testing https://github.com/holoviz/param/pull/1148 will enable mypy type assert checks that ensure this works correctly.

### DIFF
--- a/param/_utils.py
+++ b/param/_utils.py
@@ -15,6 +15,8 @@ from textwrap import dedent
 from threading import get_ident
 
 if t.TYPE_CHECKING:
+    import asyncio
+
     from param.parameterized import Parameter
 
     _P = t.ParamSpec("_P")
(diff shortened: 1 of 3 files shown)


hoxbro:
LGTM, do you have an MRE which shows the problem?

hoxbro (approved):

philippjfr:
The https://github.com/holoviz/param/pull/1148 will add enable mypy checks of the assert_types script that previously just failed on:

```python
class ListTypes(param.Parameterized):
    empty_list = param.List()

assert_type(ltypes.empty_list, list[t.Any])
```

PR #1168, opened 2026-08-12
Fix feather serialization tests

philippjfr:
Fixes failing feather tests

### DIFF
--- a/tests/testfiledeserialization.py
+++ b/tests/testfiledeserialization.py
@@ -45,7 +45,12 @@ except ModuleNotFoundError:
         xlsxm = None
 
 try:
+    import pyarrow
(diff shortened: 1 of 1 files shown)


tests/testfiledeserialization.py
```diff
@@ -45,7 +45,12 @@
         xlsxm = None
 
 try:
+    import pyarrow
     import pyarrow.feather as feather
+    # pyarrow.feather.write_feather is deprecated as of pyarrow 24.0.0 in
+    # favor of pyarrow.ipc, which pandas.DataFrame.to_feather does not yet use.
```
hoxbro, line 51: Are you sure about this? Isn't it because Python 3.10 pandas is not the latest?

PR #1171, opened 2026-08-24
Fix thread-unsafe init cache rebuild and unmask errors

philippjfr:
## Problem

Concurrent code that temporarily toggles Parameter slots (the pattern Panel uses in
`panel.util.parameters.edit_readonly`) can crash with a completely misleading error:

```
File "panel/io/state.py", line 428, in _cleanup_busy_counter
    with edit_readonly(self):
File "panel/util/parameters.py", line 47, in edit_readonly
    params = parameterized.param.objects("existing").values()
File "param/parameterized.py", line 3238, in objects
    pdict = self_._cls_parameters
File "param/parameterized.py", line 2570, in __getattr__
    raise AttributeError(f"'{self_.cls.__name__}.param' object has no attribute {attr!r}")
AttributeError: '_state.param' object has no attribute '_cls_parameters'. Did you mean: 'add_parameter'?
```

`_cls_parameters` obviously does exist, which makes this very hard to diagnose.

## Root cause

Two independent issues compound each other.

**1. The init cache rebuild is not thread safe.** Assigning `constant`, `instantiate` or
`default_factory` on a Parameter calls `Parameter._invalidate_init_cache()`, which resets
`params_to_deepcopy`, `params_to_ref` and `params_with_default_factory` on the owner's
`_ClassPrivate` to `None`. The cached-parameters branch of `_cls_parameters` rebuilt those
lists by assigning `[]` to each attribute and then appending to them *through the attribute*
while iterating the parameters. If another thread invalidated the caches part way through
that loop, the next `private.params_to_ref.append(pobj)` hit `None` and raised
`AttributeError: 'NoneType' object has no attribute 'append'`.

**2. `Parameters.__getattr__` masked the real error.** `_cls_parameters` is a property, and
an `AttributeError` raised inside a property getter is indistinguishable from a missing
attribute: CPython clears it and dispatches to `__getattr__`. `Parameters.__getattr__` then
evaluated `attr in self_._cls_parameters`, which on this retry succeeded (the three cache
attributes had already been reassigned by the failed call), found that `'_cls_parameters'`
is not a parameter name, and raised the bogus "has no attribute" error. The real exception
and its traceback were gone, with no `__context__` to follow. Had the retry failed too, the
result would have been a `RecursionError` instead.

Panel triggers this because `_state` is a process-wide singleton whose `busy` counter is
updated from multiple threads, and every `edit_readonly(state)` enter and exit flips
`constant` on every one of its parameters.

## Changes

`Parameter.__setattr__` no longer invalidates the init caches when one of the cache
attributes is set to the value it already has. The caches are derived purely from those
attributes, so a no-op set cannot change them. This removes the invalidation entirely for
the save/restore pattern above, which also avoids repeatedly rebuilding the caches for
every parameter of the class.

`Parameters._cls_parameters` now builds the three lists in locals and only publishes them
once complete, matching what the cold path already did. A concurrent invalidation can no
longer be observed mid-rebuild.

`Parameters.__getattr__` no longer masks errors raised by this class' own descriptors. It
reads the cached class parameters directly from `_param__private.params` rather than going
through the `_cls_parameters` property, so it cannot recurse into the property that sent it
there, and it falls back to re-invoking the descriptor via the new `_invoke_descriptor`
helper when the requested attribute is not a parameter but does exist on the class. The
real error then propagates with its own traceback.

## Notes

Reading `_param__private.params` directly in `__getattr__` also skips the derived cache
checks that the property performs but that `__getattr__` does not need, so `obj.param.<name>`
access gets slightly faster (roughly 490ns to 450ns per access locally).

## Tests

Four tests in `tests/testparameterizedobject.py`, all failing before this change:

- `test_no_op_slot_set_does_not_invalidate_init_cache`
- `test_cls_parameters_rebuild_survives_concurrent_invalidation`, a deterministic
  simulation of an invalidation landing in the middle of the rebuild loop
- `test_cls_parameters_rebuild_is_thread_safe`, four threads running the `edit_readonly`
  save/flip/restore pattern, which reproduced the original error in 6 of 6 runs before the
  fix
- `test_param_namespace_getattr_does_not_mask_descriptor_errors`

Verified end to end against Panel as well: eight threads hammering
`state._add_busy_event`/`_remove_busy_event` reproduce the reported `_cls_parameters` error
within seconds on `main` and run clean with this change.

Note that `edit_readonly` remains unsafe under concurrency for a separate reason (two
overlapping calls can snapshot each other's temporarily relaxed `readonly`/`constant`
values and restore the wrong ones). That needs fixing on the Panel side.

## AI Disclosure

Fix developed with assistance from Claude Opus 5

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -1855,7 +1855,7 @@ class Parameter(_ParameterBase, t.Generic[_T]):
                 pass
 
         super().__setattr__(attribute, value)
-        if is_slot and attribute in _PARAMETER_CACHE_ATTRS and value is not old and value != old:
+        if is_slot and attribute in _PARAMETER_CACHE_ATTRS and value is not old:
             self._invalidate_init_cache()
         if has_watcher and old is not NotImplemented:
             self._trigger_event(attribute, old, value)
@@ -2411,21 +2411,19 @@ class _ParametersRestorer:
             self._restore = {}
 
 
-def _invoke_descriptor(obj: t.Any, attr: str) -> t.Any:
+def _find_descriptor(cls: type, attr: str) -> t.Any:
     """
-    Invoke the descriptor implementing ``attr`` on ``type(obj)`` directly.
+    Return the descriptor implementing ``attr`` on ``cls``, or None.
 
-    An AttributeError raised inside a descriptor (e.g. a property) is
+    Used to invoke a descriptor without going through attribute lookup on the
+    object, since an AttributeError raised inside a descriptor is
     indistinguishable from a missing attribute to the attribute machinery,
-    which clears it and dispatches to ``__getattr__``. Re-invoking the
-    descriptor from there surfaces the real error instead of masking it, and
-    does not recurse since no attribute lookup on ``obj`` is involved.
+    which clears it and dispatches to ``__getattr__``.
     """
-    cls = type(obj)
     for klass in cls.__mro__:
         if attr in klass.__dict__:
-            return klass.__dict__[attr].__get__(obj, cls)
-    raise AttributeError(attr)
+            return klass.__dict__[attr]
+    return None
 
 
 class Parameters:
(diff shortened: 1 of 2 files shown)


param/parameterized.py
```diff
@@ -2562,9 +2579,20 @@ def __getattr__(self_, attr: str) -> t.Any:
...
+        # inside it would be routed back to this method and recurse.
+        params = cls._param__private.params or _invoke_descriptor(self_, '_cls_parameters')
+        if attr in params:
             return self_.__getitem__(attr)
-        elif self_.self is None:
+
+        # Kept off the Parameter lookup path above since it is only relevant
+        # once we know the attribute is not a Parameter
+        if any(attr in klass.__dict__ for klass in type(self_).__mro__):
+            return _invoke_descriptor(self_, attr)
```
hoxbro: Could we try/except here to avoid looking through the `__mro__` twice?

param/parameterized.py
```diff
@@ -2562,9 +2579,20 @@ def __getattr__(self_, attr: str) -> t.Any:
         if cls is None: # Class not initialized
             raise AttributeError
 
-        if attr in self_._cls_parameters:
+        # The cached class parameters are read directly instead of through the
+        # _cls_parameters property, both because the derived caches it also
+        # populates are not needed here and because an AttributeError raised
+        # inside it would be routed back to this method and recurse.
```
hoxbro: This sound a little confusing to me.

param/parameterized.py
```diff
@@ -1855,7 +1855,7 @@ def __setattr__(self, attribute: str, value):
                 pass
 
         super().__setattr__(attribute, value)
-        if is_slot and attribute in _PARAMETER_CACHE_ATTRS:
+        if is_slot and attribute in _PARAMETER_CACHE_ATTRS and value is not old and value != old:
```
hoxbro: Could `value != old` be a problem if it is a numpy array?

tests/testparameterizedobject.py
```diff
@@ -2002,3 +2003,108 @@ class P(param.Parameterized):
...
+            for _ in range(500):
+                if stop.is_set():
+                    return
+                params = list(p.param.objects('existing').values())
+                constants = [po.constant for po in params]
+                for po in params:
+                    po.constant = False
+                for po, constant in zip(params, constants):
+                    po.constant = constant
+        except BaseException as e:
```
hoxbro: Why `BaseException` here?

param/parameterized.py
philippjfr: This is for slots, not for the value but I suppose there's nothing guaranteeing that a Parameter subclass doesn't accept an array for validation or something.

param/parameterized.py
philippjfr: Used a single descriptor lookup for both.

param/parameterized.py
philippjfr: Tried to rephrase.

param/parameterized.py
philippjfr: I think the identity check is sufficient tbh.

tests/testparameterizedobject.py
philippjfr: Fixed.

PR #1174, opened 2026-08-24
Fix rx invalidation finalizer keeping its own node alive

philippjfr:
## Problem

The invalidation-watcher cleanup added in #1156 introduces a leak that is
stricter than the one it fixes: any **function-rooted pipeline**
(`param.rx(fn)(...)`) whose root function can reach the object that owns the
pipeline becomes permanently uncollectable. On 2.4.1 the same object graph is
an ordinary reference cycle that `gc.collect()` reclaims; since #1156 it can
never be freed.

`rx._watch_invalidation` did:

```python
watcher = owner.param._watch(_WeakInvalidator(method), names, precedence=-1)
weakref.finalize(self, _remove_watcher, owner, watcher)
```

`weakref.finalize` stores its callback and arguments **strongly**, in a
process-global registry, until the referent is collected. The CPython docs warn
about exactly this: `func`, `args` and `kwargs` must not own any references to
the referent, directly or indirectly.

Both arguments violate that. `owner` is passed directly, and `watcher.inst` is
also the owner. For a function-rooted pipeline `owner` is the `Wrapper` whose
`object` parameter holds the user's function, so as soon as that function's
closure reaches the object that stores the expression (the natural shape for an
object caching an rx expression built from its own state) the chain closes:

```
finalize registry ─strong─▶ owner (Wrapper) ─▶ fn closure ─▶ owning object ─▶ rx expr
                                                                             = the finalize referent
```

The referent stays strongly reachable from the global registry, so it is never
collected, so the finalizer never fires, so the watcher is never pruned. The
cleanup's own bookkeeping is what prevents the cleanup. The node, everything it
captured, one registry entry and one live watcher per watched source all leak,
without bound, in any long-running process.

Reproduction, param-only:

```python
import gc, weakref, param

class Owner:
    pass

def make_owner():
    owner = Owner()
    def compute(x):
        return x + id(owner) * 0   # closure reaches the owner
    owner.expr = param.rx(compute)(41)
    return owner

owner = make_owner()
ref = weakref.ref(owner)
del owner
gc.collect()
print(f"owner collected: {ref() is None}")
```

```
2.4.1     ->  owner collected: True
2.4.2rc0  ->  owner collected: False
```

## Fix

Hand the finalizer weak references only.

The owner is passed as a `weakref.ref`, and `_remove_watcher` no-ops when it is
dead (if the owner is gone its watchers went with it). The `Watcher` cannot be
referenced weakly, since it subclasses `tuple` and CPython does not allow a
`__weakref__` slot on variable-length types, so it is stored on the
`_WeakInvalidator` and reached through a weakref to *that* instead. The
invalidator is held only by `watcher.fn` in the owner's watcher list, so it
stays alive for exactly as long as the watcher is registered and a dead weakref
correctly means there is nothing left to unwatch.

The cleanup guarantee from #1156 is unchanged: watchers are still pruned as
soon as the derived node is collected, so a long-lived source's watcher list
still does not grow without bound.

Lazy pruning (removing the watcher when the `WeakMethod` resolves to `None`
during dispatch) was considered as an alternative and rejected: the slot-watcher
dispatch path at `Parameter.__setattr__` iterates `self.watchers[attribute]`
in place, so unwatching mid-dispatch would be unsafe there.

## Verification

- The reported reproduction now prints `owner collected: True`, for both a
  closure-rooted and a bound-method-rooted pipeline.
- 200 short-lived function-rooted pipelines: all 200 owners collected, and the
  `weakref.finalize` registry grows by 0.
- 100 short-lived owners against one long-lived `Parameterized` source: all
  collected and the source's watcher list returns to 0. Before the fix the same
  loop left 101 watchers registered and collected none of the owners.
- Both invalidation paths exercised: argument refs (`_invalidate_current`) and a
  `param.bind` root (`_invalidate_obj`). Live expressions still re-evaluate
  correctly while referenced.
- `test_reactive_derived_node_is_garbage_collected` from #1156 still passes;
  full suite: 1525 passed, 4 skipped, 2 xfailed.

New regression tests cover both the single-owner case and the accumulation
case, parametrized over closure-rooted and bound-method-rooted pipelines.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -115,6 +115,8 @@ from ._utils import _to_async_gen, iscoroutinefunction, full_groupby
 if t.TYPE_CHECKING:
     from typing_extensions import Self
 
+    from .parameterized import Watcher
+
     _P = t.ParamSpec('_P')
     _R = t.TypeVar('_R')
     _Y = t.TypeVar('_Y')
@@ -1376,6 +1378,8 @@ class _WeakInvalidator:
 
     __slots__ = ('_ref', '_watcher', '__weakref__')
 
+    _watcher: Watcher | None
+
     def __init__(self, method):
         self._ref = weakref.WeakMethod(method)
         self._watcher = None
@@ -1399,7 +1403,7 @@ def _remove_watcher(
     and cannot be weakly referenced, so it is reached via the invalidator.
     """
     owner, invalidator = owner_ref(), invalidator_ref()
-    if owner is None or invalidator is None:
+    if owner is None or invalidator is None or invalidator._watcher is None:
         return
     try:
         owner.param.unwatch(invalidator._watcher)
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -1374,10 +1374,11 @@ class _WeakInvalidator:
     ``rx._watch_invalidation``).
     """
 
-    __slots__ = ('_ref', '__weakref__')
+    __slots__ = ('_ref', '_watcher', '__weakref__')
 
     def __init__(self, method):
         self._ref = weakref.WeakMethod(method)
+        self._watcher = None
 
     def __call__(self, *events):
         method = self._ref()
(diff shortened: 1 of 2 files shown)


param/reactive.py
```diff
@@ -1374,21 +1374,37 @@ class _WeakInvalidator:
...
+
+    Both arguments are weak references. ``weakref.finalize`` keeps its callback
+    arguments alive in a process-global registry until the referent is
+    collected, so holding either the owner or the ``Watcher`` (whose ``inst``
+    field *is* the owner) strongly would risk making the referent permanently
+    reachable — and therefore uncollectable — whenever the owner can reach it,
+    e.g. a function-rooted pipeline whose root function closes over the object
+    that stores the pipeline. The ``Watcher`` cannot be referenced weakly (it
+    subclasses ``tuple``) so it is reached via the invalidator, which is kept
+    alive by the owner's watcher list for exactly as long as it is registered.
```
hoxbro: Very verbose text. Maybe add type hints and reduce it down.

tests/testreactive.py
```diff
@@ -1056,3 +1056,57 @@ def watcher_count(source):
...
+    return Owner()
+
+
+@pytest.mark.parametrize('factory', [_rx_from_closure, _rx_from_bound_method])
+def test_reactive_function_rooted_node_does_not_pin_its_owner(factory):
+    # weakref.finalize keeps its arguments alive in a process-global registry
+    # until the referent is collected, so the invalidation cleanup must not own
+    # a path back to the node it cleans up. A root function that can reach the
+    # object storing the pipeline used to close that path, making the object
+    # permanently reachable and therefore uncollectable.
```
```suggestion
```

I don't think it adds anything.

philippjfr: Reduced it, this stuff is rather fiddly and obscure so I'm okay with some short comments.

param/reactive.py
philippjfr: Massively reduced and types added.

PR #1175, opened 2026-08-25
Fix superseded async references escaping cancellation

philippjfr:
## Problem

When an asynchronous reference is re-resolved because one of its dependencies changed, param
is supposed to cancel the resolution it supersedes so only the newest one writes a value. In
practice only every *other* superseded resolution was cancelled, so stale coroutines ran to
completion and stale async generators kept streaming into the parameter.

```python
class Source(param.Parameterized):
    x = param.Number(default=0)

class P(param.Parameterized):
    value = param.Parameter(allow_refs=True)

async def slow(i):
    await asyncio.sleep(0.1)
    return i

source = Source()
p = P(value=param.bind(slow, source.param.x))

for i in (1, 2, 3):
    source.x = i
    await asyncio.sleep(0.01)
```

Four resolutions are scheduled (for `x` = 0, 1, 2, 3) and three of them are obsolete before
they finish. Only `0` and `2` were cancelled; `1` and `3` both completed and both assigned to
`p.value`. Which value survived came down to which task happened to finish last, so a slow
early resolution could clobber the result of a fast later one.

The same thing with an async generator is worse, because a superseded generator is not a
one-off stale write but an unbounded second writer: two generators interleave their emissions
into the same parameter for as long as they both keep yielding, and nothing ever stops the
orphan.

## Root cause

`Parameters._async_ref` tracks the resolution that owns a reference in
`_param__private.async_refs[pname]`, and each new resolution cancels whatever it finds there:

```python
running_task = self_.self._param__private.async_refs.get(pname)
if running_task is None:
    self_.self._param__private.async_refs[pname] = current_task
elif current_task is not running_task:
    self_.self._param__private.async_refs[pname].cancel()
```

The superseding task cancels the previous one but never takes ownership of the reference, and
the `finally` block of the task it just cancelled removes the entry it still owns. The
registry therefore ends up empty rather than pointing at the live task:

| task | sees in registry | action | registry afterwards |
| --- | --- | --- | --- |
| 0 | empty | registers itself | task 0 |
| 1 | task 0 | cancels task 0, does not register | empty (task 0's `finally` clears it) |
| 2 | empty | registers itself | task 2 |
| 3 | task 2 | cancels task 2, does not register | empty |

Tasks 1 and 3 are never anybody's `running_task`, so nothing cancels them. The registry
alternates between "owned" and "empty" and half the resolutions slip through.

This did not show up in the existing coverage because `test_async_generator_ref_cancelled`
and `test_generator_ref_cancelled` supersede a ref by *reassigning* it, and reassignment is
handled elsewhere: `Parameter.__set__` pops and cancels the running task synchronously when a
ref is replaced, so the registry is already empty when the new task starts and the new task
registers itself. Only the dependency-change path goes through the broken branch, and only
from the second supersession onwards.

## Changes

`Parameters._async_ref` now takes ownership before cancelling, so the live resolution is
always the registered one:

```python
running_task = self_.self._param__private.async_refs.get(pname)
if running_task is not current_task:
    if running_task is not None:
        running_task.cancel()
    self_.self._param__private.async_refs[pname] = current_task
```

The `finally` cleanup already only clears the entry when it still points at the current task,
which is now exactly right: a cancelled task finds its successor registered and leaves the
entry alone, and only the last surviving resolution clears it. It is also no longer possible
for the registry to hold a task that has already finished, which matters for the two other
places that cancel a reference (`Parameter.__set__` and `Parameters._update_ref`) since both
cancel whatever the registry holds.

That cleanup is additionally guarded with a `pname in async_refs` check. Both the old and new
ownership code leave the reference unregistered when `_async_ref` runs outside a task
(`asyncio.current_task()` returns `None`), and in that case the old `get(pname) is
current_task` test compared `None` to `None`, passed, and raised `KeyError` on the `del`.

## Tests

Two tests in `tests/testrefs.py`, both failing before this change:

- `test_async_ref_cancelled_on_dependency_change` asserts that of four scheduled
  resolutions only the last completes. Before: `assert [1, 3] == [3]`.
- `test_async_generator_ref_cancelled_on_dependency_change` asserts that after two
  supersessions only the newest generator still emits. Before: `assert {1, 2} == {2}`.

The existing reassignment-based cancellation tests still pass, as does the rest of the suite
(1551 passed, 4 skipped, 2 xfailed).

## AI Disclosure

Fixed with the aid of Claude Opus 5.

### DIFF
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2579,10 +2579,11 @@ class Parameters:
 
         ns_type = type(self_)
 
-        # The cached parameters are read from _param__private directly rather
-        # than via the _cls_parameters property. If that property raised an
-        # AttributeError we would be called to handle it and recurse, so it is
-        # only invoked as a descriptor, i.e. without attribute lookup on self_.
+        # Read the cached parameters from _param__private rather than from the
+        # _cls_parameters property: an AttributeError raised by that property is
+        # dispatched to this method, so accessing it as an attribute here would
+        # recurse. When the cache is empty the property is invoked directly as a
+        # descriptor instead, which bypasses that dispatch.
         params = cls._param__private.params
         if not params:
             params = _find_descriptor(ns_type, '_cls_parameters').__get__(self_, ns_type)
@@ -2827,11 +2828,6 @@ class Parameters:
         current_task = asyncio.current_task()
         running_task = self_.self._param__private.async_refs.get(pname)
         if running_task is not current_task:
-            # Take ownership of the reference before cancelling the resolution
-            # we supersede. Cancelling without registering would leave the
-            # registry empty once the cancelled task cleaned up, so the next
-            # resolution would find no owner to cancel and every other task
-            # would escape cancellation and go on writing superseded values.
             if running_task is not None:
                 running_task.cancel()
             self_.self._param__private.async_refs[pname] = current_task
@@ -2849,9 +2845,6 @@ class Parameters:
                         pass
                 self_._settle_async_ref(pname, generation)
         finally:
-            # A resolution that ends without producing a value, because it was
-            # cancelled or raised, still has to settle so the reference is not
-            # left looking like it is perpetually in flight.
             self_._settle_async_ref(pname, generation)
             # Ensure we clean up but only if the task matches the current task,
             # i.e. only the resolution that still owns the reference clears it.
--- a/param/parameterized.py
+++ b/param/parameterized.py
@@ -2742,7 +2742,10 @@ class Parameters:
             if new_val is Skip or new_val is Undefined:
                 continue
             elif is_async:
-                async_executor(partial(self_._async_ref, pname, t.cast("t.Awaitable[t.Any]", new_val)))
+                generation = self_._schedule_async_ref(pname)
+                async_executor(partial(
+                    self_._async_ref, pname, t.cast("t.Awaitable[t.Any]", new_val), generation
+                ))
                 continue
 
             updates[pname] = new_val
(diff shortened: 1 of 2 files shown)


param/parameterized.py
```diff
@@ -2763,39 +2767,90 @@ def _resolve_ref(self_, pobj: Parameter, value: t.Any):
...
+
+        The generation is bumped synchronously, before the task is handed to
+        the executor, so that the reference reads as unsettled from the moment
+        it is superseded rather than only once the task starts running.
+        """
+        if self_.self is None:
+            return 0
+        private = self_.self._param__private
+        generation = private.async_ref_scheduled.get(pname, 0) + 1
+        private.async_ref_scheduled[pname] = generation
```
hoxbro: I think there could be a potential need for a lock here.

param/parameterized.py
```diff
@@ -5681,6 +5740,8 @@ def __init__(
             }
         self.ref_watchers = []
         self.async_refs = {}
+        self.async_ref_scheduled = {}
+        self.async_ref_settled = {}
```
hoxbro: Maybe use `defaultdict(int)`? I think that will make a lot of the code clearer.

tests/testrefs.py
```diff
@@ -304,6 +304,60 @@ async def gen_strings2():
     assert task2 is async_task2
     assert p._param__private.async_refs['string'] is task2
 
+async def test_async_ref_cancelled_on_dependency_change():
+    """
+    A ref re-resolved because a dependency changed must cancel the resolution it
+    supersedes, otherwise a superseded coroutine goes on to write a stale value.
+    """
```
hoxbro: NIT: I don't like docstrings for test. I think the name often is good enough + one line comments if needed.

tests/testrefs.py
```diff
@@ -304,6 +304,60 @@ async def gen_strings2():
     assert task2 is async_task2
     assert p._param__private.async_refs['string'] is task2
 
+async def test_async_ref_cancelled_on_dependency_change():
```
hoxbro, line 307: You have tested that these fails without your changes?

philippjfr: Yes

param/parameterized.py
philippjfr: Done.

tests/testrefs.py
philippjfr: Removed.

param/parameterized.py
philippjfr: Doesn't need an async lock, and thread-safety requires a concerted effort outside the scope of this PR.

PR #1176, opened 2026-08-25
Add `.rx.awaiting` to report whether an expression is still resolving

philippjfr:
## Problem

Since #1173 an `rx` node whose asynchronous operation is in flight reports `param.Undefined`
rather than the value it computed from inputs that have since been superseded. That fixed the
stale value, but it left no way to ask a reactive expression whether anything feeding it is
still in flight. 

## API

`reactive_ops.awaiting` is a read-only property on the `.rx` namespace:

```python
>>> expr = param.rx(1).rx.pipe(slow_double) + 1
>>> expr.rx.value is param.Undefined
True
>>> expr.rx.awaiting
True
```

It is a plain `bool`, not a reactive expression, which makes it the only member of the
namespace that is not composable (apart from `.rx.value`).

## Changes

`reactive_ops.awaiting` walks the graph and asks each node whether it is settling:

```python
return any(node._settling for node in reactive._upstream())
```

`rx._upstream()` yields the node and everything it derives its value from. Inputs reach a node
by three routes and all of them have to be followed, because an operation is only as settled
as the nodes feeding it:

- the `_prev` chain of the pipeline the node belongs to,
- the `_shared` input it was cloned from when a pipeline branches,
- and any `rx` passed as an argument to one of its operations, which
  `_iter_rx` finds by descending the same containers `resolve_value` does (list, tuple, set,
  dict keys and values, slice components).

The walk is iterative with an `id()`-keyed seen-set, because reusing an input clones it and
links the clones through `_shared`, so the graph is a DAG rather than a chain and a naive walk
would revisit nodes.

`rx._settling` is `self._awaiting or self._awaiting_ref`, i.e. either of the two resolution
mechanisms. `rx._awaiting_ref` checks the node's internal parameters against
`Parameters._awaiting_ref` from #1175, covering the case where the asynchronous callable is
held on a parameter rather than applied as an operation. Both are private; `_awaiting` keeps
its existing meaning and `awaiting` is the only public addition.

## Design notes

**Reading it does not schedule anything.** `awaiting` deliberately does not touch `_obj` or
otherwise resolve the node, so an expression whose value has never been requested reports
`False` rather than `True`. "Nothing has been asked for yet" is not the same as "a result is
pending", and a property that started work as a side effect of being read would be a trap in
a loading indicator.

**Generators settle per emission.** A generator or async generator ref settles on each value
it yields rather than on exhaustion, so `awaiting` is `True` only until the next value
arrives. A spinner driven by it stops between emissions instead of spinning until the stream
ends, which is the useful reading for a streaming source.

**`_skipped` was not usable for this.** It conflates "the async result is not in yet" with
"an operation raised `Skip`", which are precisely the two states `awaiting` exists to tell
apart.

## Docs

Adds a `.rx.awaiting` section to the `Reactive_Expressions.ipynb` user guide, contrasting it
with `.rx.updating()`: `updating()` tracks a synchronous computation, while an asynchronous
operation outlives the update that scheduled it. The list of special methods on `.rx` closes
by saying each entry returns a reactive expression, so `awaiting` is mentioned after it with
the difference called out rather than being added to the list.

The example uses a single top-level `await` in one cell so a docs build is deterministic
rather than depending on when the notebook's kernel gets around to a background task.

## AI Disclosure

Written with the assistance of Claude Opus 5.

### DIFF
--- a/doc/user_guide/Reactive_Expressions.ipynb
+++ b/doc/user_guide/Reactive_Expressions.ipynb
@@ -321,27 +321,7 @@
    "cell_type": "markdown",
    "id": "227a230b-98b2-4097-8a6d-798ddb63b74a",
    "metadata": {},
-   "source": [
-    "## Special Methods on `.rx`\n",
-    "\n",
-    "To circumvent the limitations explained above, the `.rx` namespace provides reactive versions of the operations that can't be made reactive through overloading:\n",
-    "\n",
-    "- `.rx.and_`: Reactive version of `and`.\n",
-    "- `.rx.bool`: Reactive version of `bool()`.\n",
-    "- `.rx.in_`: Reactive version of `in`, testing if the value is in the provided collection.\n",
-    "- `.rx.is_`: Reactive version of `is`, testing the object identity against another object.\n",
-    "- `.rx.is_not`: Reactive version of `is not`, testing the absence of object identity with another object.\n",
-    "- `.rx.len`: Reactive version of `len()`, returning the length of the expression\n",
-    "- `.rx.map`: Applies a function to each item in a collection.\n",
-    "- `.rx.not_`: Reactive version of `not`.\n",
-    "- `.rx.or_`: Reactive version of `or`.\n",
-    "- `.rx.pipe`: Applies the given function (with static or reactive arguments) to this object.\n",
-    "- `.rx.updating`: Returns a boolean indicating whether the expression is currently updating.\n",
-    "- `.rx.when`: Generates a new expression that only updates when the provided dependency updates.\n",
-    "- `.rx.where`: Returns either the first or the second argument, depending on the current value of the expression.\n",
-    "\n",
-    "Unlike their corresponding standard Python equivalent, each of these returns a reactive expression that can thus be combined with other reactive expressions to make reactive pipelines."
-   ]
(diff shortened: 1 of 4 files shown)


param/parameters.py
```diff
@@ -2647,7 +2647,10 @@ def __init__(
         object.__setattr__(self, 'check_on_set', check_on_set)
 
         instantiate = params.pop("instantiate", Undefined)
-        params["instantiate"] = False if instantiate is Undefined else instantiate  # pyrefly: ignore[bad-assignment]
+        instantiate_value = False
+        if isinstance(instantiate, bool):
+            instantiate_value = instantiate
+        params["instantiate"] = instantiate_value
```
philippjfr:
```suggestion
        params["instantiate"] = False if instantiate is Undefined else instantiate  # pyrefly: ignore[bad-assignment]
```

doc/user_guide/Reactive_Expressions.ipynb
```diff
@@ -321,27 +321,7 @@
...
... of `is`, testing the object identity against another object.\n- `.rx.is_not`: Reactive version of `is not`, testing the absence of object identity with another object.\n- `.rx.len`: Reactive version of `len()`, returning the length of the expression\n- `.rx.map`: Applies a function to each item in a collection.\n- `.rx.not_`: Reactive version of `not`.\n- `.rx.or_`: Reactive version of `or`.\n- `.rx.pipe`: Applies the given function (with static or reactive arguments) to this object.\n- `.rx.updating`: Returns a boolean indicating whether the expression is currently updating.\n- `.rx.when`: Generates a new expression that only updates when the provided dependency updates.\n- `.rx.where`: Returns either the first or the second argument, depending on the current value of the expression.\n\nUnlike their corresponding standard Python equivalent, each of these returns a reactive expression that can thus be combined with other reactive expressions to make reactive pipelines.\n\nThe namespace also provides `.rx.awaiting`, which reports whether an asynchronous operation in the expression is still resolving. Unlike the methods above it is a plain boolean rather than a reactive expression."
```
hoxbro, line 324: Looking at the raw notebook it has joined up the list to a string. Don't think that is correct. 

Image

param/reactive.py
```diff
@@ -1708,9 +1770,64 @@ def _awaiting(self) -> bool:
...
+        input it was cloned from when a pipeline branches, and any ``rx``
+        passed as an argument to one of its operations.
+        """
+        seen: set[int] = set()
+        stack: list[rx] = [self]
+        while stack:
+            node = stack.pop()
+            if id(node) in seen:
+                continue
+            seen.add(id(node))
```
hoxbro, line 1820:
```suggestion
            node = stack.
            if (id_node := id(node)) in seen:
                continue
            seen.add(id_node)
```

nit

tests/testreactive.py
```diff
@@ -677,6 +677,246 @@ def test_reactive_resolve_recursive(lazy):
...
+    assert expr.rx.value is param.Undefined
+    assert expr.rx.awaiting
+    await async_wait_until(lambda: expr.rx.value == 2)
+    assert not expr.rx.awaiting
+
+async def test_reactive_awaiting_not_scheduled_until_requested():
+    """
+    Reading awaiting must not itself schedule a resolution, so an expression
+    whose value has never been requested is not awaiting.
+    """
```
hoxbro, line 720: Not sure if this docstring adds anything, the function name seem pretty good.

param/reactive.py
```diff
@@ -1708,9 +1770,64 @@ def _awaiting(self) -> bool:
...
+        for p in self._internal_params:
+            owner, name = p.owner, p.name
+            if name is None or not isinstance(owner, Parameterized):
+                continue
+            if owner.param._awaiting_ref(name):
+                return True
+        return False
+
+    @property
+    def _settling(self) -> bool:
```
hoxbro, line 1800: Maybe this is a bit of overkill as it is only used once.

hoxbro (approved):
Small comments

param/reactive.py
hoxbro: Can see it is used more in the stacked PR. So my suggestion would maybe be a better name, hard to understand this is related to awaiting without seeing the function.

PR #1181, opened 2026-09-03
Node-local metadata for rx

philippjfr:
Libraries built on `param.rx` had nowhere to attach per-node state, such as a provenance
record or a cache key. The only thing consumers could do was reassign
`__class__` to a subclass and stash state in `__dict__`, which is broken:
`_clone` reconstructs derived nodes through `type(self)(...)`, so the
*class* propagates to every downstream node while the `__dict__` state does
not, making the subclass a false-positive marker on every derived
expression , code that treats the subclass as "this node carries a record"
misclassifies every node derived from it.

This PR adds a real, node-local metadata slot.

## Changes

**`expr.rx.meta`**

A per-node, mutable `dict` reachable as `expr.rx.meta`. It is not part of a
node's identity: mutating it does not dirty the node, notify watchers, or
affect evaluation. It is not inherited by any node derived from it (through
operators, attribute access, `.rx.pipe`, indexing/branching, etc.) — each
node gets its own, lazily-created mapping. This falls out of `_clone`
essentially for free, since `_meta` is never part of `_kwargs`.

```python
a = param.rx(1)
a.rx.meta['trace'] = 'created at step 1'
b = a + 1
'trace' in b.rx.meta  # False — b is a different node
```

Only available on `rx` nodes; accessing `.rx.meta` on the `.rx` namespace of
a `Parameter` (or other non-`rx` object) raises `AttributeError`.

**Docs**

`rx`'s class docstring now states that subclassing is unsupported —
`_clone`'s `type(self)(...)` reads as an invitation to subclass, but any
instance state on a subclass is silently dropped from every derived node
while the subclass itself keeps propagating — and points at
`register_accessor` (for behaviour) and `.rx.meta` (for state) instead.

## Decisions taken (see plan for the alternatives considered)

- Branch mirrors (`expr[0]`, `expr[1]`) do **not** inherit metadata: strict
  per-node semantics, no exception for "same computation, different node."
- `meta` keys are a plain, unnamespaced `dict`; libraries are expected to
  prefix their own keys by convention.
- `meta` is created lazily (`None` until first accessed) rather than
  allocating a `dict` on every node.


param/reactive.py
```diff
@@ -1615,8 +1657,8 @@ def __init__(
         self._trigger: Trigger | None
         if operation and (iscoroutinefunction(operation['fn']) or inspect.isgeneratorfunction(operation['fn'])):
             self._trigger = Trigger(internal=True)
-            self._current_ = Undefined
-            self._dirty = True  # Otherwise current will be stuck as Undefined.
+            self._current_ = Undefined
+            self._dirty = True  # Otherwise current will be stuck as Undefined.
```
hoxbro, line 1723: Line ending again?

tests/testreactive.py
```diff
@@ -1511,3 +1511,123 @@ def test_reactive_function_rooted_nodes_do_not_accumulate(factory):
...
+    assert watched == []
+
+
+def test_reactive_meta_not_available_on_parameter_rx():
+    p = Parameters()
+    with pytest.raises(AttributeError, match="only available on `rx` nodes"):
+        p.param.integer.rx.meta
+
+
+# register_accessor laziness
```
hoxbro, line 1850: this seems to have sneaked in from another PR?

param/reactive.py
```diff
@@ -1601,6 +1642,7 @@ def __init__(
         self._skipped = False
         self._error_state = None
         self._current_ = _current
+        self._meta: dict[t.Any, t.Any] | None = None
```
hoxbro: Why not define the dict directly here? Is it because it would shared between different nodes or something else?

hoxbro (approved):

param/reactive.py
philippjfr: Seems to have been missed in your cleanup. Will try to fix.

tests/testreactive.py
philippjfr: Indeed. Will fix.

param/reactive.py
philippjfr: Just no reason to allocate one when it's never used. Don't feel strongly though.

hoxbro: That is also valid. Maybe just write that in a comment on the line.

param/reactive.py
philippjfr: Ah, that was on the other PR.

tests/testreactive.py
```diff
@@ -1511,3 +1511,123 @@ def test_reactive_function_rooted_nodes_do_not_accumulate(factory):
...
+    # Still listed once it has actually been instantiated.
+    n.my_accessor
+    assert 'my_accessor' in dir(n)
+
+
+def test_reactive_dir_does_not_list_accessor_when_predicate_fails(clean_accessors):
+    rx.register_accessor('my_accessor', lambda node: node, predicate=lambda value: isinstance(value, str))
+
+    n = rx(1)
+    assert 'my_accessor' not in dir(n)
```
philippjfr, line 1897:
```suggestion
```

PR #1182, opened 2026-09-03
Lazily evaluate registered accessors

philippjfr:
`register_accessor`'s `predicate` was evaluated against `self._current` for every registered accessor, for every `rx` node, at construction time (`rx.__init__`). Because `_current` resolves a dirty, non-lazy node, merely having *any* accessor registered — even one whose predicate always returns `False` — was enough to force every such node to resolve at construction, including scheduling async/coroutine-based work that nobody had asked for yet:

```python
async def body(x):
    calls.append(x)
    await asyncio.sleep(0.01)
    return x * 2

e = param.rx(1).rx.pipe(body)     # never read, never watched
await asyncio.sleep(0.1)

# with no accessor registered:                   body ran 0x
# after register_accessor(..., predicate=...):   body ran 1x
```

The predicate in that measurement returns `False` for everything, so the accessor is never even installed — merely being registered is enough to change the evaluation behaviour of every `rx` node in the process. For a lazily built graph this resolves itself as it's constructed; for an async or threaded node it kicks off work that was never requested. It's also a correctness gap in the other direction: a predicate that answers `False` at construction is never reconsidered, so a node whose value later becomes the type the accessor was meant for never receives it.

## Changes

Accessors are now instantiated lazily, the first time their name is accessed on a node, in `__getattribute__`. The check runs before the fallback path that would otherwise turn an unknown attribute into a `getattr` operation on the wrapped value, so a registered accessor name is never misinterpreted as a pass-through attribute access — including on a node whose current value is `Undefined`.

`__dir__` explicitly lists registered accessor names (evaluating `predicate`, but without instantiating the accessor) so tab-completion and discovery are unaffected.

`register_accessor`'s docstring is updated to document the new timing.

## Backward compatibility

This is a behaviour change in two directions, both fixes, and worth a release note: a predicate with side effects now runs later (on first access) rather than at construction, and a node whose value later becomes the predicate's type now receives the accessor instead of never getting it. The `predicate(value)` signature itself is unchanged.


tests/testreactive.py
```diff
@@ -1511,3 +1511,114 @@ def test_reactive_function_rooted_nodes_do_not_accumulate(factory):
         del owner
     gc.collect()
     assert all(ref() is None for ref in refs)
+# register_accessor laziness
```
philippjfr, line 1849:
```suggestion

```

param/reactive.py
```diff
@@ -2019,14 +2027,22 @@ def _clone(self, operation=None, copy=False, **kwargs) -> Self:
...
     def __dir__(self):
-        current = self._current
+        resolved = self._current
+        current = resolved
         if self._method:
             current = getattr(current, self._method)
         extras = {attr for attr in dir(current) if not attr.startswith('_')}
+        # Registered accessors are instantiated lazily (see __getattribute__),
+        # so list their names explicitly to keep discovery/tab-completion
+        # working for accessors not yet instantiated on this instance.
```
```suggestion
        # Explicitly list registered but uninstantiated accessors
```

param/reactive.py
```diff
@@ -2064,6 +2080,18 @@ def __getattribute__(self, name):
             self._resolve()
             current = self_dict['_current_']
 
+        # Registered accessors (see `register_accessor`) are instantiated
+        # lazily, on first access, rather than at construction time: this
+        # check must run before the expression-building fallback below, so a
+        # registered accessor name is never turned into a `getattr` operation
+        # on the wrapped value instead of returning the accessor.
```
```suggestion
        # Capture uninstantiated accessor access
```

maximlt:
What's up with the narrow text formatting in the OP?

philippjfr:
Fixed, artifact of one of my AI skills, which I've sinced fixed. I'm more confused by the massive diff though. Very odd.

maximlt:
Thanks!

> I'm more confused by the massive diff though. Very odd.

Ah yes 🤔

hoxbro:
It was the line ending.

philippjfr:
It happened after merging a GH code suggestion, how does that change line endings?

hoxbro:
A Windows machine :laughing:

param/reactive.py
```diff
@@ -2181,6 +2197,18 @@ def __getattribute__(self, name):
...
```
```suggestion
        # Capture uninstantiated accessor access
```

param/reactive.py
```diff
@@ -2136,14 +2144,22 @@ def _clone(self, operation=None, copy=False, **kwargs) -> Self:
...
     def __dir__(self):
-        current = self._current
+        resolved = self._current
+        current = resolved
         if self._method:
             current = getattr(current, self._method)
         extras = {attr for attr in dir(current) if not attr.startswith('_')}
+        # Registered accessors are instantiated lazily (see __getattribute__),
+        # so list their names explicitly to keep discovery/tab-completion
+        # working for accessors not yet instantiated on this instance.
```
```suggestion
        # Explicitly list registered but uninstantiated accessors
```

philippjfr:
Okay, the line ending saga is resolved.

tests/testreactive.py
```diff
@@ -1895,3 +1894,65 @@ def test_reactive_dir_does_not_list_accessor_when_predicate_fails(clean_accessor
...
+    rx.register_accessor(
+        'my_accessor', lambda node: 'matched', predicate=lambda value: isinstance(value, str)
+    )
+
+    n = rx(1)
+    with pytest.raises(AttributeError):
+        n.my_accessor
+
+    n.rx.value = 'a string now'
+    assert n.my_accessor == 'matched'
```
```suggestion
    assert n.my_accessor == 'matched'

```

tests/testreactive.py
```diff
@@ -1895,3 +1894,65 @@ def test_reactive_dir_does_not_list_accessor_when_predicate_fails(clean_accessor
 
     n = rx(1)
     assert 'my_accessor' not in dir(n)
+
+    def test_reactive_register_accessor_predicate_not_evaluated_at_construction(clean_accessors):
```
```suggestion
def test_reactive_register_accessor_predicate_not_evaluated_at_construction(clean_accessors):
```

param/reactive.py
```diff
@@ -1602,6 +1602,14 @@ def register_accessor(
         """
         Register an accessor that extends ``rx`` with custom behavior.
 
+        The accessor is not instantiated when the node is constructed. It is
+        instantiated lazily, the first time the accessor's name is accessed on
+        a given node, at which point ``predicate`` is evaluated against the
+        node's current value. This means registering an accessor never
+        forces a node to resolve, and a node whose value only later becomes
+        the type ``predicate`` looks for can still pick up the accessor at
+        that point, on the next access of its name.
```
hoxbro: This reads like it is explaining based on the knowledge of what used to happen. I think it would be clearer if this is rewritten to explain what happens not without any prior knowledge.

param/reactive.py
```diff
@@ -1610,6 +1618,9 @@ def register_accessor(
           A callable that will return the accessor namespace object
           given the ``rx`` object it is registered on.
         predicate: Callable[[Any], bool] | None
+          Called with the node's current value the first time ``name`` is
+          accessed on that node; the accessor is only instantiated if this
+          returns True (or is None).
```
hoxbro: I think there is some ambiguity around the parenthesis.


```suggestion
          accessed on that node; the accessor is only instantiated if a
          callable returns True or if ``predicate`` is None.
```

param/reactive.py
```suggestion
        Accessors are instantiated lazily the first time it is accessed on a given
        node. If a ``predicate`` is provided it is evaluated against the node's
        current value at that point in time. If it does not evaluate as true the first
        time, e.g. because the value has not yet settled, it may still be created on
        subsequent accesses.
```

hoxbro (approved):

PR #1185, opened 2026-09-07
Add opt-in error propagation to rx

philippjfr:
Reactive expressions currently have two ways to represent the absence of a usable result, and neither is suitable for displaying a failure in a rendered application. `Skip` and `Undefined` suppress an update, so downstream operations are not called and the previous value is retained. An ordinary exception is stored in `_error_state` and re-raised when the graph is read, which means it cannot be passed to a downstream node or rendered where the value would normally appear. Consumers therefore need exception handling around graph reads and separate logic to preserve the dependency graph while showing an error.

This PR addresses this by adding an opt-in error channel to reactive expressions. `rx` accepts `error_mode="propagate"`, which converts exceptions raised while evaluating an expression into falsy `param.Error` values; downstream operations pass those values through unchanged, while `process_failures=True` on `.rx.pipe` lets a handler inspect the failure. The default `error_mode="raise"` preserves the existing behavior of caching and re-raising exceptions.

```python
import param

def divide(value):
    return 10 / value

value = param.rx(0, error_mode="propagate")
result = value.rx.pipe(divide)

isinstance(result.rx.value, param.Error)  # True
str(result.rx.value)                      # "division by zero"
result.rx.error is result.rx.value        # True

handled = result.rx.pipe(
    lambda error: f"failed: {error.exception}",
    process_failures=True,
)
handled.rx.value  # "failed: division by zero"
```

`param.ReactiveError` carries the original exception and the node that failed, formats as the exception message, is falsy, and delegates attribute access and subscripting to the exception. It makes failure an ordinary graph value: a pane or other consumer can render the failure at the point where the result would have appeared, and a downstream operation can either preserve it or explicitly handle it. This is an additive API change with no migration required for existing expressions.

## AI Disclosure

Tool & Model: Kilo Code + openai/gpt-5.6-luna
Usage: Used to inspect the existing reactive evaluation paths, implement the error channel, add tests, run the targeted test suite, and draft this description.

## Checklist

- [x] Added documentation

### DIFF
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -1615,7 +1615,7 @@ class rx:
     def __init__(
         self, obj=None, operation=None, fn=None, depth=0, method=None, prev=None, lazy=False,
         _shared_obj=None, _current=None, _wrapper=None, _shared=None, error_mode='raise',
-        _error_mode=None, **kwargs
+        **kwargs
     ):
         # _init is used to prevent to __getattribute__ to execute its
         # specialized code.
@@ -1640,8 +1640,8 @@ class rx:
         self._finished_generation = 0
         self._skipped = False
         self._error_state = None
-        self._error_mode = error_mode if _error_mode is None else _error_mode
-        if self._error_mode not in ('raise', 'propagate'):
+        self._error_mode = error_mode
+        if error_mode not in ('raise', 'propagate'):
             raise ValueError("error_mode must be either 'raise' or 'propagate'")
         self._current_ = _current
         # _shared is used for branching rx pipelines where we clone the input.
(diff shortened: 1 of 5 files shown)


param/reactive.py
```diff
@@ -125,6 +125,29 @@
 logger = logging.getLogger(__name__)
 
 
+class Error:
+    """A reactive value representing an error in an expression."""
```
hoxbro: Is this doc string right? I don't think this is a reactive value.

Also not sure about the generic of the name. Maybe `PipelineError`?

param/reactive.py
```diff
@@ -1906,6 +1949,12 @@ def _resolve(self):
         elif self._dirty or self._root._dirty_obj:
             try:
                 obj = self._obj if self._prev is None else self._prev._resolve()
+                operation = self._operation
+                if isinstance(obj, Error) and not (operation or {}).get('process_failures'):
+                    self._current_ = obj
```
hoxbro, line 1954: Will this cause a cyclic reference?

hoxbro (commented):
I think the original post need to explain why this feature is needed.

param/reactive.py
philippjfr: Perhaps poor phrasing. The naming is in keeping with the simple `Skip`, but I'll try to come up with a better yet still simple name.

hoxbro (approved):

PR #1190, opened 2026-09-11
Implement rx input overrides

philippjfr:
When building complex reactive graphs you sometimes want to test assumptions or temporarily freeze specific values. To achieve this you have to fully rebuild the `rx` graph from scratch right now, this PR instead offers a mechanism to register `overrides` on a node that supersede the actual inputs.

This PR adds `.rx.overrides`, a mutable mapping on a node that addresses the inputs it was wired with, by keyword name or by positional index:

```python
import param

fx = param.rx(2)
value = param.rx(100).rx.pipe(lambda price, fx: price * fx, fx=fx)
value.rx.value  # 200

value.rx.overrides['fx'] = 1
value.rx.value  # 100

fx.rx.value = 5
value.rx.value  # still 100, the override masks the update

value.rx.overrides['fx'] = None  # None unmasks
value.rx.value  # 500
```

The interposition is local to this node's consumption of the input, so every other consumer of `fx` keeps seeing the live value and is neither notified nor recomputed. Because the override replaces the input before it is resolved, it lands ahead of every guard the node applies to its inputs, and masks an input that failed, one that has not resolved yet and one that skipped, without the node having to be wired with `process_failures=True`:

```python
broken = param.rx(0, error_mode="propagate").rx.pipe(lambda divisor: 1 / divisor)
total = param.rx(7).rx.pipe(lambda value, extra: value + extra, extra=broken)
isinstance(total.rx.value, param.ReactiveError)  # True

total.rx.overrides['extra'] = 5
total.rx.value  # 12
```

Overrides are node-local, like `.rx.meta`: a derived node has its own, empty mapping, and only an input the node was wired with can be overridden, so an unknown key raises `KeyError` naming the inputs that exist. Overriding is a plain replacement; a consumer that needs to merge an override into the live value computes the merged value itself.

## AI Disclosure

Tool & Model: Kilo Code + amazon-bedrock/us.anthropic.claude-opus-5
Usage: Used to spike the invalidation options and measure them, implement the feature, add tests, run the test suite and type checkers, and draft this description.

## Checklist

- [x] Added documentation

### DIFF
--- a/doc/reference/param/reactive.md
+++ b/doc/reference/param/reactive.md
@@ -86,6 +86,27 @@ value.rx.overrides['fx'] = None  # None unmasks, so it cannot be an override val
 value.rx.value  # 500
 ```
 
+An override can also be set to a reference — a `Parameter`, another expression, a
+bound function, a widget — and the node then follows it, so a UI control can drive
+one node's view of an input without touching the input itself:
+
+```python
+class Scenario(param.Parameterized):
+    fx = param.Number(default=3)
+
+scenario = Scenario()
+value.rx.overrides['fx'] = scenario.param.fx
+value.rx.value  # 300
+
+scenario.fx = 4
+value.rx.value  # 400
+```
+
+A reference that resolves to `None` unmasks the input, and one that has not
+resolved yet skips, because the override stands in for the input rather than
+alongside it. Replacing or unmasking an override stops following the reference it
+was set to.
+
 The override replaces the input ahead of every guard the node applies to its
 inputs, so it also masks an input that failed, one that has not resolved yet and
 one that skipped, without the node having to be wired with `process_failures=True`:
--- a/doc/reference/param/reactive.md
+++ b/doc/reference/param/reactive.md
@@ -59,7 +59,7 @@ for an expression without a failure.
 `.rx.overrides` makes one node compute as if one of its inputs held a different
 value, addressed by keyword name or positional index. It does not set the input, so
 every other consumer of that input keeps seeing the live value, and only this node
-and the nodes reading its result are invalidated. Setting it back to `None` unmasks
+and the nodes reading its result are invalidated. Deleting the key unmasks
 the original input:
 
 ```python
@@ -75,7 +75,7 @@ value.rx.value  # 100
 fx.rx.value = 5
 value.rx.value  # still 100, the override masks the update
 
-value.rx.overrides['fx'] = None
+del value.rx.overrides['fx']
 value.rx.value  # 500
 ```
 
@@ -92,7 +92,8 @@ value.rx.value  # 300
 ```
 
 The override stands in for the input and is resolved in its place, ahead of the
-guards the input would have faced.
+guards the input would have faced. Any value masks the input, including `None`, so
+an override following a reference keeps masking when that reference holds `None`.
 
 ```{eval-rst}
 .. autosummary::
@@ -101,6 +102,7 @@ guards the input would have faced.
    rx
    reactive_ops
    ReactiveError
+   InputOverrides
 ```
 
 These methods and properties are available under the `.rx` namespace of reactive expressions ({py:class}`rx`):
--- a/doc/reference/param/reactive.md
+++ b/doc/reference/param/reactive.md
@@ -56,75 +56,43 @@ for an expression without a failure.
 
 ### Overriding an input
 
-`.rx.overrides` interposes a value on one of a node's inputs, addressed by keyword
-name or by positional index. The node then computes as if that input held the
-given value:
+`.rx.overrides` makes one node compute as if one of its inputs held a different
+value, addressed by keyword name or positional index. It does not set the input, so
+every other consumer of that input keeps seeing the live value, and only this node
+and the nodes reading its result are invalidated. Setting it back to `None` unmasks
+the original input:
 
 ```python
 import param
 
 fx = param.rx(2)
-price = param.rx(100)
-value = price.rx.pipe(lambda price, fx: price * fx, fx=fx)
+value = param.rx(100).rx.pipe(lambda price, fx: price * fx, fx=fx)
 value.rx.value  # 200
 
 value.rx.overrides['fx'] = 1
 value.rx.value  # 100
-```
 
-This is not the same as setting `fx`: the interposition is local to this node's
-consumption of the input, so every other consumer of `fx` keeps seeing the live
-value, and only this node and the nodes reading its result are invalidated. While
-an input is overridden its updates are masked, and unmasking picks up whatever
-value arrived in the meantime:
-
-```python
 fx.rx.value = 5
-value.rx.value  # still 100
+value.rx.value  # still 100, the override masks the update
 
-value.rx.overrides['fx'] = None  # None unmasks, so it cannot be an override value
+value.rx.overrides['fx'] = None
 value.rx.value  # 500
 ```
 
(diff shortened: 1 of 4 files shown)


param/reactive.py
```diff
@@ -219,6 +220,104 @@ class NestedResolver(Resolver):
...
+        key = self._resolve_key(key)
+        overrides = self._overrides
+        if not overrides or key not in overrides:
+            raise KeyError(key)
+        del overrides[key]
+        self._unwatch(key)
+        self._node._invalidate_overrides()
+
+    def __iter__(self) -> Iterator[t.Any]:
+        return iter(dict(self._overrides or {}))
```
hoxbro, line 312: is dict needed here?

param/reactive.py
```diff
@@ -219,6 +220,104 @@ class NestedResolver(Resolver):
...
+
+    def __len__(self) -> int:
+        return len(self._overrides or {})
+
+    def __repr__(self) -> str:
+        return f"overrides({dict(self._overrides or {})!r})"
```
hoxbro, line 318: is dict needed here?

param/reactive.py
```diff
@@ -671,6 +770,66 @@ def meta(self) -> dict[t.Any, t.Any]:
...
+        The override stands in for the input and is resolved in its place, ahead
+        of the guards the input would have faced, i.e. it even overrides input
+        holding errors or undefined values.
+
+        Overrides are node-local, like ``.rx.meta``: a derived node
+        (``expr + 1``) has its own, empty mapping.
+
+        Returns
+        -------
+        _InputOverrides
```
hoxbro, line 810: Not sure if this is then really private if it returned, but not a deal breaker.

doc/reference/param/reactive.md
```diff
@@ -54,6 +54,46 @@ handled.rx.value  # "failed: division by zero"
 The `.rx.error` property returns the current `ReactiveError` or exception, and `None`
 for an expression without a failure.
 
+### Overriding an input
+
+`.rx.overrides` makes one node compute as if one of its inputs held a different
+value, addressed by keyword name or positional index. It does not set the input, so
+every other consumer of that input keeps seeing the live value, and only this node
+and the nodes reading its result are invalidated. Setting it back to `None` unmasks
```
hoxbro, line 62: What if you want the value to be `None`?

hoxbro (commented):
Is it expected for an overridden parameter to still trigger events? I would imagine if you have made something static it does not make sense to trigger something downstream, could be wrong though. 

``` python
import param

a = param.rx(2)
b = param.rx(10) * a 
c = param.rx(10) * b

b_events, c_events = [], []
b.rx.watch(b_events.append)
c.rx.watch(c_events.append)

b.rx.overrides[0] = 1 

a.rx.value = 5
a.rx.value = 9

print("b:", b.rx.value, "c:", c.rx.value) # b: 10 c: 100
print("b_events:", b_events)  # b_events: [10, 10, 10]
print("c_events:", c_events)  # c_events: [100, 100, 100]
```

param/reactive.py
philippjfr: Agreed, fixed.

doc/reference/param/reactive.md
philippjfr: Agreed, fixed.

param/ipython.py
```diff
@@ -400,8 +405,17 @@ def update_handle(*args, **kwargs):
             cb = cb_reactive_cb
         try:
             obj = cb()
+        except Skip:
+            obj = None
+        except TypeError:
+            # E.g. a partially bound function that still needs a positional
+            # argument: not displayable, so tell IPython to fall back to the
+            # plain repr instead of crashing the cell.
```
philippjfr:
```suggestion
            # Fall back to the plain repr for partially bound functions
```

param/reactive.py
philippjfr: It wasn't.

param/reactive.py
philippjfr: It wasn't.

param/reactive.py
```diff
@@ -2315,15 +2614,26 @@ def _transform_output(self, obj):
 
     @property
     def _callback(self) -> Callable[..., t.Any]:
-        params = self._params
+        # Consuming this node as a display callback is exactly the kind of
+        # reference consumption `_rx_transform` already handles correctly, so
+        # this mirrors it: the override channel makes a masked input's change
+        # visible, and the `_skipped` check stops a still-resolving async node
+        # from being rendered with its previous, stale value.
```
```suggestion
```

philippjfr:
Likely will follow up here by exposing public methods to access both the upstream and the downstream nodes.

param/ipython.py
```diff
@@ -400,8 +405,15 @@ def update_handle(*args, **kwargs):
             cb = cb_reactive_cb
         try:
             obj = cb()
+        except Skip:
+            obj = None
+        except TypeError:
+            # Fall back to the plain repr for partially bound functions
+            raise NotImplementedError
```
hoxbro, line 412: Not sure I follow this.

param/reactive.py
```diff
@@ -1362,10 +1521,15 @@ def watch(self, fn=None, onlychanged=True, queued=False, precedence=0):
         self._watch(fn, onlychanged=onlychanged, queued=queued, precedence=precedence)
 
     def _watch(self, fn=None, onlychanged=True, queued=False, precedence=0):
+        _unset = object()
+        last = [_unset]
```
hoxbro: maybe just use `nonlocal` here?

param/ipython.py
philippjfr: IPython will fall back to a string repr if the custom repr raises. Will update the comment.

hoxbro: But this has nothing to do with IPython, that has been moved into a separated `try/except`.

philippjfr: It does, but the other try/except can be removed.

param/ipython.py
```diff
@@ -400,8 +405,15 @@ def update_handle(*args, **kwargs):
...
+        else:
             if obj is Undefined:
                 obj = None
+        try:
             handle = display(obj, display_id=uuid.uuid4().hex)
         except TypeError:
             raise NotImplementedError
```
philippjfr:
```suggestion
        handle = display(obj, display_id=uuid.uuid4().hex)
```

param/reactive.py
philippjfr: Done.

hoxbro: There were one other place where this pattern happened.

param/reactive.py
```diff
@@ -2404,15 +2703,21 @@ def _transform_output(self, obj):
...
                 out = getattr(out, self._method)
-            return self._transform_output(out)
-        if params:
-            return bind(evaluate, *params)
-        return evaluate
+            out = self._transform_output(out)
+            if last[0] is not _unset and Comparator.is_equal(out, last[0]):
+                raise Skip
+            last[0] = out
+            return out
```
```suggestion
        last = _unset = object()
        def evaluate(*args, **kwargs):
            nonlocal last
            out = self._current
            if self._skipped:
                raise Skip
            if self._method:
                out = getattr(out, self._method)
            out = self._transform_output(out)
            if last is not _unset and Comparator.is_equal(out, last):
                raise Skip
            last = out
            return out
```

param/reactive.py
philippjfr: Done.

hoxbro (approved):

PR #1191, opened 2026-09-11
Implement .rx.stale

philippjfr:
This PR exposes the internal `rx._dirty` invalidation flag as `.rx.stale`, a plain boolean reporting whether an expression has produced a value for its *current* inputs. It is `True` from the moment an input changes until the expression recomputes, and before the first evaluation of an expression whose value has never been requested, which makes it possible to ask "is the value I can read up to date?" without resolving the expression to find out.

The property is deliberately a superset of the raw `_dirty` flag rather than a one-to-one alias, because `_dirty` is cleared when an async operation is *scheduled*, not when it produces a value. A node awaiting an async result therefore has `_dirty == False` while still holding a value computed from superseded inputs, so a flag mapped straight onto `_dirty` would report "up to date" at exactly the moment the value is not. `.rx.stale` reports `node._dirty or node._root._dirty_obj or node._settling` across the whole graph returned by `_upstream()`, matching how `.rx.awaiting` already aggregates, so an operation downstream of a stale one is itself stale and a bound function whose parameters changed (`_dirty_obj`) counts too. `.rx.awaiting` remains the async-specific subcase, giving the reader `stale and not awaiting` for "the next read recomputes synchronously". Reading the property neither resolves nor schedules anything, so it reports `True` for a never-evaluated expression instead of quietly evaluating it, and it is always `False` when accessed on a parameter rather than an expression.

Synchronous behavior:

```python
import param

number = param.rx(1)
expr = number + 1

assert expr.rx.stale          # never evaluated
assert expr.rx.value == 2
assert not expr.rx.stale

number.rx.value = 5
assert expr.rx.stale          # input changed, not recomputed yet
assert expr.rx.value == 6
assert not expr.rx.stale
```

Asynchronous behavior, where scheduling alone does not clear it:

```python
import asyncio
import param

async def slow_double(value):
    await asyncio.sleep(0.1)
    return value * 2

async def main():
    expr = param.rx(1).rx.pipe(slow_double) + 1

    assert expr.rx.value is param.Undefined
    assert expr.rx.stale and expr.rx.awaiting   # scheduled, no value yet

    await asyncio.sleep(0.2)
    assert expr.rx.value == 3
    assert not expr.rx.stale and not expr.rx.awaiting

asyncio.run(main())
```

An operation that raises `Skip` deliberately keeps its previous value, so it is not reported as stale once the skip has been processed.

No behavior changes, no new dependencies, and no breaking changes: this only adds a read-only property on the `.rx` namespace.

## AI Disclosure

Tool & Model: Kilo Code + Claude Opus

Usage: Explored the invalidation machinery, implemented the property, wrote the tests and the user guide section, and verified the sync/async state transitions empirically before settling on the semantics.

## Checklist

- [x] Added documentation

### DIFF
--- a/param/reactive.py
+++ b/param/reactive.py
@@ -70,7 +70,9 @@ the total number of reactive instances in a pipeline.
 Instances also track their dependencies to ensure accurate updates:
 - `_method`: Temporarily stores the method or attribute accessed (e.g., `'head'`
   in `dfi.head()`).
-- `_dirty`: Indicates whether the current value needs re-computation.
+- `_dirty`: Indicates whether the current value needs re-computation. Exposed
+  publicly, together with the state of the nodes feeding a node, as
+  `.rx.stale`.
 - `_current`: Stores the result of the most recent computation.
 
 Benefits and Use Cases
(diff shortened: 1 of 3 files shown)


param/reactive.py
```diff
@@ -902,6 +904,57 @@ def awaiting(self) -> builtins.bool:
...
+        has been scheduled, since scheduling is not the same as producing a
+        value, so ``stale`` and not ``.rx.awaiting`` means the next request for
+        the value recomputes it synchronously.
+
+        The whole graph feeding the expression is considered, but only inputs
+        that actually invalidate it, so an expression gated with ``.rx.when`` is
+        not stale until its gate fires. Reading this neither resolves nor
+        schedules anything. On a parameter or a bound function it is always
+        ``False``, since reading their value evaluates rather than returning a
+        cached one.
```
```suggestion
```

tests/testreactive.py
```diff
@@ -976,6 +976,269 @@ def gen():
     assert expr.rx.awaiting
     await async_wait_until(lambda: not expr.rx.awaiting)
 
+def test_reactive_stale_until_first_evaluation():
+    expr = rx(1) + 2
+
+    assert expr.rx.stale
```
hoxbro, line 982: Does it make sense to say something is `stale` if has never been evaluated?

param/reactive.py
```diff
@@ -902,6 +904,50 @@ def awaiting(self) -> builtins.bool:
             return False
         return any(node._settling for node in reactive._upstream())
 
+    @property
+    def stale(self) -> builtins.bool:
```
hoxbro, line 908: What would it report for a generator?

tests/testreactive.py
philippjfr: Yes, for sure. `stale` means does the internal value reflect all current inputs without triggering any new computation. Never evaluated counts.

param/reactive.py
philippjfr: Very good question! I'll check what the behavior is, and write out my thinking simultaneously. My assumption is that: If the latest yield is reflected it's not stale, if it's not reflected it is. As soon as something upstream yields, it's stale again.

philippjfr: That is the behavior, thankfully! Though checking into it uncovered a separate bug, I'll push a new PR for.

hoxbro (approved):

PR #1193, opened 2026-09-14
Fix/rx async error channel

philippjfr:
Addresses three related issues with the `rx` error channel implementation:

- `_resolve_async` now honors `error_mode='propagate'`, mirroring the synchronous `_resolve branch`: a failure resolves to a `ReactiveError` instead of poisoning `_error_state`, and the method returns immediately instead of letting `trigger.param.trigger('value')` re-raise inside the exception handler.
- `param.bind` gains a process_failures keyword (default False, matching .rx.pipe/_apply_operator). When a bound argument resolves to a `ReactiveError`, the wrapped function is not called and the `ReactiveError` is returned/yielded unchanged, closing the gap where a failure crossing from an `rx` pipeline into a plain bound function would previously crash the consumer instead of propagating as a value.
- `rx` nodes accept an optional `label`, inherited through `_clone` (so it survives `.rx.pipe` and operator overloads), and ReactiveError gains a matching label property that reads it lazily from the originating node so a consumer can identify which computation failed without dereferencing the node itself.

## AI Disclosure

Usage:

## Checklist

- [ ] Added documentation


param/reactive.py
```diff
@@ -135,6 +135,12 @@ def __init__(self, exception: Exception, node=None):
         self.exception = exception
         self.node = weakref.ref(node) if node is not None else None
 
+    @property
+    def label(self):
```
hoxbro: Maybe a type hint here?


```suggestion
    def label(self) -> str | None:
```

tests/testreactive.py
```diff
@@ -1831,6 +1912,54 @@ def gen(value):
     with pytest.raises(RuntimeError, match='boom'):
         async_rx.rx.value
 
+async def test_reactive_async_error_propagate_mode_no_unhandled_exception():
+    # Regression test: error_mode='propagate' on the async path used to be
+    # ignored, and the watcher dispatch inside the exception handler used to
+    # re-raise the same exception, producing an unretrieved task exception.
```
hoxbro: I don' think this is an regression. We haven't had an official release with this yet.

hoxbro (commented):
Maybe add some documentation around `bind(..., process_failures=...)`?

Note that not that i see is a deal break it is a tiny breaking change to no more being able to bind to a function with a keyword argument called `process_failures`. We should likely mention that in a release note/documentation somewhere.  Maybe also add it to docstring that this was added in 2.5.0. 

Maybe also add some documentation around `rx(..., label=...)`?

hoxbro:
Also should label have a getter/setter functionality?

philippjfr:
Pushed `c391d32` addressing the review feedback:

- Added a `.rx.label` getter/setter on `reactive_ops`, mirroring `.rx.value`, so `label` can be read and set on a live `rx` expression, not just passed at construction.
- Added a `str | None` return type hint to `ReactiveError.label`.
- Documented `bind(..., process_failures=...)` and `rx(..., label=...)` in their docstrings with `.. versionadded:: 2.5.0` notes, plus a `Notes` section on `bind` calling out the small breaking change (a wrapped function's own `process_failures` kwarg is no longer forwarded). Also updated `doc/reference/param/reactive.md` with the same info and added `~reactive_ops.label` to the autosummary.
- Reworded the comment in `test_reactive_async_error_propagate_mode_no_unhandled_exception` to not call it a "regression test", since this behavior predates any release.

Added tests for the new `.rx.label` getter/setter (including the `AttributeError` on non-`rx` reactive references). Full suite (`pytest tests`) and `ruff check` both pass.

hoxbro (approved):

tests/testreactive.py
```diff
@@ -1831,6 +1931,54 @@ def gen(value):
     with pytest.raises(RuntimeError, match='boom'):
         async_rx.rx.value
 
+async def test_reactive_async_error_propagate_mode_no_unhandled_exception():
+    # error_mode='propagate' on the async path used to be ignored, and the
+    # watcher dispatch inside the exception handler used to re-raise the same
+    # exception, producing an unretrieved task exception.
```
```suggestion
```

PR #1195, opened 2026-09-14
Make .rx.updating() span an asynchronous wait

philippjfr:
`.rx.updating()` only flipped True/False around watcher dispatch, which fires once a node has produced a value. For a synchronous recompute that happens to be the same instant the value is ready, but for an asynchronous operation (e.g. `.rx.pipe(coroutine_fn)`) scheduling the work returns immediately, so both watchers fired together only once the whole thing had already settled. `updating()` therefore reported False for the entire wait and flashed true right as the result arrived, the opposite of a useful loading indicator.

In this PR `rx` gains a small per-node notification list (`_settle_watchers` / `_watch_settle_change` / `_notify_settle_change`) fired from `_lazy_resolve` at the point `_settling` flips `True`. `updating()` subscribes to every node in the tracked expressions (structurally static) `_upstream()` set once, at construction, and sets the wrapper `True` as soon as any of them schedules async work. 

Also seeds `updating()`s initial value from the upstream graphs current settling state, so attaching it to an expression that is already mid-flight reports the correct state immediately.

## AI Disclosure

Tool & Model: Kilo Code using Claude Sonnet 5
Usage:

## Checklist

- [ ] Added documentation


param/reactive.py
```diff
@@ -982,9 +986,20 @@ def updating(self) -> 'rx':
...
+        initial = any(node._settling for node in upstream)
+        wrapper = t.cast('Callable', Wrapper)(object=initial)
+
         self._watch(lambda e: wrapper.param.update(object=True), precedence=-999)
         self._watch(lambda e: wrapper.param.update(object=False), precedence=999)
+
+        # The watchers above only fire once a value is produced, which misses
+        # an asynchronous wait entirely, so also flip on scheduling.
+        for node in upstream:
+            node._watch_settle_change(lambda: wrapper.param.update(object=True))
```
hoxbro: These do not seem to ever be removed? So it will likely keep `wrapper` alive after the updating expression and keep on accumulate on every call causing dead wrappers being updated forever.   

We likely need a weak reference to the wrapper to avoid this, and then drop it from `self._settle_watchers` when the wrapper is gc'd.

hoxbro (changes requested):

param/reactive.py
philippjfr: Will fix. I'll probably have to work on a general improvement for the other watchers as well but will do that in a separate PR.

hoxbro (approved):

PR #1196, opened 2026-09-14
Ensure a sync child of an async node does not resolve it at construction

philippjfr:
Synchronous (non-lazy) nodes eagerly evaluate their ancestors when attempting an attribute access to validate the chain of operations. The check for whether the node is synchronous used `self._is_async`, but `self` is the node an accessor or `.rx.pipe` is being resolved on, not necessarily the async part of the chain. A plain sync operation sitting on top of an unread, unwatched async ancestor was still resolved eagerly through the `self._current` property, which scheduled the async body purely as a side effect of constructing a further node off the sync child.

This PR checks whether anything upstream (not just `self`) is async instead, and skip the eager resolve in that case; a genuinely sync chain still resolves eagerly here as before, preserving the existing non-lazy semantics.

## AI Disclosure

Usage:

## Checklist

- [ ] Added documentation


param/reactive.py
```diff
@@ -2285,9 +2285,10 @@ def _clone(self, operation=None, copy=False, **kwargs) -> Self:
         operation = operation or self._operation
         depth = self._depth + 1
         if copy:
-            # Do not trigger resolve via self._current since result is discarded
-            # by the cloned nodes constructor anyway
-            current = self._current_ if self._is_async else self._current
+            if any(node._is_async for node in self._upstream()):
```
hoxbro, line 2288: what if `self` is async and no upstream is? 


```suggestion
            if self._is_async or any(node._is_async for node in self._upstream()):
```

philippjfr: `self` is included in `_upstream`.

PR #1197, opened 2026-09-15
Improve rx lifecycle and add .rx.dispose

philippjfr:
An `rx` pipeline had no way to clean up all the watchers it had installed and watchers created with `.rx.watch` could not be individually cleaned up. Lastly `.rx.resolve` caused an outright leak where the node would be kept alive as long as it's owner was kept alive.

## .rx.dispose

Added `.rx.dispose()` to detach a node from its pipeline immediately rather than waiting for GC. Each node now records, weakly, which other nodes read its value via `_prev`, `_shared`, or an operation argument; `dispose()` releases the node's own invalidation watchers and, if it was the only reader of an input, disposes that input too, stopping at any input still read elsewhere in the graph:

  ```python
  import param

  a = param.rx(1)
  b = a + 1
  b.rx.value  # 2
  b.rx.dispose()  # releases b's watchers, and a's too, since b was its only reader
  ```

## .rx.watch returns watchers

`.rx.watch()` now returns the `Watcher`(s) it registers instead of discarding them, and a new `.rx.unwatch()` undoes exactly that call, leaving any other watcher on the same expression untouched. Getting the `Watcher` out required `depends.py`'s `_depends_impl` to stop discarding the return value of `owner.param.watch(...)` and expose it as `_watchers` on the function it returns, which `bind()` now also carries over for its generator and async-generator wrapping paths.

  ```python
  expr = param.rx(1)
  watcher = expr.rx.watch(print)
  expr.rx.unwatch(watcher)  # print is no longer called on updates
  ```

## Fix `.rx.resolve` leak

Fixed a leak in `.rx.resolve()`: the `Resolver` it constructs registered a strong, bound-method watcher on the owner of any reference it followed, so that owner kept the resolver, and everything it was resolving, alive for as long as the owner existed, independent of whether the resolved expression was still referenced. `Resolver._update_refs` now uses the same weakly-bound watcher plus `weakref.finalize` pattern `rx._watch_invalidation` already uses for the main pipeline.

## AI Disclosure

Tool & Model: Kilo Code + Claude Sonnet 5

Usage: Used to design and implement the reader-tracking and dispose mechanism, find and fix the `.rx.resolve()` leak, add tests, and run the test suite and type checkers.

## Checklist

- [ ] Added documentation


param/reactive.py
```diff
@@ -213,13 +213,24 @@ def _resolve_object(self):
...
-            (w.inst or w.cls).param.unwatch(w)
+        """
+        Watch every ``Parameterized`` a resolved reference points to.
+
+        Weakly bound (see ``_WeakInvalidator``): the reference's owner is
+        typically longer-lived than this resolver, so a strong watcher would
+        keep it, and whatever it was resolving, alive for as long as the
+        owner exists.
+        """
+        for finalizer in self._watchers:
```
hoxbro: Just want to be sure this shouldn't be:

```suggestion
        for finalizer in self._finalizers:
```

param/reactive.py
```diff
@@ -1779,6 +1850,17 @@ class rx:
...
     _settle_watchers: list[weakref.ref] | None = None
 
+    # Weak refs to rx nodes that derive their value from this one, via
+    # `_prev`, `_shared`, or as an operation argument. Lets `_dispose()`
+    # tell whether an input is still read by something else.
+    _readers: list[weakref.ref] | None = None
+
+    # Finalizer handles for this node's own invalidation watchers, kept so
+    # `_dispose()` can release them immediately instead of waiting for GC.
+    _finalizers: list[weakref.finalize] | None = None
```
hoxbro: Compact these comments.

param/reactive.py
```diff
@@ -2171,12 +2302,17 @@ def _watch_invalidation(self, owner, method, names):
...
         invalidator = _WeakInvalidator(method)
         invalidator._watcher = owner.param._watch(invalidator, names, precedence=-1)
-        weakref.finalize(
+        finalizer = weakref.finalize(
             self, _remove_watcher, weakref.ref(owner), weakref.ref(invalidator)
         )
+        finalizers = self._finalizers
+        if finalizers is None:
+            finalizers = self._finalizers = []
+        finalizers.append(finalizer)
```
hoxbro, line 2651:
```suggestion
        if self._finalizers is None:
            self._finalizers = []
        self._finalizers.append(finalizer)
```

This seems simpler. Also don' think it should have significant performance problem, but please correct me if I'm wrong.

tests/testreactive.py
```diff
@@ -1574,6 +1574,124 @@ def test_reactive_upstream_walk_terminates_on_reused_input():
...
+def test_reactive_dispose_is_idempotent():
+    a = rx(1)
+    b = a + 1
+    assert b.rx.value == 2
+    b.rx.dispose()
+    b.rx.dispose()  # does not raise, and does not touch a a second time
+    assert _watcher_count(a) == 0
+
+def test_reactive_dispose_is_not_a_plain_attribute():
+    """A plain ``.dispose`` would shadow a same-named method on the wrapped value."""
```
hoxbro: This seems to be written with prior knowledge.

tests/testreactive.py
```diff
@@ -1574,6 +1574,124 @@ def test_reactive_upstream_walk_terminates_on_reused_input():
...
+
+    wrapped = Wrapped()
+    expr = rx(wrapped)
+
+    result = expr.dispose()
+    assert result.rx.value == 'disposed'
+    assert wrapped.disposed is True
+
+def test_reactive_readers_pruned_on_gc():
+    """A reader entry is a weak ref, so it does not keep a dropped reader alive."""
```
hoxbro: remove

PR #1198, opened 2026-09-15
Add current_node() so an operation body can reach the node being resolved

philippjfr:
A function passed to an `rx` operation (`.rx.pipe`, `bind`, an operator, etc.) is called with only its resolved value and arguments, so it has no way to attach data to the specific node its computation belongs to. We did recently add `.rx.meta`, which can be used to store provenance information, partial results and more. Because this is only reachable from outside a node but not from inside the function computing its value it can't easily be used for the purpose it was designed for (provenance and caching partial results)

This PR adds `param.current_node()`, which returns the node currently being resolved (or `None` outside of any operation body), backed by a `contextvars.ContextVar` set for the duration of `_eval_operation`'s call and the following `_lazy_resolve` scheduling call in `_resolve`.

This makes it possible for the function to assign provenance metadata in its body:

```python
def kernel(price, fx):
    node = param.current_node()
    if node is not None:
        node.rx.meta['trace'] = {'fx_used': fx}
    return price * fx

expr = param.rx(price).rx.pipe(kernel, fx=fx_input)
expr.rx.value          # value, unaffected
expr.rx.meta['trace']  # {'fx_used': ...}, written by the body itself
```

## AI Disclosure

Usage: implementation, tests, and this description were drafted with Claude Sonnet 5 in Kilo Code, following a design discussion I directed on the two shapes considered and the decision between them.

## Checklist

- [ ] Added documentation


param/reactive.py
```diff
@@ -1709,6 +1710,57 @@ async def _close_stale(obj):
...
 # See https://peps.python.org/pep-0673/
 
+# Set for the duration of an operation body's execution (sync, coroutine,
+# generator, or async generator alike) so the body can look up the node it is
+# being evaluated for without it being passed as a positional argument. A
+# contextvars.ContextVar rather than a thread-local because two concurrently
+# running async nodes on the same event loop must not see each other's node;
+# each asyncio Task gets its own copy of the context, which gives that
+# isolation for free. See `rx._resolve` for where this is set, and
+# `current_node` below for the accessor.
```
hoxbro: Maybe just have this be part of _current_node docstring?

param/reactive.py
```diff
@@ -1709,6 +1710,57 @@ async def _close_stale(obj):
...
+    straight through to ``.rx.meta``.
+
+    This is a plain module-level function rather than an attribute of
+    :class:`rx`, deliberately: `rx.__getattribute__` forwards any name
+    absent from `super().__dir__()` into a `getattr` operation on the
+    wrapped value, so an attribute added to the class would permanently
+    shadow a same-named attribute on whatever object a node happens to
+    wrap (see the `rx-node-local-metadata.md` plan, which is why `.rx.meta`
+    lives under the `.rx` namespace rather than directly on the expression,
+    the one name `__getattribute__` short-circuits unconditionally).
```
hoxbro: I don't think this is relevant. This is a technical reason, but from an user perspective it does not make any sense.

hoxbro (commented):
some minor stuff related to comments.

param/reactive.py
philippjfr: Thought I had cleaned this up, sorry.

param/reactive.py
```diff
@@ -2343,16 +2382,23 @@ def _resolve(self):
...
+                                # Handle case where async call is resolved synchronously
+                                # e.g. when there is no running event loop
+                                self._skipped = False
+                                self._dirty = False
+                                return self._current_
+                            obj = Skip
+                    finally:
+                        # Safe to reset even for an async body: _lazy_resolve
+                        # above already created its Task, which copied this
+                        # context var's value for its own (deferred) run.
```
```suggestion
                        # Safe to reset even for async task as context is copied on schedule
```

param/reactive.py
```diff
@@ -2539,22 +2588,38 @@ def stale():
                 self._finished_generation = generation
                 trigger.param.trigger('value')
             elif inspect.isasyncgen(obj):
-                async for val in obj:
+                # Manually driven (rather than `async for`) so the node is only
+                # current while the generator body is actually advancing, not
+                # while `trigger` below runs and may synchronously invoke
+                # watchers.
```
```suggestion
                # Manually drive generator (as opposed to async for) to ensure
                # current_node is only set while generator body is advancing
```

hoxbro (approved):

PR #1199, opened 2026-09-15
Add .rx.collect for partial results across several async inputs

philippjfr:
An operation with more than one reactive argument does not run until every argument has
settled, even if some settle much sooner than others:

```python
a = param.rx(1).rx.pipe(slow, delay=0.02)   # settles quickly
b = param.rx(2).rx.pipe(slow, delay=0.2)    # settles much later
combined = a.rx.pipe(lambda x, y: (x, y), y=b)
combined.rx.value                            # Undefined until *both* settle
```

`_eval_operation` resolves every argument in order and raises `Skip` for the whole node if any
one is still settling, which is correct for a function that genuinely needs every argument but
leaves no way to build a "two of three sources have answered" view on top of `rx` itself.

This PR introduces `.rx.collect(*args, error_mode='raise', **kwargs)` which combines the expression it's called on with
further inputs into one expression holding whichever have settled, keyed the way `.rx.pipe`
addresses arguments: the calling expression at position `0`, like `.rx.pipe`'s function receives
it as its first argument, then further positional arguments at `1`, `2`, ..., then keyword ones
by name. It does not call a function: the result is a mapping that grows as each input settles
rather than waiting for the slowest one. `.rx.awaiting` on the collected node is `True` while
any input is unsettled, via the same `_upstream()` walk the rest of `.rx.awaiting` already
uses. 

`error_mode='propagate'` resolves a failing input to a `ReactiveError` at its own key
instead of failing every other key too. Implemented as an operation whose `fn` is a
`_collect_marker` sentinel, resolved by its own path (`rx._eval_collect`) instead of the
ordinary one, since a `Skip` from an unsettled argument means "not ready yet" here rather than
failing the whole node.

## Docs

Adds `.rx.collect` to the "Special Methods on `.rx`" bullet list and a `#### .rx.collect(*args,
**kwargs)` section to the `Reactive_Expressions.ipynb` user guide, after `.rx.stale` and before
`.rx.when`: partial settlement across two async inputs (one of them the calling expression at
key `0`), and `error_mode='propagate'` holding a `ReactiveError` at one key. Verified by
executing the full notebook end to end.

## AI Disclosure

Usage:


tests/testreactive.py
```diff
@@ -2845,3 +2845,91 @@ def accessor(node):
     assert n.my_accessor == 'accessor-value'
     assert n.my_accessor == 'accessor-value'
     assert installed_for == [n]
+
+
+# rx.gather
```
hoxbro: I think it would make sense to move these test (and other) into classes but not something to do in this PR.

param/reactive.py
```diff
@@ -2736,6 +2786,36 @@ def _eval_operation(self, obj, operation):
...
+        result = dict(self._current_) if isinstance(self._current_, dict) else {}
+        settled = bool(result)
+        for key, arg in chain(enumerate(operation['args']), operation['kwargs'].items()):
+            if any(ref._settling for ref in _iter_rx(arg)):
+                continue
+            try:
+                val = resolve_value(arg)
+            except Skip:
+                # Not a failure, just not ready yet; keep this key's previous
+                # value (if any) and try again on the next recompute.
```
hoxbro: Isn't most of this redundant?

doc/user_guide/Reactive_Expressions.ipynb
```diff
@@ -651,6 +651,78 @@
    "execution_count": null,
    "outputs": []
   },
+  {
+   "cell_type": "markdown",
+   "id": "8001f682",
+   "source": [
+    "#### `rx.gather(*args, **kwargs)`\n",
```
hoxbro: Maybe just


```suggestion
    "#### `rx.gather`\n",
```

param/reactive.py
philippjfr: The comment or the code?

hoxbro: comment

philippjfr:
Decided to rename to `.rx.collect`, `gather` sounded like `asyncio.gather` which has opposite semantics (where it waits on all the inputs before returning).

hoxbro:
> Decided to rename to .rx.collect

Would that cause a problem if the object is a LazyFrame (Narwhals/Polars)?

philippjfr:
No, that's why we have it on the `.rx` namespace object.

hoxbro (approved):

PR #1200, opened 2026-09-15
Allow walking the rx graph via public methods

philippjfr:
An `rx` node currently has no way public way to walk its dependency graphs. `_upstream()` (what a node computes its value from) already existed but was private, and there was no way to walk the reverse direction to find what reads a node at all. When building a graph store on top of `rx` it is often necessary to walk the graph both ways to handle graph lifecycle concerns, e.g. knowing what reads an `rx` nodes value before destroying it.

Additionally `rx` was also unhashable, an side-effect of overriding `__eq__` (used to build a comparison expression). This tripped Python's rule that a class defining `__eq__` without `__hash__` becomes unhashable, a rule meant to stop equality and hashing from disagreeing, something that cannot happen here since == never returns a bool to disagree with in the first place. 

This PR closes both gaps: `.rx.upstream()` and `.rx.dependents()` expose the existing forward walk and a new reverse counterpart as public, transitive graph traversals, and `__hash__ = object.__hash__` restores identity hashing so a node can be used as a dict key or set member, which is exactly what deduplicating the nodes dependents()/upstream() return requires.

## AI Disclosure

Usage:

## Checklist

- [x] Added documentation


param/reactive.py
```diff
@@ -1137,6 +1137,65 @@ def stale(self) -> builtins.bool:
...
+        >>> a = param.rx(1)
+        >>> b = a.rx.pipe(lambda x, y: x + y, y=param.rx(2))
+        >>> a in set(b.rx.upstream())
+        True
+        """
+        reactive = self._reactive
+        if not isinstance(reactive, rx):
+            return
+        upstream = reactive._upstream()
+        next(upstream, None)  # Skip `reactive` itself.
```
```suggestion
        next(upstream, None)  # Skip itself.
```

param/reactive.py
```diff
@@ -1137,6 +1137,65 @@ def stale(self) -> builtins.bool:
...
+        >>> a = param.rx(1)
+        >>> b = a.rx.pipe(lambda x: x + 1)
+        >>> b in set(a.rx.dependents())
+        True
+        """
+        reactive = self._reactive
+        if not isinstance(reactive, rx):
+            return
+        dependents = reactive._dependents()
+        next(dependents, None)  # Skip `reactive` itself.
```
```suggestion
        next(dependents, None)  # Skip itself.
```

param/reactive.py
```diff
@@ -1952,6 +2011,12 @@ class rx:
     # Weak refs to targets notified when this node schedules async work.
     _settle_watchers: list[weakref.ref] | None = None
 
+    # `__eq__` (below) builds a comparison expression rather than a bool, so
+    # it can never be inconsistent with a hash, unlike the usual case Python's
+    # __eq__-without-__hash__ rule protects against. Restored explicitly so a
+    # node can be used as a `dict` key or `set` member.
```
hoxbro: compact.

param/reactive.py
```diff
@@ -2224,15 +2288,39 @@ def _settling(self) -> bool:
         """Whether this node is waiting on an asynchronous result of its own."""
         return self._awaiting or self._awaiting_ref
 
-    def _upstream(self) -> Iterator[t.Any]:
+    def _direct_inputs(self) -> Iterator[t.Any]:
```
hoxbro, line 2399: Just want to be sure this matches up with the same `_direct_input` as done in #1197.

hoxbro (commented):
I also think the names should match, some options at the top of my head:

- `upstream`/`downstream`
- `parents`/`children`
- `dependencies` / `dependents`
- `providers` / `consumers`

hoxbro (approved):

PR #1202, opened 2026-09-15
fix: Unduplicate same name in `watch`

hoxbro:
## Description

See https://github.com/holoviz/panel/pull/8764

## Checklist

- [x] Added documentation


philippjfr:
Could see an argument that we should warn here since this is obviously a silly thing to do.

hoxbro:
Was thinking about it but then we would need to update Panel to avoid the warning for the `Matplotlib`/`PNG` pane.

philippjfr:
Let's merge for now.

PR #1203, opened 2026-09-16
Review, fixes and cleanup for new rx features

philippjfr:
This PR fixes four issues found reviewing the recent run of independent `rx` PRs (input overrides, `.rx.collect()`, lifecycle), where each landed cleanly on its own but combined behavior wasn't checked:

- `_eval_collect` never consulted `operation['overrides']`, so `.rx.overrides` silently accepted a key on a `.rx.collect()` node, masked that key's invalidation in `_live_params()`, but had no effect on the resolved value, freezing reactivity for that key with no visible error.
- The `.rx.awaiting` docstring claimed `.rx.value` reports `Undefined` while awaiting; `_resolve()` actually keeps returning the last computed value on every await after the first, so the docstring now matches the real (and tested) behavior.
- `InputOverrides._unwatch` called `_remove_watcher` directly instead of firing the `weakref.finalize` `_watch_invalidation` registered, so repeated overrides on the same key left dead finalizer objects piling up in `node._finalizers` indefinitely.
- `_live_params_cache` was keyed on a single process-wide `rx._masking_generation` counter, so overriding one node invalidated the cache of every other unrelated `rx` node in the process; it's now cleared only for the affected node and its downstream readers, via `_invalidate_overrides`'s existing reader walk.

```python
import param

a, b = param.rx(1), param.rx(2)
collected = a.rx.collect(b)
collected.rx.overrides[1] = 999
collected.rx.value            # before: {0: 1, 1: 2} (override ignored); after: {0: 1, 1: 999}
b.rx.value = 5
collected.rx.value            # before: unchanged forever, even after del override; after: reacts normally
```

## AI Disclosure

Tool & Model: Kilo + Claude Sonnet 5
Usage: Used to review the cumulative `rx` diff, localize and fix the four issues, and draft this PR description.

## Checklist


param/reactive.py
```diff
@@ -282,9 +282,20 @@ def _resolve_key(self, key: t.Any) -> t.Any:
...
-        for owner, invalidator in (operation.get('override_watchers') or {}).pop(key, ()):
-            _remove_watcher(weakref.ref(owner), weakref.ref(invalidator))
+        node = self._node
+        operation = t.cast('dict', node._operation)
+        finalizers = node._finalizers
+        for finalizer in (operation.get('override_watchers') or {}).pop(key, ()):
+            # Firing the finalizer now (instead of calling `_remove_watcher`
+            # out-of-band) both unwatches the source and marks it dead, so it
+            # is safe to drop from `_finalizers` instead of accumulating there
+            # until the node itself is disposed or collected.
```
hoxbro: update to not reference previous behavior.

hoxbro (commented):
I assume the test are just updating to classes, so haven't reviewed it.

param/reactive.py
```suggestion
            # Firing the finalizer now both unwatches the source and
            # marks it dead, so it is safe to drop from `_finalizers`.
```

philippjfr:
> I assume the test are just updating to classes, so haven't reviewed it.

Correct.

PR #1204, opened 2026-09-16
Add param.batch() to globally defer watcher notification

philippjfr:
Setting two independent sources that feed one computation, one statement after another, delivers an intermediate value that mixes the new value of one with the stale value of the other, because each set fires its own watcher pass immediately:

```python
import param

class A(param.Parameterized):
    x = param.Number(default=1)

class B(param.Parameterized):
    y = param.Number(default=10)

a, b = A(), B()
calls = []
param.bind(lambda x, y: calls.append(x + y), a.param.x, b.param.y, watch=True)

a.x = 2
b.y = 20
calls  # [12, 22] -- 12 crosses the new x with the stale y
```

The existing `batch_call_watchers` already defers this for one object named up front, and composes correctly when nested over every object involved, but the caller has to enumerate them. That does not work for a sheet of overrides applied from a store, or a dynamic rx graph rebuild, where the full set of objects touched is only known once the operation runs.

This PR adds `param.batch()`, a context manager that defers notification for every `Parameterized` object touched inside it, discovered as it is touched rather than declared up front, and coalesces a callback registered on more than one of them into a single call:

```python
with param.batch():
    a.x = 2
    b.y = 20
calls  # [22] -- one call, seeing both settled, not one per source touched
```

`param.bind` registers the identical callback separately on `a` and `b`, so a drain that simply deferred each object independently would still call it twice, both times correctly but redundantly. `batch()` avoids that by gathering every touched object's queued events before executing anything, grouping watchers across all of them by callback identity, and running each group once with the union of its events.

It generalizes past `.rx`: it also closes the same gap for a sheet of `.rx.overrides` spanning several different nodes' override channels, for `@param.depends(watch=True)`, and for plain `.param.watch()` on more than one parameter, since all of them share the same underlying gate.

This is an additive API change with no migration required for existing code.

## AI Disclosure

Tool & Model: Kilo Code + anthropic/claude-sonnet-5
Usage: Used to investigate the gap, spike and verify the mechanism, implement `batch()`, add tests across `Parameterized` and `rx`, update the user guide notebook, run the test suite and linter, draft this description, and address review feedback (fixing the permanently-batched-object bug, lost queued/cascading events, one-watcher-failure dropping other objects' events, nondeterministic ordering, over-aggressive coalescing, and a stale transaction reference from a task started inside the block, each with a regression test verified against the reported repro).

## Checklist

- [x] Added documentation


param/parameterized.py
```diff
@@ -426,6 +427,98 @@ def batch_call_watchers(parameterized):
...
+    are the very same callback registered on more than one of them (how
+    `bind()`/`depends()` wire one callback across several source objects)
+    so the callback runs once for the transaction, not once per object.
+    """
+    groups: dict[int, list] = {}
+    for obj in objects:
+        self_ = obj.param
+        if not self_._events:
+            continue
+        event_dict = OrderedDict([((event.name, event.what), event) for event in self_._events])
```
hoxbro: Why OrderedDict?

param/parameterized.py
```diff
@@ -426,6 +427,98 @@ def batch_call_watchers(parameterized):
...
+                for name in watcher.parameter_names
+                if (name, watcher.what) in event_dict
+            ]
+            if not events:
+                continue
+            group = groups.setdefault(id(watcher.fn), [watcher, [], []])
+            group[1].append(obj)
+            group[2].extend(events)
+
+    for watcher, owners, events in sorted(groups.values(), key=lambda g: g[0].precedence):
```
hoxbro: Correct it should it run lower precedence first?

tests/testreactive.py
```diff
@@ -3470,6 +3470,54 @@ def test_reactive_override_propagates_without_any_parameters():
...
+    combined.rx.watch(calls.append)
+    calls.clear()
+
+    node1.rx.overrides['x'] = 100
+    node2.rx.overrides['y'] = 200
+
+    assert calls == [1100, 3000]
+
+
+def test_reactive_override_sheet_across_nodes_inside_batch():
```
hoxbro: These could be parameterized with a `contextlib.nullcontext`

tests/testreactive.py
```diff
@@ -3470,6 +3470,54 @@ def test_reactive_override_propagates_without_any_parameters():
...
+    from param.parameterized import batch
```
hoxbro: why not at top?

param/parameterized.py
philippjfr: Right, no need for ordering here, since it's only ever looked up by key, never iterated. Switched to a plain `dict`.

param/parameterized.py
philippjfr: Yes, sorted() is ascending so lower precedence already runs first here, matching the single-object `_batch_call_watchers` (parameterized.py:3449). What was missing: the representative watcher for a merged group was whichever owner got iterated first, not necessarily the lowest-precedence one if the same callback was registered with different precedences on different owners. Fixed by tracking the lowest-precedence watcher seen as the group builds, and added a test for the mismatched-precedence case.

tests/testreactive.py
philippjfr: Merged the two into one `pytest.mark.parametrize` test using `contextlib.nullcontext()` vs `batch()` as the context manager.

tests/testreactive.py
philippjfr: Moved to the top-level import, now that the two tests are merged there's just one `batch` import shared by the module.

tests/testreactive.py
```diff
@@ -3470,6 +3471,37 @@ def test_reactive_override_propagates_without_any_parameters():
     assert (n.rx.value, derived.rx.value) == (10, 11)
 
 
+@pytest.mark.parametrize(('ctx', 'expected'), [
+    (contextlib.nullcontext(), [1100, 3000]),
+    (batch(), [3000]),
```
```suggestion
    (contextlib.nullcontext, [1100, 3000]),
    (batch, [3000]),
```

And then call them inside function

tests/testreactive.py
```diff
@@ -3470,6 +3471,37 @@ def test_reactive_override_propagates_without_any_parameters():
...
+    combined = rx(lambda n1, n2: n1 + n2)(n1=node1, n2=node2)
+    calls = []
+    combined.rx.watch(calls.append)
+    calls.clear()
+
+    with ctx:
+        node1.rx.overrides['x'] = 100
+        node2.rx.overrides['y'] = 200
+
+    assert calls == expected
```
```suggestion
@pytest.mark.parametrize(('ctx', 'expected'), [
    (contextlib.nullcontext, [1100, 3000]),
    (batch, [3000]),
], ids=['without_batch', 'inside_batch'])
def test_reactive_override_sheet_across_nodes(ctx, expected):
    """
    ``node1`` and ``node2`` have distinct override channels, so setting one
    then the other, without ``batch()``, delivers a value crossing the
    first with the second's stale input; ``batch()`` avoids it.
    """
    class A(param.Parameterized):
        x = param.Number(default=1)

    class B(param.Parameterized):
        y = param.Number(default=10)

    a, b = A(), B()
    node1 = rx(lambda x: x * 10)(x=a.param.x)
    node2 = rx(lambda y: y * 10)(y=b.param.y)
    combined = rx(lambda n1, n2: n1 + n2)(n1=node1, n2=node2)
    calls = []
    combined.rx.watch(calls.append)
    calls.clear()

    with ctx():
        node1.rx.overrides['x'] = 100
        node2.rx.overrides['y'] = 200

    assert calls == expected
```

param/parameterized.py
philippjfr: Sorry, AI nonsense, dict itself is obviously ordered. I think it picked up on other internal OrderedDict usage.

param/parameterized.py
```diff
@@ -2474,6 +2680,12 @@ def __init__(self_, cls: type[Parameterized], self: Parameterized | None = None)
 
     @property
     def _BATCH_WATCH(self_):
+        # Deliberately just the raw per-instance flag, with no knowledge
+        # of a `batch()` transaction: every save/restore site (this
+        # property's own callers) needs the real prior value to restore,
+        # not a value that lies while a transaction happens to be open.
+        # `_is_batched()` below is what call sites that need to know
+        # "should this defer right now" should use instead.
```
philippjfr:
```suggestion
```

tests/testwatch.py
```diff
@@ -1191,3 +1193,514 @@ class P(param.Parameterized):
...
+        a, b = A(), B()
+        calls = []
+        param.bind(lambda x, y: calls.append(x + y), a.param.x, b.param.y, watch=True)
+
+        with batch():
+            a.x = 2
+            b.y = 20
+
+        # param.bind registers the same callback separately on each source,
+        # but batch() coalesces it into one call, seeing both settled.
```
philippjfr:
```suggestion
```

tests/testwatch.py
```diff
@@ -1191,3 +1193,514 @@ class P(param.Parameterized):
...
+        calls = []
+        shared = lambda *events: calls.append(len(events))
+        a.param.watch(shared, 'x')
+        b.param.watch(shared, 'y', onlychanged=False)
+
+        with batch():
+            a.x = 2
+            b.y = 20
+
+        # Two separate calls, not one merged call with two events.
```
philippjfr:
```suggestion
```

param/parameterized.py
```diff
@@ -422,10 +423,215 @@ def batch_call_watchers(parameterized):
...
+    Every object's events are gathered before any watcher runs, so one
+    raising does not stop the rest; the first exception is re-raised
+    once every group has been attempted.
+    """
+    groups: dict[t.Any, _WatcherGroup] = {}
+    for obj in objects:
+        self_ = obj.param
+        if not self_._events:
+            continue
+        # Only looked up by key below, never iterated, so plain dict is enough.
```
philippjfr:
```suggestion
```

PR #1205, opened 2026-09-16
Add Watcher.remove() to avoid the .param.unwatch(watcher) dance

philippjfr:
Stopping a watcher required calling back into the object it was registered on, `(watcher.inst or watcher.cls).param.unwatch(watcher)`, even though `Watcher` already stores `inst` and `cls` and is used exactly that way internally in three separate places (`_update_ref`, `_update_deps`, `.rx.unwatch()`). This PR adds a `remove()` method to `Watcher` that does the lookup itself:

```python
import param

class P(param.Parameterized):
    x = param.Number(default=0)

p = P()
watcher = p.param.watch(print, 'x')
watcher.remove()   # instead of (watcher.inst or watcher.cls).param.unwatch(watcher)
p.x = 1             # nothing printed, watcher is gone
```

`Watcher` was already a proper class, not a bare namedtuple (a recent change due to typing): `.param.unwatch(watcher)` keeps working unchanged. 

The three internal call sites listed above now call `watcher.remove()` instead of reimplementing the `inst`/`cls` lookup.

In drafting this PR we also discovered that `Parameters._register_watcher` removed watchers with `list.remove(watcher)`, which matches by value equality since `Watcher` is a namedtuple. Two independently constructed `Watcher`s with identical fields, e.g. the same callback watched twice on the same parameter, are indistinguishable to `list.remove`, so removing one could silently remove the other instead. This was also fixed.

## AI Disclosure

Tool & Model: Kilo Code + anthropic/claude-sonnet-5

Usage: Used to research the existing Watcher/watch/unwatch implementation and storage across the codebase, implement `Watcher.remove()` and the identity-based removal fix, update internal call sites, write and run a verification script, and draft this description.

## Checklist

- [x] Added documentation


param/parameterized.py
```diff
@@ -1115,6 +1115,9 @@ class Watcher(_Watcher):
     `precedence` : A numeric value which determines the precedence of
     the watcher.  Lower precedence values are executed
     with higher priority.
+
+    Call `.remove()` on the returned `Watcher` to stop it from being
+    triggered, as an alternative to `(inst or cls).param.unwatch(watcher)`.
```
```suggestion
    Call `.remove()` on the returned `Watcher` to stop it from being
    triggered`.
```

I don't even think we should mention the prior way to do it. This is the way!

Likely also remove it from the notebook.

param/parameterized.py
```diff
@@ -1146,6 +1149,29 @@ def __str__(self):
...
 
+    def remove(self) -> None:
+        """
+        Remove this watcher, stopping it from being triggered by
+        subsequent events on the parameters it is watching.
+
+        Equivalent to `(self.inst or self.cls).param.unwatch(self)` but
+        does not require holding a reference to the watched object.
+        Calling `remove` on an already removed watcher is a no-op.
+
```
```suggestion
```

param/parameterized.py
```diff
@@ -4187,19 +4212,26 @@ def _register_watcher(
...
+            if action == 'append':
+                watchers.append(watcher)
+            else:
+                # Remove by identity rather than equality: two separately
+                # constructed Watcher objects with identical field values
+                # (e.g. the same callback watched twice) are indistinguishable
+                # to list.remove, which could remove the wrong (but equal)
+                # entry. Already-removed watchers are silently ignored since
+                # unwatch is idempotent.
```
hoxbro: shorten

param/parameterized.py
philippjfr: Agree!

philippjfr: Simplified per the suggestion, and applied the same treatment to the notebook (dropped the standalone `.param.unwatch()` bullet, single `.remove()` bullet now).

param/parameterized.py
philippjfr: Removed.

param/parameterized.py
philippjfr: Shortened.

philippjfr:
Thanks for the thorough review, both critical issues confirmed and fixed in ad05442.

**Critical: unwatch on rebuilt/copied/pickled watchers**

Reverted `_register_watcher`'s removal to value-based (`list.remove`), same as base. The identity-based change didn't fix a real bug: two watchers with identical fields are interchangeable, so `unwatch(w1)`/`unwatch(w2)` give the same result either way (confirmed `test_watcher_remove_only_removes_one_registration`, renamed since it isn't actually about identity, still passes with equality-based removal). Meanwhile it broke the real case, `Parameterized.__setstate__` rebuilding watchers with new `Watcher(*args)` on copy/pickle. Added regression tests for deepcopy and pickle round-trips, verified both fail on the pre-fix code and pass now.

**Important: remove() on falsy instances**

Fixed `Watcher.remove()` to use `self.cls if self.inst is None else self.inst`, matching the explicit `is None` check `_update_ref`/`_update_deps` used before this PR (rather than `self.inst or self.cls`). Added a regression test for `remove()` on a falsy instance, and a `testrefs.py` test reproducing your exact scenario (ref source switches away from a falsy `src1`), asserting directly on `src1._param__private.watchers` since `_sync_refs`'s dependency-owner guard happens to mask the leak from a value-propagation angle, worth noting so nobody "fixes" this test to check values instead and loses the coverage.

**Minor**

- Reverted `reactive.py`'s `rx._watchers` bookkeeping filter to the equality-based form for consistency with the above.
- The "does not require holding a reference" docstring sentence was already dropped in the previous round addressing @hoxbro's comments.
- Added a `See Also` pointer from `Parameters.unwatch()`'s docstring to `Watcher.remove()`.
- The `deepcopy` + `@param.depends('sub.a', watch=True)` `__setstate__` bug is real but pre-existing on `main` and unrelated to this PR; agreed it's worth its own issue rather than fixing here.

Full suite: 1785 passed, 2 xfailed.

PR #1206, opened 2026-09-16
Redesign .rx.collect as a top-level collect() returning a namedtuple

philippjfr:
`.rx.collect` (merged in #1199) required chaining off an arbitrary `rx` just to reach the method, and the result was a dict keyed by integer position or string name:

```python
a, b = param.rx(1), param.rx(2)
collected = a.rx.collect(b=b)   # 'a' folded in at key 0
collected.rx.value               # {0: 1, 'b': 2}
```

Neither the anchor requirement nor the dict-of-mixed-keys result reads well, especially for the common case of combining several positional inputs and doing something with all of them, e.g. summing them.

Instead this PR replaces the method with `collect` : a plain function, not a method, returning an expression whose value is a `Collected` namedtuple with an `args` field (one slot per positional argument) and a `kwargs` field (one entry per keyword argument):

```python
collected = param.reactive.collect(a, b)
collected.rx.value                        # Collected(args=(1, 2), kwargs={})
collected.args.rx.pipe(sum).rx.value      # 3
```

`.args`/`.kwargs` are ordinary attributes of the resolved namedtuple, picked up by `rx`'s existing attribute-forwarding (the same mechanism that lets `df_rx.sum()` reach a wrapped DataFrame's `.sum()`), so either can be piped further with no special-casing needed.

Partial-settlement semantics are unchanged from #1199: a slot updates as soon as its input settles rather than waiting for the slowest one, keeps its last value while its input is unsettled again, and holds `Undefined` if it has never settled; `.rx.awaiting` is `True` while any input is unsettled. `error_mode='propagate'` resolves a failing input to a `ReactiveError` in its own slot instead of failing every other slot too; the default `'raise'` fails the whole node, including when an upstream input is already itself propagating its own error. `_eval_collect` no longer republishes an unchanged result to watchers, comparing the freshly built `Collected` against the previous one and raising `Skip` if nothing changed.

## AI Disclosure

Usage:


doc/user_guide/Reactive_Expressions.ipynb
```diff
@@ -1429,6 +1356,99 @@
     "Pass `cascade=False` to release only the expression's own watchers, and never touch its inputs, for a reader `.rx.dispose()` cannot see, e.g. a `Parameter` with `allow_refs=True`, or a `bind`/`pn.bind` consumer. Either way, using a disposed expression afterwards, reading its value or building a new one from it, raises rather than returning a stale value."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "cb7b302e",
+   "source": [
+    "(combining-with-collect)=\n",
```
hoxbro: remove.

doc/user_guide/Reactive_Expressions.ipynb
```diff
@@ -1429,6 +1356,99 @@
...
+   "id": "fb88b872",
+   "source": [
+    "from param.reactive import collect\n",
+    "\n",
+    "async def slow(value, delay):\n",
+    "    await asyncio.sleep(delay)\n",
+    "    return value\n",
+    "\n",
+    "a = rx(1).rx.pipe(slow, delay=0.2)\n",
+    "b = rx(2).rx.pipe(slow, delay=1)\n",
```
hoxbro: Lets change these name to `first` and `second` . Will make the following read easier: ``"Once `first` settles its slot already has a value``

param/__init__.py
```diff
@@ -99,7 +99,7 @@
     CalendarDateRange,
     Event,
 )
-from .reactive import ReactiveError, bind, current_node, rx
+from .reactive import ReactiveError, bind, collect, current_node, rx
```
hoxbro: Is `collect` to generic?

tests/testreactive.py
```diff
@@ -4360,129 +4360,146 @@ def accessor(node):
     assert installed_for == [n]
 
 
-# .rx.collect
+# param.reactive.collect
 
 async def test_reactive_collect_reports_partial_results_as_inputs_settle():
```
hoxbro: Make it a class?

param/reactive.py
```diff
@@ -3411,38 +3413,35 @@ def _eval_collect(self, operation):
...
+        new_args = tuple(
+            resolve_one(arg, prev_args[i] if i < len(prev_args) else Undefined)
+            for i, arg in enumerate(args)
+        )
+        new_kwargs = {
+            k: resolve_one(arg, prev_kwargs.get(k, Undefined))
+            for k, arg in kwargs.items()
+        }
+        result = Collected(args=new_args, kwargs=new_kwargs)
         if result == previous:
```
hoxbro: Would this crash if it contains pandas/numpy arrays?

PR #1208, opened 2026-09-17
Fix .rx.overrides and Parameter(allow_refs=True) invisibility to the rx graph walk

philippjfr:
Various methods including `_upstream()`/`_downstream()`, `.rx.awaiting`, `.rx.stale`, and `.rx.updating()` all silently
misreported across two kinds of boundary the pipeline walk (`_direct_inputs()`) never followed:

- An `.rx.overrides` replacement. `_direct_inputs()` only looked at an operation's originally
  wired `args`/`kwargs`, never `operation['overrides']`, so once an input was masked, the node
  actually feeding the computation was invisible to every walk and to `.rx.awaiting`/`.rx.stale`.
- A `Parameter(allow_refs=True)` holding an already-constructed `rx` (the "stable outlet" idiom).
  `_awaiting_ref` only recognized a bare async callable as "still resolving"; it did not know an
  `rx` assigned to such a Parameter could itself be awaiting.

```python
a = param.rx(10)
placeholder = param.rx(1)
b = a * placeholder
new_sub = param.rx(2).rx.pipe(some_async_op)
b.rx.overrides[0] = new_sub
b.rx.value                      # schedules some_async_op through new_sub

new_sub.rx.awaiting             # True (correct)
b.rx.awaiting                   # False (wrong: some_async_op backs one of b's live inputs)
b.rx.updating().rx.value        # False throughout, never flips
```

`.rx.updating()` had a second, independent bug on top of this: it snapshots `_upstream()` once at
construction and never recomputes it, so a node that starts feeding the pipeline later, through an
override or a ref rewire, never gets a settle-watcher attached, even once the two boundary bugs
above are fixed.

## Solution

- `_direct_inputs()` now substitutes `operation['overrides']` for a masked argument, mirroring
  what `_live_params()` already did, and gained a new `_ref_inputs()` step that walks the raw `rx`
  behind a `Parameter(allow_refs=True)`'s reference (`owner._param__private.refs[name]`). This
  fixes `_upstream()`/`_downstream()`, `.rx.awaiting`, and `.rx.stale` outright.
- `InputOverrides.__setitem__`/`__delitem__` now register/drop the override's value as a reader of
  the overriding node, symmetric with what `rx.__init__` already does for an ordinary operation
  argument, so `_downstream()` and `.rx.dispose()`'s cascade-safety check see through an override
  too.
- `.rx.updating()` no longer takes a one-time snapshot of `_upstream()`. It keeps a live set and
  re-derives it whenever a subscribed node's overrides change or a `Parameter(allow_refs=True)` fn
  param it depends on is reassigned, catching up immediately on a node that is already settling by
  the time it is discovered rather than waiting for a settle notification that already fired.
- The ref-rewire half of that needed a new hook `Parameter.__set__` didn't provide: assigning a
  fresh async `rx` to an `allow_refs=True` Parameter always resolves to `Undefined` synchronously
  and returns before firing any parameter-changed watcher, so nothing in the existing reference
  machinery told `rx` that a ref had changed. Added `register_ref_change_callback()` in
  `parameterized.py`, mirroring the existing `register_reference_transform()`, fired whenever
  `_param__private.refs[name]` is set (by the constructor or by `__set__`), independent of whether
  the resolved value differs. `reactive.py` uses it to notify whichever `rx` nodes depend on that
  `(owner, name)` through a small weak registry, populated lazily and pruned once the owner or the
  interested node is garbage collected.

## AI Disclosure

Tool & Model: Kilo Code + Claude Sonnet 5

Usage: Investigated and triaged the review feedback that motivated this fix, designed and wrote
the implementation across `param/reactive.py` and `param/parameterized.py`, and wrote the test
coverage in `tests/testreactive.py`, including discovering and fixing two gaps not anticipated by
the initial design (the notification-ordering issue between eager `.rx.watch()` resolution and
graph-change subscription, and the missing ref-change hook for async reassignment).


param/parameterized.py
```diff
@@ -219,6 +220,36 @@ def transform_reference(arg):
         arg = transform(arg)
     return arg
 
+def watch_ref_change(owner: 'Parameterized', name: str, callback: Callable[[], t.Any]) -> None:
+    """
+    Weakly register ``callback`` to run whenever ``owner``'s raw dynamic
+    reference for parameter ``name`` (``owner._param__private.refs[name]``)
+    is replaced, including by the constructor and regardless of whether the
+    resolved *value* changed.
```
hoxbro: The parenthesis is very technical, why would an user care?

param/parameterized.py
```diff
@@ -219,6 +220,36 @@ def transform_reference(arg):
...
+    resolves to ``Undefined`` at assignment time, which does not trigger an
+    ordinary parameter-changed watcher, so this is the only way to learn a
+    reference was rewired.
+    """
+    private = owner._param__private
+    watchers = private.ref_change_watchers
+    if watchers is None:
+        watchers = private.ref_change_watchers = {}
+    listeners = watchers.setdefault(name, [])
+    ref_type = weakref.WeakMethod if inspect.ismethod(callback) else weakref.ref
```
hoxbro, line 224: What if it is a partial on a method?

param/parameterized.py
```diff
@@ -219,6 +220,36 @@ def transform_reference(arg):
...
+    An asynchronous reference (e.g. a fresh, unsettled ``rx``) always
+    resolves to ``Undefined`` at assignment time, which does not trigger an
+    ordinary parameter-changed watcher, so this is the only way to learn a
+    reference was rewired.
+    """
+    private = owner._param__private
+    watchers = private.ref_change_watchers
+    if watchers is None:
+        watchers = private.ref_change_watchers = {}
+    listeners = watchers.setdefault(name, [])
```
hoxbro: should we make this a set?

param/parameterized.py
```diff
@@ -5912,6 +5944,9 @@ class _InstancePrivate:
                 parameter_attribute (e.g. 'value'): list of `Watcher`s
     values: dict
         Dict of parameter name: value.
+    ref_change_watchers: dict | None
```
```suggestion
    ref_change_watchers: dict[str, list[weakref.ReferenceType]] | None
```

Maybe?

param/parameterized.py
philippjfr: Not a problem, the `partial` will hold the reference to the method, and `ref` will hold the reference to the partial, so everything works out.

PR #1209, opened 2026-09-18
Add .rx.distinct() to skip propagation

philippjfr:
An `rx` node currently recomputes and propagates to its readers on every upstream invalidation, whether or not the recomputed value actually differs from the previous one:

```python
src = param.rx(df)
out = src.rx.pipe(body)
out.rx.value
src.rx.value = df                              # the same object
out.rx.value
src.rx.value = pd.DataFrame({"a": [1, 2, 3]})  # equal but distinct
out.rx.value
# body ran 3 times for one distinct input value
```

A user should be able to express that no downstream computation should be triggered unless the value has changed.

The new `.rx.distinct(fn=None)` special method compares the new value against the previous one (default `Comparator.is_equal`, overridable via `fn(old, new)`) and, when they compare equal, does not update — so its own readers neither see a new value nor recompute, and neither do nodes further downstream.

Implemented on the same pattern as `.rx.where()`: a `Trigger` helper `Parameterized` holds the gate. An internal watcher observes every recompute of the calling expression, compares the new value against a closure-held `current` and only calls `trigger.param.trigger('value')` when they differ. Unless the value changed nothing downstream of `.rx.distinct()` is invalidated by a recompute that left the value unchanged.

## Alternatives

The alternative would be to automatically gate all recomputes but this has two problems:

1) It is potentially expensive to gate every node change like this.
2) It leaves no API surface for registering a custom equality check.

## AI Disclosure

Usage:


param/reactive.py
```diff
@@ -185,6 +185,33 @@ def __init__(self, parameters=None, internal=False, **params):
...
+    reports a change, since there is nothing yet to compare against.
+    """
+
+    _unset = object()
+
+    def __init__(self, is_equal=Comparator.is_equal):
+        self._is_equal = is_equal
+        self.value = self._unset
+
+    def update(self, new) -> bool:
```
hoxbro: Maybe a better name? `update` name is related to the internals of `_EqualityGate` whereas we are using it from external, where the name can be confusing.

hoxbro (approved):

PR #1210, opened 2026-09-18
Stop a skipped reference from discarding a sibling argument's genuine change

philippjfr:
## What

A skipped `rx` argument used to raise `Skip` while an operation resolved its arguments. That aborted the whole operation, including any sibling argument that genuinely changed. The problem predates `.rx.distinct()` but that API makes it common.

```python
a = quotes.rx.pipe(select, "TGT").rx.distinct()
b = quotes.rx.pipe(select, "ACME").rx.distinct()
out = rx(combine)(a=a, b=b)
```

When only `ACME` changed, `a` skipped and `out` stayed stale.

## Fix

Each node now keeps a monotonic settle count. Operations compare the counts of their trackable bare-`rx` inputs, including `_prev`, with the last evaluation:

- A skipped input with a prior settled value is usable when another input moved.
- If every tracked input is unchanged, the operation raises `Skip`, preserving `.rx.distinct()`'s downstream no-recompute contract.
- `_prev` uses the same comparison, so `a + y` and `y + a` agree.
- References whose changes cannot be proven from settle counts (`bind`, Parameters, custom reference transforms, containers containing references, and active overrides) opt out conservatively and evaluate.

Literal containers are constants, so they remain trackable. Shared clones use their shared source's settle state before resolving independently.

Nested `rx` values in container arguments remain conservative: they may recompute unnecessarily, but do not suppress genuine changes.

## Tests

`TestSkipSiblingArguments` covers:

- `.rx.distinct()` and plain `Skip` sibling updates
- all-skipped watcher suppression
- left-operand propagation
- never-settled and settling inputs
- literal list arguments
- custom reference transforms
- `batch()` and `.rx.when()`

Validation: full suite **1,877 passed, 4 skipped, 2 xfailed**. Downstream weiss_mna suite: **245 passed** with this branch on `PYTHONPATH`.


param/reactive.py
```diff
@@ -3481,9 +3496,37 @@ def _resolve_input(self, arg):
...
+        treatment yet.
+        """
+        if isinstance(arg, rx):
+            if arg._settling:
+                raise Skip
+            value = arg.rx.value
+            if value is Skip or value is Undefined:
+                raise Skip
+            if arg._skipped and not arg._has_settled:
+                raise Skip
```
hoxbro, line 3539: Could be one if statement.

param/reactive.py
```diff
@@ -3481,9 +3496,37 @@ def _resolve_input(self, arg):
...
+        argument still go through ``resolve_value`` and do not get this
+        treatment yet.
+        """
+        if isinstance(arg, rx):
+            if arg._settling:
+                raise Skip
+            value = arg.rx.value
+            if value is Skip or value is Undefined:
+                raise Skip
+            if arg._skipped and not arg._has_settled:
```
```suggestion
            if value is Skip or value is Undefined or (arg._skipped and not arg._has_settled):
```

philippjfr:
Thank you. Both findings were correct. The first version tolerated a settled skip but had no way to distinguish "this input stayed skipped and nothing else moved" from "this input stayed skipped while a sibling moved", and it also missed `_prev` (the left-operand/method path).

Reworked the fix and force-pushed `86e2f4f`:

- Replaced `_has_settled` with a monotonic `_settle_count` per node.
- An operation records the settle counts of its trackable bare-`rx` inputs, including `_prev`, from its last real evaluation. It raises `Skip` when they are unchanged, including an all-skipped first attempt. That restores `.rx.distinct()`'s "downstream nodes do not recompute either" contract and prevents the watcher notification in the XYZ-only repro.
- `_prev` now uses the same tolerance and generation comparison, so `a + y` and `y + a` agree after `a` skips.
- Any input the implementation cannot prove unchanged (references resolved through `bind()`, Parameter/function/container references, or a currently-overridden input that may have been reassigned to a different plain value) opts out of the suppression and runs conservatively. This preserves the existing override/following-reference and generator behavior.
- Nested `rx` values inside a container remain an explicit limitation, but now only in the conservative direction: they may recompute unnecessarily, never suppress a genuine change.

Added regressions for the XYZ-only watcher case and `a + y`; `TestSkipSiblingArguments` now has seven focused tests. Validation: full Param suite **1,830 passed, 4 skipped, 2 xfailed**; downstream weiss_mna suite **245 passed** with this worktree on `PYTHONPATH`.

philippjfr:
Addressed the follow-up review. Literal containers now remain trackable unless they contain a reference, custom reference transforms are classified after transform_reference(), and shared clones use their shared source's settle state. Added behavioral coverage for literal lists, custom transforms, batch(), .rx.when(), all-skipped watcher suppression, and the left-operand path. The description now matches the settle-count implementation and states the conservative nested-container limitation. Validation: 1,879 Param tests and 245 downstream weiss_mna tests pass.

